PR #35, opened 2014-06-09
Airbrake error fallback

eapache:
Fallback to `msg` in airbrake hook if:

- no `error` was provided (or it wasn't an `error` type)
- `msg` is an `error` type

For convenience since often `msg` will just contain the anyways.

Also fixes a panic if `error` was not an `error`.


aybabtme:
I'll close this since it's been here forever.

PR #100, opened 2015-01-04
make sure no trailing spaces

### LINKED ISSUE #99: text log should remove trailing space
current code is generating every log line with a trailing space, which is not good and should be eliminated

``` go
tf := &TextFormatter{DisableColors: true}
byts, _ := tf.Format(&Entry{Message: "msg content"})
fmt.Printf("%q\n", string(byts))
```

``` text
"time=\"0001-01-01T00:00:00Z\" level=panic msg=\"msg content\" \n"
```

crquan:
Closes #99 text log should not have trailing spaces.

just reorganize the patch into 3 commits:
03377c6 rename f.appendKeyValue to printKeyValue
a243bba share common calling path in printKeyValue
dcbe8d6 make sure no leading or trailing spaces

In benchmark testing, it increased ~300 ns/op in SmallTextFormatter,
that should be due to the extra slice operation [1:];
in LargeTextFormatter it reduced ~700 ns/op;

```text
BenchmarkSmallTextFormatter       200000             10683 ns/op           8.14 MB/s
BenchmarkLargeTextFormatter        30000             43537 ns/op           6.59 MB/s
```

```text
BenchmarkSmallTextFormatter       200000             10996 ns/op           7.82 MB/s
BenchmarkLargeTextFormatter        30000             42862 ns/op           6.67 MB/s
```

Signed-off-by: Derek Che <drc@yahoo-inc.com>

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -107,18 +107,16 @@ func needsQuoting(text string) bool {
 func printKeyValue(b *bytes.Buffer, key, value interface{}) {
 	switch value.(type) {
 	case string:
-		if needsQuoting(value.(string)) {
-			fmt.Fprintf(b, "%v=%s ", key, value)
-		} else {
-			fmt.Fprintf(b, "%v=%q ", key, value)
-		}
+		break
 	case error:
-		if needsQuoting(value.(error).Error()) {
-			fmt.Fprintf(b, "%v=%s ", key, value)
-		} else {
-			fmt.Fprintf(b, "%v=%q ", key, value)
-		}
+		value = value.(error).Error()
 	default:
 		fmt.Fprintf(b, "%v=%v ", key, value)
+	}
+
+	if needsQuoting(value.(string)) {
+		fmt.Fprintf(b, "%v=%s ", key, value)
+	} else {
+		fmt.Fprintf(b, "%v=%q ", key, value)
 	}
 }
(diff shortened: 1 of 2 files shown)


Sirupsen:
I think this solves the problem at the wrong layer, rather you should be able to pass an argument to `appendKeyValue` to remove the space, or always return `b[0..len(b)-2]`—the latter seems simpler. A comment would be appropriate too, since it's not obvious why that is.

Thanks for working on this!

crquan:
or maybe code like this, just remove one possible space,

```go
        if byts[len(byts)-1] == ' ' {
                byts = byts[:len(byts)-1]
        }
```

crquan:
I have another TestColorPrint which tests how many spaces can the printColored func make: there could be many of them:

```text
=== RUN TestColorPrint
"\x1b[31mPANI\x1b[0m[0000] msg content                                  \n"
```

So it's better to run in a loop:

```go
for byts[len(byts)-1] == ' ' {
    byts = byts[:len(byts)-1]
}
```

or call bytes.TrimRightFunc:

```go
byts = bytes.TrimRightFunc(byts, unicode.IsSpace)
```

crquan:
@Sirupsen do you have an update on this? I'm looking into this because docker code is referencing here, and every docker ci has log lines ends a space.
https://ci.dockerproject.com/github.com/docker/docker/commit/e00e161f4738cede19b37f4cd33ae643863ea570?branch=10081-fix-renaming

Sirupsen:
Can you squash? Otherwise looks good

crquan:
this was initially in a single patch but I made it 3 separate to make each one's intention be clear; now to squash them is surely easy, but why does here prefer a squashed single commit? or any worries on number of commits?

Issue #47, opened 2014-07-22
Provide the implementation as a custom io.Writer


ant0ine:
Hi,

Is there a possibility or a plan to provide the Logrus implementation as a custom io.Writer that can be used in conjonction with the standard library logger ?

https://godoc.org/github.com/Sirupsen/logrus#StdLogger is nice but it still makes the package depend on Logrus, only to get the interface.

I'm discussing the issue here: https://github.com/ant0ine/go-json-rest/issues/62

Thanks,

Antoine

phemmer:
I'm in favor of logrus implementing the `io.Writer` interface. The reason for this is that many libraries use the default logger (e.g. `log.Println("foo")`). By allowing us to set the default logger to send to logrus, we can take advantage of logrus's hooks. This can be very important if the we are using the hooks to send to multiple destinations.

Sirupsen:
#103 implemented this!

Thanks @phemmer :heart:

PR #103, opened 2015-01-09
add io.Writer interface compatability

phemmer:
This is an implementation for #47.
*Note: this PR has been re-written since the original submission*

This adds a `(*Logger) InputWriter()` function which returns an `io.PipeWriter` which can be passed to the stdlib `log.SetOutput()`.

The `InputWriter()` call creates an `io.Pipe` and a goroutine which reads from it, and returns the `io.PipeWriter`. Once the `io.PipeWriter` is garbage collected, `close()` is called on it, which causes the goroutine to return.

I can add documentation & tests if it looks ok.

Example:

```
package main

import (
    "github.com/Sirupsen/logrus"
    "log"
)

func main() {
    logrus.Print("logrus")
    logrusInputWriter := logrus.StandardLogger().InputWriter()
    log.SetOutput(logrusInputWriter)
    log.SetFlags(0)
    log.Print("log")
    logrusInputWriter.Write([]byte{'a'})
    logrusInputWriter.Write([]byte{'b'})
    logrusInputWriter.Write([]byte{'c'})
    logrus.Print("interrupt")
    logrusInputWriter.Write([]byte{'\n'})
    log.Fatal("FOO!")
}
```

```
INFO[0000] logrus                                       
INFO[0000] log                                          
INFO[0000] interrupt                                    
INFO[0000] abc                                          
INFO[0000] FOO!   
exit status 1                                      
```


phemmer:
Added support for parallel operations (previous implementation could have issues).

This version functions exactly the same, but has a slightly different interface from the one documented in the PR description.

```
package main

import (
    "github.com/Sirupsen/logrus"
    "log"
)

func main() {
    logrus.Print("logrus")
    logrusInputWriter := logrus.StandardLogger().InputWriter()
    log.SetOutput(logrusInputWriter)
    log.SetFlags(0)
    log.Print("log")
    logrusInputWriter.Write([]byte{'a'})
    logrusInputWriter.Write([]byte{'b'})
    logrusInputWriter.Write([]byte{'c'})
    logrus.Print("interrupt")
    logrusInputWriter.Write([]byte{'\n'})
    log.Fatal("FOO!")
}
```

The `InputWriter()` call creates an `io.Pipe` and a goroutine which reads from it, and returns the `io.PipeWriter`. Once the `io.PipeWriter` is garbage collected, `close()` is called on it, which causes the goroutine to return.

input_writer.go
```diff
@@ -0,0 +1,30 @@
...
+	return inputWriter
+}
+
+func (logger *Logger) inputWriterScanner(inputReader *io.PipeReader) {
+	scanner := bufio.NewScanner(inputReader)
+	for scanner.Scan() {
+		logger.Print(scanner.Text())
+	}
+	if err := scanner.Err(); err != nil {
+		logger.Fatal(err)
```
aybabtme: this should be `Error` or `Warn`, not `Fatal`

input_writer.go
```diff
@@ -0,0 +1,30 @@
...
+	"bufio"
+	"io"
+	"runtime"
+)
+
+func (logger *Logger) InputWriter() (*io.PipeWriter) {
+	inputReader, inputWriter := io.Pipe()
+
+	go logger.inputWriterScanner(inputReader)
+	runtime.SetFinalizer(inputWriter, inputWriterFinalizer)
```
aybabtme: not sure this is desired or necessary, the returned object is a `io.PipeWriter` which implements `io.Closer`, so it should be closed by the caller.

input_writer.go
aybabtme: when the `inputWriter` get closed, I think this end will get an `err == io.EOF`, so this would `os.Exit` the process as soon as the external writer stops writing.

aybabtme:
This is not crazy and I can think of places where I'd need this. I'm commented a few things.

input_writer.go
phemmer: EOF is not considered an error condition by bufio.Scanner ([doc](http://golang.org/pkg/bufio/#Scanner.Scan)). But I'll change it to non-fatal, and have it close the PipeReader so writes will fail.

input_writer.go
phemmer: But if someone does something such as:

```
log.SetOutput(logger.InputWriter())
...
log.SetOutput(somethingElse)
```

We'll leak a goroutine since the io.PipeWriter never gets closed (I tested this scenario). If the user wants to explicitly close the io.PipeWriter we give them, they can, but I didn't see any harm in closing it ourself if the garbage collector comes around and removes the InputWriter.

aybabtme: Maybe return an `io.WriteCloser` and specify in the doc that the writer must be closed. Failing to close it is a leak, but its reasonable to expect people to close their resources. 

Since the writer should be closed, I'm not super keen on using `runtime.SetFinalizer` to make up for people writing buggy code.

input_writer.go
aybabtme: cool

input_writer.go
```diff
@@ -0,0 +1,30 @@
+package logrus
+
+import (
+	"bufio"
+	"io"
+	"runtime"
+)
+
+func (logger *Logger) InputWriter() (*io.PipeWriter) {
```
aybabtme: that's just a nitpick, but I think I'd prefer `Writer` to `InputWriter`

input_writer.go
phemmer: I wouldn't consider relying on that behavior as buggy code. The [os](http://golang.org/pkg/os) package does exactly this for any File objects it creates (https://golang.org/src/os/file_unix.go#L47)

input_writer.go
phemmer: The reason I called it `InputWriter` was because of ambiguity. I felt that as just `Writer`, it was unclear if this was something logrus writes to, or something for writing to logrus.
But it's your project, I'll change it :-)

input_writer.go
aybabtme: I'll admit that this surprises me a lot.

input_writer.go
aybabtme: hum makes sense

@Sirupsen is the owner, I'm just helping out with PRs and issues =]

phemmer:
Pushed a commit for the `logger.Fatal()` line. Didn't do the other 2 as it seems like their fate might still be undecided.

input_writer.go
Sirupsen: Hm, I'm not sure I understand why `InputWriter`? Why `Input`? How do you feel it was unclear if it was just called `Writer`?

```go
log = logrus.New()
log.Writer()
# or
logrus.Writer() # if we expose at package level, which makes sense
```

Sirupsen:
Why do we need a goroutine here? Couldn't the writer just hook into `logrus`?

aybabtme:
@Sirupsen by using an `io.Pipe`, the thing that will write to the writer will see it's message written in structure form (time/level/msg).  If instead, we exposed a writer to the internal buffer of logrus, we wouldn't get this nicety and also, there would be races on writes to that buffer. Thus the `io.Pipe`.

When working with a pipe, there needs to be two different goroutines matching writes with reads, otherwise the writer will block forever.

phemmer:
Added a commit renaming `InputWriter` to just `Writer`

phemmer:
Expanding on the `io.Pipe`. aybabtme pretty much hit it. In fact the first version of this pull request did hook directly into logrus without using an `io.Pipe`. But it was extremely racy, and prone to failure. So I switched to using the pipe which solves the issue.

aybabtme:
lgtm, @Sirupsen ?

Sirupsen:
Needs some documentation in the READMe otherwise LGTM.

aybabtme:
I'm fine writing the docs in the README, I think we've strained @phemmer enough =P

Sirupsen:
haha yes, sorry

johntdyer:
Sorry for the possibly dumb question, but it i not clear to me how to use this...

aybabtme:
Hey John, another pending PR of mine shows an example of how to use it. 




Antoine Grondin
Live long and prosper

aybabtme:
@johntdyer see https://github.com/Sirupsen/logrus#logger-as-an-iowriter

```go
w := logger.Writer()
defer w.Close()

srv := http.Server{
    // create a stdlib log.Logger that writes to
    // logrus.Logger.
    ErrorLog: log.New(w, "", 0),
}
```

robbiet480:
Is there no way to pass through the log levels to Logrus, instead of having everything appear as INFO?

aybabtme:
@robbiet480 there is no way to do this right now, but if you have a clean way to do it, we can look at it in a PR

PR #104, opened 2015-01-10
The format is logfmt, not l2met, although that is where it "originated"

freeformz:


Sirupsen:
l2met is deprecated or something?

freeformz:
l2met is software that converts log lines into metrics. logfmt is the formalization of the key/value format that l2met and others understand. The format has grown beyond l2met and has been formalized as logfmt. c.f. https://github.com/search?utf8=%E2%9C%93&q=logfmt

PR #106, opened 2015-01-12
fix typo in README (WithFields log/logrus)

phemmer:


Sirupsen:
That's only if you alias the `logrus` import to `log`.

phemmer:
1. All the surrounding lines use `log`.
2. The very same line uses both `log` and `logrus`. it should either be `log.WithFields(log{` or `logrus.WithFields(logrus{`

Sirupsen:
This example is not for the package-level logger, but for an instance of the log: `log := logrus.New()`. The field type is defined on the package-level, but `WithFields` is defined on `log`, an instance of the log. `logrus.WithFields(logrus.Fields{})` is valid too, but since the example highlights instance logging, and not package-level logging, that's not the case here.

PR #111, opened 2015-01-15
Added a missing type check to the printKeyValue function

pzduniak:
Seems pretty vital unless you like panics.


Sirupsen:
Why was this closed?

PR #124, opened 2015-02-09
Better API to prevent unnecessary argument evaluation

pjvds:
I always loved the idea of structured logging, and when I stumbled upon logrus last week I decided to improve some of our logging code. But while migrating from `glog` to `logrus` I discovered that it lacks an good API to check the current log level to prevent unnecessary argument evaluation.
## Current state

``` go
log.Debugf("π: %v", calculate_pi())
```

In this example if the current log level is above debug, the message will not be logged. That is great, but this decision is made inside the `Debug` function. This means that  arguments are evaluated and the `calculate_pi` method will be executed for no good reason.

This can easily be fixed with an `if` statement:

``` go
if log.GetLevel() >= log.DebugLevel {
    log.Debugf("π: %v", calculate_pi())
}
```

But too be honest, I don't really like that API for the following reasons:
1. Inconsistent: direct package api has `GetLevel()` while `logger` instance has `Level` member.
2. Doesn't hide the way log levels are compared.
## My API proposal

I added `IsDebug`, `IsInfo`, `IsWarn`, `IsFatal` and `IsPanic` as static functions in `exported.go` and as methods to the `Logger` type in `logger.go`.

Here is how code that uses my changes is looks like:

``` go
if log.IsDebug() {
    log.Debugf("π: %v", calculate_pi())
}
```
## Tests

I've added tests to proof the methods are returning the right answer depending on the current log level. These can be found in `logrus_test.go`.

### DIFF
--- a/exported.go
+++ b/exported.go
@@ -39,6 +39,93 @@ func GetLevel() Level {
 	return std.Level
 }
 
+// Reports whether log level is at least debug level.
+//
+// This method can be used to prevent evaluation of arguments if
+// the log level isn't high enough:
+//
+//      if log.IsDebug() {
+//           log.Debugf("π = %v", calculatePi())
+//      }
+//
+// In this case the method `calculatePi` will not be evaluated
+// when the log level isn't at least debug.
+func IsDebug() bool {
+	return std.Level >= DebugLevel
+}
+
+// Reports whether log level is at least info level.
+//
+// This method can be used to prevent evaluation of arguments if
+// the log level isn't high enough:
+//
+//      if log.IsDebug() {
+//           log.Debugf("π = %v", calculatePi())
+//      }
+//
+// In this case the method `calculatePi` will not be evaluated
+// when the log level isn't at least debug.
+func IsInfo() bool {
+	return std.Level >= InfoLevel
+}
+
+// Reports whether log level is at least warning level.
+//
+// This method can be used to prevent evaluation of arguments if
+// the log level isn't high enough:
+//
+//      if log.IsDebug() {
+//           log.Debugf("π = %v", calculatePi())
+//      }
+//
+// In this case the method `calculatePi` will not be evaluated
+// when the log level isn't at least debug.
+func IsWarn() bool {
+	return std.Level >= WarnLevel
+}
+
+// Reports whether log level is at least error level.
+//
+// This method can be used to prevent evaluation of arguments if
+// the log level isn't high enough:
+//
+//      if log.IsDebug() {
+//           log.Debugf("π = %v", calculatePi())
+//      }
+//
+// In this case the method `calculatePi` will not be evaluated
+func IsError() bool {
+	return std.Level >= ErrorLevel
+}
+
+// Reports whether log level is at least fatal level.
+//
+// This method can be used to prevent evaluation of arguments if
+// the log level isn't high enough:
+//
+//      if log.IsDebug() {
+//           log.Debugf("π = %v", calculatePi())
+//      }
+//
+// In this case the method `calculatePi` will not be evaluated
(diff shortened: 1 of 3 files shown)


pi3r:
@pjvds this is really cool. Well done for the nicely documented pull request :)

Sirupsen:
Would agree with @pi3r, thanks for taking the time to explain your problem and solution in depth. I've thought about this use case since I wrote the first version of Logrus (but not implemented it, as I haven't found need for it myself), however, I don't love having to add another 2 lines for every log statement. I've found in real life Logrus logging already takes up a lot of lines, and I don't want to encourage more of that. Logging is not the core of the code.

The way I thought about implementing it initially is a little simpler, just passing the function pointer. When iterating over the fields we'd look for function types and execute them in Logrus, meaning in the majority of cases only the pointer is passed (cheap), but not evaluated.

```go
log.WithField("pi", calculate_pi).Debug("calculated pi")
```

This doesn't work for your `Debugf` case, but we could hack even the `strf` family functions to have consistent behaviour. I do think that these cases should be using `WithField{,s}` though. The use of `strf`-type functions in Logrus is discouraged. It's a strong hint you should be using a field.

pjvds:
Thanks for the kind words. I thought about that function pointer as well, or sending a complete closure to the log. I agree that it might simplify the code in some cases, but it fails for other situations, for example where you init state based on the log level as well.

Here is a little example from our current code base where we init state based on the log leven, and also do calculations later. Wrapping that `bytes_per_second` calculation inside a function will not make it prettier.

``` go
var started time.Time

if log.IsDebug() {
    started = time.Now()
}

// ...

if glog.IsDebug() {
    // we failed to do zero copy write if a buffer is created
    // this can happen when the values exceed 5mb
    zeroCopy := (buffer == nil)
    elapsed := time.Since(started)
    
    log.WithFields(log.Fields{
        "zerocopy": zeroCopy,
        "total_size": writtenBytes,
        "elapsed": elapsed,
        "fsync", opt.Sync,
        "bytes_per_second": float64(writtenBytes) / elapsed.Seconds(),
    }).Debugf("writer finished")
}
```

xujinzheng:
@pjvds the project of logrus mate may help you, https://github.com/gogap/logrus_mate

pjvds:
@xujinzheng Thanks for the reply, but I am using [`tidy`](https://github.com/pjvds/tidy) now that focusses on performance rather than features.

crantok:
The OP's original suggestion was gently rejected because
> I don't love having to add another 2 lines for every log statement

however, the suggested API would not cause 2 lines to be added for every log statement. It would only need to be used by client code when a redundant expensive operation was being avoided. Most log statements would therefore not require these extra two lines.

I would support the original suggestion (which would be quick to implement) if lazy evaluation was going to take a while to arrive. I would however change the function name to make it obvious that we are not asking whether the logger is configured with a given Level, but whether the logger will log entries that have a given level. So instead of
```go
if log.IsDebug() {
    log.Debugf("π: %v", calculate_pi())
}
```
I would have
```go
if log.WillLogDebug() {
    log.WithFields(log.Fields{"π": calculate_pi()}).Debug("It's a pi!")
}
```
or (even quicker to code)
```go
if log.WillLog(log.DebugLevel) {
    log.WithFields(log.Fields{"π": calculate_pi()}).Debug("It's a pi!")
}
```
If this sounds acceptable then I'm happy to code it and submit a pull request.

aybabtme:
I like the `log.WillLog(log.DebugLevel)` idea, makes the API change smaller. `log.IsLevel(log.DebugLevel)` would read better IMO.

@Sirupsen, wdyt?

dgsb:
duplicate of #762

PR #127, opened 2015-02-17
[Race] Fix datarace in GetLevel

noxiouz:
`std.Level` is protected by mutex in setter (SetLevel), so it must be protected in geetter (GetLevel) too.

``` 
go test -v -race github.com/Sirupsen/logrus
=== RUN TestGetSetLevelRace
==================
WARNING: DATA RACE
Read by goroutine 32:
  github.com/Sirupsen/logrus.func·041()
      /Users/noxiouz/Go/src/github.com/Sirupsen/logrus/logrus_test.go:291 +0x7c

Previous write by goroutine 31:
  github.com/Sirupsen/logrus.SetLevel()
      /Users/noxiouz/Go/src/github.com/Sirupsen/logrus/exported.go:34 +0xba
  github.com/Sirupsen/logrus.func·041()
      /Users/noxiouz/Go/src/github.com/Sirupsen/logrus/logrus_test.go:289 +0x4c

Goroutine 32 (running) created at:
  github.com/Sirupsen/logrus.TestGetSetLevelRace()
      /Users/noxiouz/Go/src/github.com/Sirupsen/logrus/logrus_test.go:293 +0x48
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:447 +0x133

Goroutine 31 (finished) created at:
  github.com/Sirupsen/logrus.TestGetSetLevelRace()
      /Users/noxiouz/Go/src/github.com/Sirupsen/logrus/logrus_test.go:293 +0x48
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:447 +0x133
==================
```

PR #128, opened 2015-02-17
[Race] Fix datarace in GetLevel using atomics

noxiouz:
Use atomics to operate with `std.Level` to prevent datarace #127 and to speedup access. 
```
with mutex
BenchmarkGetlevelConcurrent	   50000	     33696 ns/op
BenchmarkGetlevel	10000000	       112 ns/op

with atomics
BenchmarkGetlevelConcurrent	  100000	     22216 ns/op
BenchmarkGetlevel	500000000	         3.58 ns/op
```

Signed-off-by: Anton Tiurin <noxiouz@yandex.ru>

### DIFF
--- a/exported.go
+++ b/exported.go
@@ -2,6 +2,7 @@ package logrus
 
 import (
 	"io"
+	"sync/atomic"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


noxiouz:
closed in favor of #127

PR #75, opened 2014-11-04
Adding File, Line, Stacktrace

derekdowling:
This is just an initial approach to get the ball rolling in reference to #63 that I implemented last night really quickly. It needs test, hardening,etc. I am all for a much more permanent solution, but this is at least a start. The two things that need to be addressed are whether or not we are losing context by combining the file & line number like so:

```bash
INFO[0006] Request completed                             file=/bursa/src/bursa.io/kernel/kernel.go:96
```

and how to format the stack trace. Right now it is pretty gross:

```bash
DEBU[0001]                                               file=/bursa/.godeps/src/github.com/Sirupsen/logrus/logger.go line=100 stack=goroutine 3 [running]:
bursa.io/middleware/logtext.getTrace(0x756ec0, 0xc2100c8240)
	/bursa/src/bursa.io/middleware/logtext/logtext.go:46 +0x6e
bursa.io/middleware/logtext.(*Logtext).Format(0xc2100915e0, 0xc2100928c0, 0x43029a, 0x7d9860, 0xc21001ea80, ...)
	/bursa/src/bursa.io/middleware/logtext/logtext.go:36 +0x181
github.com/Sirupsen/logrus.(*Entry).Reader(0xc2100928c0, 0x376d4605, 0xc2100928c0, 0x0)
	/bursa/.godeps/src/github.com/Sirupsen/logrus/entry.go:41 +0x38
github.com/Sirupsen/logrus.(*Entry).log(0xc2100928c0, 0x5, 0x0, 0x0)
	/bursa/.godeps/src/github.com/Sirupsen/logrus/entry.go:84 +0x1e9
github.com/Sirupsen/logrus.(*Entry).Debug(0xc2100928c0, 0x7f64f0934a80, 0x1, 0x1)
	/bursa/.godeps/src/github.com/Sirupsen/logrus/entry.go:109 +0x75
github.com/Sirupsen/logrus.(*Logger).Debug(0xc21004a180, 0x7f64f0934a80, 0x1, 0x1)
	/bursa/.godeps/src/github.com/Sirupsen/logrus/logger.go:100 +0xa3
github.com/Sirupsen/logrus.Debug(0x7f64f0934a80, 0x1, 0x1)
	/bursa/.godeps/src/github.com/Sirupsen/logrus/exported.go:61 +0x48
bursa.io/controller/home.HandleSignup(0x7f64f0ab9ce0, 0xc210092740, 0xc21005e750)
[...]

PR #130, opened 2015-02-18
Logstash formatter and sourcefile hook

alexdemidov:
Hi,

This pull request contains two features:
1. Source File Hook which add source file and line number to entries (log output) for predefined log levels.
2. Logstash Formatter which write output log as Logstash Event (http://logstash.net/).

Thanks,
Alex Demidov


Sirupsen:
Please split your PR into two. There's already a patch like `sourcefile` that has not been merged yet, so it won't be accepted (perhaps you can pitch in on the other ticket? #75).

However, the Logstash formatter I'll happily accept.

alexdemidov:
@Sirupsen I close this pull request and will split it into two. Thanks.

PR #131, opened 2015-02-20
Change DebugLevel color to gray

lvillani:
This is highly subjective but we have an application that spits lots and lots of debug output and demoting it to be nearly "colorless" helps us to visually distinguish it from interesting output.


aybabtme:
I think this could be solved with a custom formatter. But otherwise I'm not against and would merge.

@Sirupsen, any reason against this?

Sirupsen:
This is a good idea!

PR #132, opened 2015-02-20
Logstash formatter

alexdemidov:
Hi,

Pull request contains Logstash Formatter which write output log as Logstash Event (http://logstash.net/).
My previous PR is #130

Thanks,
Alex Demidov

### DIFF
--- a/formatters/logstash/logstash_formatter.go
+++ b/formatters/logstash/logstash_formatter.go
@@ -1,8 +1,9 @@
-package logrus
+package logrus_logstash
 
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/Sirupsen/logrus"
 	"time"
 )
 
@@ -12,7 +13,7 @@ type LogstashFormatter struct {
 	Type string // if not empty use for logstash type field.
 }
 
-func (f *LogstashFormatter) Format(entry *Entry) ([]byte, error) {
+func (f *LogstashFormatter) Format(entry *logrus.Entry) ([]byte, error) {
 	entry.Data["@version"] = 1
 	entry.Data["@timestamp"] = entry.Time.Format(time.RFC3339)
 
--- a/README.md
+++ b/README.md
@@ -323,6 +323,10 @@ The built-in logging formatters are:
 * `logrus.JSONFormatter`. Logs fields as JSON.
 * `logrus.LogstashFormatter`. Logs fields as Logstash Events (http://logstash.net).
 
+    ```go
+      logrus.SetFormatter(&logrus.LogstashFormatter{Type: “application_name"})
+    ```
+
 Third party logging formatters:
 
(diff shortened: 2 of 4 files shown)


oryband:
:+1:

Sirupsen:
The thing that makes me a little hesitant here is to have the formatter in the main directory. How do you feel about putting it in a `formatters` directory ala hooks?

alexdemidov:
@Sirupsen I think it's good idea to have special directory for formatters. I will create such directory and update pull request.

Sirupsen:
Thanks!

alexdemidov:
I put formatter to special directory.

formatters/logstash/logstash_formatter.go
```diff
@@ -0,0 +1,48 @@
...
+	entry.Data["level"] = entry.Level.String()
+
+	// set type field
+	if f.Type != "" {
+		v, ok = entry.Data["type"]
+		if ok {
+			entry.Data["fields.type"] = v
+		}
+		entry.Data["type"] = f.Type
+	}
```
Sirupsen: What about the `msg` field? Perhaps we should consider making [`prefixFieldClashes`](https://github.com/Sirupsen/logrus/blob/a020ac2471759b4903deca15bc6b470cd96498c4/formatter.go#L29-44) public so you can use it here?

formatters/logstash/logstash_formatter.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Formatter generates json in logstash format.
+// Logstash site: http://logstash.net/
+type LogstashFormatter struct {
+	Type string // if not empty use for logstash type field.
+}
+
+func (f *LogstashFormatter) Format(entry *logrus.Entry) ([]byte, error) {
+	entry.Data["@version"] = 1
+	entry.Data["@timestamp"] = entry.Time.Format(time.RFC3339)
```
Sirupsen: What if the user has a field with this name?

formatters/logstash/logstash_formatter.go
alexdemidov: I replaced `msg` to `message`  because Logstash uses `message` field for event description. So i can’t use `prefixFieldClashes`.

formatters/logstash/logstash_formatter.go
alexdemidov: I think the right way is to set `@version` and `@timestamp` fields to user defined values. But in such case maybe problem if user sets incorrect type of fields.
I prefer to leave as is because it's logstash predefined fields.

Sirupsen:
This needs tests, then it's ready to get in master.

alexdemidov:
I will write tests during some days.

alexdemidov:
@Sirupsen Test is ready.

alexdemidov:
@Sirupsen Thanks.

PR #133, opened 2015-02-20
sourcefile hook

alexdemidov:
Hi,

Pull Request contains Source File Hook which add source file and line number to entries (log output) for predefined log levels. Format of output is `package/sourcefile.go:<linenumber>`
Previouse PR is #130

We use this hook in our projects minimum 4 months.

Thanks,
Alex Demidov

### DIFF
--- a/hooks/sourcefile/sourcefile.go
+++ b/hooks/sourcefile/sourcefile.go
@@ -1,0 +1,38 @@
+package logrus_sourcefile
+
+import (
+	"fmt"
+	"github.com/Sirupsen/logrus"
+	"runtime"
+	"strings"
+)
+
+type SourceFileHook struct {
+	LogLevel logrus.Level
+}
+
+func (hook *SourceFileHook) Fire(entry *logrus.Entry) (_ error) {
+	for skip := 4; skip < 9; skip++ {
+		_, file, line, _ := runtime.Caller(skip)
+		split := strings.Split(file, "/")
+		if l := len(split); l > 1 {
+			pkg := split[l-2]
+			if pkg != "logrus" {
+				file = fmt.Sprintf("%s/%s:%d", split[l-2], split[l-1], line)
+				// set source_file field
+				entry.Data["source_file"] = file
+				return
+			}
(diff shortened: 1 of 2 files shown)


alexdemidov:
Maybe should merge it with #75 ?
This hook doesn’t write full stack trace but only add one field (source_file) to log.

alexdemidov:
I know http://godoc.org/runtime#Caller takes time.

In developing or test environment we may write `source_file` for all log levels:
```go
log.Hooks.Add(&logrus_sourcefile.SourceFileHook{LogLevel: logrus.InfoLevel})
```

In production should use Error level:
```go
log.Hooks.Add(&logrus_sourcefile.SourceFileHook{LogLevel: logrus.ErrorLevel})
```

hooks/sourcefile/sourcefile.go
```diff
@@ -0,0 +1,38 @@
...
+	"strings"
+)
+
+type SourceFileHook struct {
+	LogLevel logrus.Level
+}
+
+func (hook *SourceFileHook) Fire(entry *logrus.Entry) (_ error) {
+	for skip := 4; skip < 9; skip++ {
+		_, file, line, _ := runtime.Caller(skip)
```
Sirupsen: Yeah, my problem with this type of patch is that this feels very sketchy and error prone for something as critical as a logger.

alexdemidov:
Thanks. I understand you.

The target of this hook is to add source file and line number to errors logs without modification in Logrus.
I found only way to analyse call stack but there is range of the depth of the call.

Other problem is the searching logrus package name in call stack.

Should i close this pull request?

Sirupsen:
This is a common request, but you're welcome to create your own repository and add it to the README with a disclaimer. Then I'd make the file and line number two different fields.

alexdemidov:
Thanks. I close this pull request.
As for separating fields initially i created two fields for go file and line number without package. In practice such approach is not good because source file and line number are atomic, imho. Also need package name to find error in third party library.

PR #138, opened 2015-02-27
Add Honeybadger.com hook

agonzalezro:


hooks/honeybadger/honeybadger.go
```diff
@@ -0,0 +1,36 @@
...
+type HoneybadgerHook struct {
+	conn *gobadger.Conn
+}
+
+func NewHoneyBadgerHook(apiKey string) *HoneybadgerHook {
+	return &HoneybadgerHook{gobadger.NewConn(apiKey)}
+}
+
+func (hook *HoneybadgerHook) Fire(entry *logrus.Entry) error {
+	err := hook.conn.Error(entry.Message)
```
Sirupsen: What if this is slow?

Sirupsen:
Thanks!

Besides my comment, this needs a README, and an entry in the main README as well.

hooks/honeybadger/honeybadger.go
agonzalezro: Hi @Sirupsen,

mainly I was following the implementation of the airbrake hook: https://github.com/Sirupsen/logrus/blob/master/hooks/airbrake/airbrake.go#L36

TBH, if we want to return an error there we will to need to wait always for the response and check it. Perhaps it could be a configuration on this hook to launch the notification async:

```
go hook.coon.Error(entry.Message)
```

But this will raise some other issues than can or can not be important (for example, what happens if it errors? and if the project goes out before returning from the subroutine?).

If your main concern is more: what happens if it hangs that what happens if it is slow, I suspect that we could add some timeout.

Thanks for the feedback! It's a good point, but I don't really know how to approach it.

Sirupsen: Excellent point on the Airbrake hook, we should make a timeout for that one as well!

I'm cool with a timeout yep, it's not ideal behaviour (I'd add something like a circuit break for that, if your software is really critical I'd take a look at that, I'm happy to discuss it more too if you like), but enough to ship it for sure.

agonzalezro: What do you think about something like this?

```
func (hook *HoneybadgerHook) Fire(entry *logrus.Entry) error {
	doneChan := make(chan struct{})

	go func(doneChan chan struct{}) {
		err := hook.conn.Error(entry.Message)
		if err != nil {
			entry.Logger.WithFields(logrus.Fields{
				"source": "honeybadger",
				"error":  err,
			}).Warn("Failed to send error to HoneyBadger")
		}
		doneChan <- struct{}{}
	}(doneChan)

	select {
	case <-time.After(5 * time.Second):
		message := "Timed out connecting to HoneyBadger"
		entry.Logger.WithFields(logrus.Fields{
			"source": "honeybadger",
			"error":  errors.New(message),
		}).Warn(message)
	case <-doneChan:
	}

	return nil
}
```

It doesn't look like the honeybadger client that I am using support any kind of timeout.

If you think that the slowness on the hooks is critical, perhaps the timeout should be implemented on the hook call and not inside the hook?

agonzalezro: BTW, the timeout should be configurable when we calling `NewHoneyBadgerHook`.

Sirupsen: Your idea of implementing slowness protection in logrus core to protect against bad hooks is really not bad. It means we could circuit break it as well.

agonzalezro: Do you want me to try to provide you another PR with that change? Are you happy doing the timeout'ing with something like the `time.After` that I posted before?

Sirupsen: The problem is that you can leak goroutines like that.

Sirupsen: But a generic approach in Logrus would have the same issue, since AFAIK you can't forcibly kill a thread in Go? :/

agonzalezro: Yep, I though about that, but in my case sooner or later the http client was going to timeout so I didn't bother about the subroutines.

For killing threads in Go we could investigate an approach using the ctx package? http://dahernan.github.io/2015/02/04/context-and-cancellation-of-goroutines/

aybabtme: @agonzalezro : `context.Context` can't cancel a goroutine anymore than what you can do yourself from within the goroutine. To truly cancel the goroutine, you'd need something like Java's `Thread.interrupt()` or similar. This is not possible in Go by design. When people design their Go packages, they should keep that in mind and add a mean to configure timeouts or abort operations with `context.Context`.

The HoneyBadger client that is imported here is very simplistic and doesn't do that. I'd recommend using (or implementing, looks pretty simple) another one, one that can provide explicit timeouts. The http client used by this package is unexported and there's no way to configure it with custom timeouts.

https://github.com/DavidHuie/gobadger/blob/master/gobadger.go#L158

By default, the `http.Client` has a timeout of `0` which means it never times out.

Sirupsen:
We should at least have a timeout, can you PR to `gobadger` to add an overridable timeout and then add it as a config option to `logrus` with a sane default?

Then I'm OK with shipping this. We can open a ticket about generally preventing runaways.

agonzalezro:
I've implemented my own client. If you think that there is too much logic to be on a hook and you want a separate project for it let me know.

hooks/honeybadger/honeybadger.go
```diff
@@ -0,0 +1,86 @@
...
+// with the Honeybadger API.
+type HoneybadgerHook struct {
+	apiKey     string
+	env        string
+	httpClient *http.Client
+}
+
+func NewHoneyBadgerHook(apiKey, env string, timeout time.Duration) *HoneybadgerHook {
+	return &HoneybadgerHook{
+		apiKey,
```
aybabtme: can you use named keys

hooks/honeybadger/honeybadger.go
```diff
@@ -0,0 +1,86 @@
...
+		req.Header.Set(k, v)
+	}
+
+	resp, err := hook.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != http.StatusCreated {
+		return fmt.Errorf(
+			"The API returned a %d instead an expected %d",
```
aybabtme: can you lowercase the first letter? this makes chained errors look better

hooks/honeybadger/honeybadger.go
```diff
@@ -0,0 +1,86 @@
...
+			resp.StatusCode,
+			http.StatusCreated)
+	}
+	println("CRAP")
```
aybabtme: think this shouldn't be here

hooks/honeybadger/honeybadger.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	for k, v := range map[string]string{
+		"X-API-Key":    hook.apiKey,
+		"Content-Type": "application/json",
+		"Accept":       "application/json",
+	} {
+		req.Header.Set(k, v)
+	}
+
+	resp, err := hook.httpClient.Do(req)
```
aybabtme: the `resp.Body` should be closed

hooks/honeybadger/payload.go
```diff
@@ -0,0 +1,48 @@
...
+	File   string `json:"file"`
+}
+
+type Server struct {
+	EnvironmentName string `json:"environment_name"`
+	Hostname        string `json:"hostname"`
+}
+
+func NewPayload(message, env string) (*Payload, error) {
+	hostname, err := os.Hostname()
```
aybabtme: hostnames arguably don't change during the process' lifetime, it's safe to save the value in the client

hooks/honeybadger/payload.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	payload := &Payload{
+		&Notifier{"logrus", "https://github.com/Sirupsen/logrus", "0.1"},
```
aybabtme: use named keys

hooks/honeybadger/payload.go
```diff
@@ -0,0 +1,48 @@
+package logrus_honeybadger
+
+import "os"
+
+type Payload struct {
```
aybabtme: i'm not sure the types in this file should be exported

hooks/honeybadger/payload.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+type Notifier struct {
+	Name    string `json:"name"`
+	URL     string `json:"url"`
+	Version string `json:"version"`
+}
+
+type Error struct {
+	Backtrace []*Backtrace `json:"backtrace"`
```
aybabtme: this bit isn't used, don't know if its a mistake or on purpose. if on purpose, I'd remove this field and the related `type Backtrace struct`

hooks/honeybadger/payload.go
```diff
@@ -0,0 +1,48 @@
...
+	Message   string       `json:"message"`
+}
+
+// Not used for now
```
aybabtme: i see, can you remove it?

aybabtme:
@Sirupsen build fails because Go 1.2 doesn't have a `Timeout` on `http.Client`. I think we could remove support for Go 1.2.

agonzalezro:
@aybabtme thanks for the feedback it was really useful. I've already changed it.

Sorry as well for the `println` debugging.

hooks/honeybadger/honeybadger.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
```
Sirupsen: Shouldn't this be before the `err != nil`?

Sirupsen:
Needs tests with a mocked HTTP client. Otherwise, this is looking good now!

hooks/honeybadger/honeybadger.go
agonzalezro: If `err == nil` it's possible that `resp` is nil, so `resp.Body` would be `nil.Body`.

agonzalezro:
If you want it tested I would prefer to move the client out to my own honeybadger package and use it here. Do you agree?

hooks/honeybadger/honeybadger.go
Sirupsen: What if it's not `nil` though? I don't know if `err` guarantees that the body doesn't need to be closed.

Sirupsen:
@agonzalezro that sounds good to me :)

hooks/honeybadger/honeybadger.go
aybabtme: nope =P

aybabtme: woups comment got in the wrong order. If `err != nil`, resp is always nil

aybabtme: the only case where you would get a `resp != nil && err != nil` is here:

http://golang.org/src/net/http/client.go?#L400

agonzalezro:
I've moved the client relevant code to a new project under: https://github.com/agonzalezro/goneybadger and properly tested it. Any feedback there will be welcomed.

I've used that new client here so the code is much more simple and the client is properly tested.

hooks/honeybadger/honeybadger.go
```diff
@@ -0,0 +1,44 @@
+package logrus_honeybadger
+
+import (
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/agonzalezro/goneybadger"
```
Sirupsen: A+ for naming

hooks/honeybadger/honeybadger.go
```diff
@@ -0,0 +1,44 @@
...
+	return &HoneybadgerHook{
+		apiKey:  apiKey,
+		env:     env,
+		timeout: timeout,
+	}
+}
+
+func (h *HoneybadgerHook) Fire(entry *logrus.Entry) error {
+	gb := goneybadger.NewWithTimeout(h.apiKey, h.env, h.timeout)
+	err := gb.Notify(entry.Message)
```
Sirupsen: Can you make the API consistent with [the Airbrake hook](https://github.com/Sirupsen/logrus/blob/master/hooks/airbrake/airbrake.go)? I.e. use `error` field if it's there, otherwise fall back to the message.

hooks/honeybadger/honeybadger.go
agonzalezro: Hahaha I am pretty bad at naming, but `gobadger` was already taken :)

aybabtme:
LGTM, the test failure is due to go 1.2 not supporting `http.Client.Timeout`.  

@Sirupsen: I think it would be fair to drop support for go1.2 and remove it from the TravisCI builds. Thoughts?

agonzalezro:
Guys, what are we doing with this? I would like to use the hook from here instead using my fork :)

Sirupsen:
Sorry @agonzalezro, thanks for bumping this. For 1.0 I'm going to move all custom hooks and formatters out of core, and the ones I can't find maintainers for I'll put in my own repositories. Since you wrote this, I'd love for you to be the maintainer of it and turn this PR into simply a README change to reference you this hook in your own repository.

Sirupsen:
bump

agonzalezro:
Sorry for the delay @Sirupsen. I am ok moving this out to another repo with their own `README` and so on. 

Any ETA for version 2 of logrus?

agonzalezro:
Hi guys, I've created the external repo for this hook: https://github.com/agonzalezro/logrus_honeybadger

If you want to close this PR, go for it. If you need something else let me know.

Sorry for being late moving it to another repo, I just forgot about it.

Sirupsen:
@agonzalezro cheers, added to README in https://github.com/Sirupsen/logrus/commit/c8ed7a5eae8c6081fd38f38ebd375b3f75b1119d

Issue #137, opened 2015-02-26
Error value in log.Fields disappears when using JSON formatter


rsterbin:
When passing an error as a field and using the JSON formatter, the error message is omitted:

    package main
    
    import "errors"
    import log "github.com/Sirupsen/logrus"
    
    func main() {
        err := errors.New("Here is an error message")
        log.WithFields(log.Fields{
            "error": err,
        }).Error("Logging without json")
        log.SetFormatter(&log.JSONFormatter{})
        log.WithFields(log.Fields{
            "error": err,
        }).Error("Logging with json")
    }

Results:

    ERRO[0000] Logging without json                          error=Here is an error message
    {"error":{},"level":"error","msg":"Logging with json","time":"2015-02-26T20:32:02Z"}

hasryan:
I ran into the same issue.

From what I can tell the `TextFormatter` handles errors specially (rendering the output of `Error()`): https://github.com/Sirupsen/logrus/blob/c0f7e35ed2e48f188c37581b4b743cf7383f85c6/text_formatter.go#L127

While the `JSONFormatter` does no such thing prior to encoding: https://github.com/Sirupsen/logrus/blob/2835e150d968bd0cd8acaed3a97016342bcb2dea/json_formatter.go#L13-L21

I think it makes sense to do the same thing in the `JSONFormatter`, but for the time being I'm carefully passing `err.Error()` rather than the raw `err`.

rsterbin:
I've got a pull request out for this issue doing something a little different from the text formatter -- would be great to know if that solution meets your needs as well.

Right now my workaround is using a hook to add the error message, because sometimes a custom error has extra fields that are correctly picked up by `json.Marshal` (e.g., `net.OpError`).

dankinder:
Hi there, just wondering what version of go this was an issue one? In golang 1.6 @rsterbin's original issue does not seem to happen:
```
ERRO[0000] Logging without json                          error=Here is an error message
{"error":"Here is an error message","level":"error","msg":"Logging with json","time":"2016-04-27T08:38:59-07:00"}
```

This is relevant because one thing we would like to do is return custom errors that also implement MarshalJSON() as well as Error(), so that they can be treated like plain old `error` types but also be serialized intelligently in logrus messages.

At the moment this particular patch prevents this. If we have a type that implements Error() then the current logrus code seems to see it as an error type and call Error() on it, which stringify's it.

Example:
```go
package main                                                                                                                
                                                                                                                            
[...]

PR #139, opened 2015-02-28
Possible fix for issue #137

rsterbin:
If you pass an error to json.Marshal, you lose the message.  Here's one way the message might get added back in.

### DIFF
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -11,12 +11,10 @@ type JSONFormatter struct{}
 func (f *JSONFormatter) Format(entry *Entry) ([]byte, error) {
 	data := make(Fields, len(entry.Data)+3)
 	for k, v := range entry.Data {
-		switch v.(type) {
-		case error:
-			data[k] = fmt.Sprintf("%s", v)
-		default:
-			data[k] = v
+		if err, ok := v.(error); ok {
+			data[k+".msg"] = fmt.Sprintf("%s", v)
 		}
+		data[k] = v
 	}
 	prefixFieldClashes(data)
 	data["time"] = entry.Time.Format(time.RFC3339)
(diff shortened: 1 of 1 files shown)


Sirupsen:
What's the root of this "bug"? There's nothing in `logrus` that would remove this message as far as I can see, in fact, the Airbrake hook actively uses this field. I haven't tried to reproduce this on my own yet.

`prefixFieldClashes` is what deals with this type of error, but I'm hesitant to merge in this fix without understanding the root cause.

rsterbin:
The root of the bug is how json.Marshal handles errors:

{code}
package main

import (
    "encoding/json"
    "errors"
    "fmt"
)

func main() {
    data := make(map[string]interface{}, 1)
    data["error"] = errors.New("Here is an error message")
    serialized, _ := json.Marshal(data)
    fmt.Println(string(serialized))
}
{code}

Produces:

{code}
{"error":{}}
{code}

Sirupsen:
In that case don't we just need to cast the errors to strings before marshalling, rather than prefixing the fields?

rsterbin:
We could, but it would mean that custom errors might lose information.  For example:

```
package main

import (
    "encoding/json"
    "fmt"
)

type CustomError struct {
    Message string `json:"message"`
    Count   int    `json:"count"`
}

func (e *CustomError) Error() string {
    return fmt.Sprintf("%d - %s", e.Count, e.Message)
}

func main() {
    data := make(map[string]interface{}, 1)
    data["error"] = &CustomError{"Here is an error message", 2}
    if err, ok := data["error"].(error); ok {
        data["error.msg"] = err.Error()
    }
    serialized, _ := json.Marshal(data)
    fmt.Println(string(serialized))
}
```

Produces:

```
{"error":{"message":"Here is an error message","count":2},"error.msg":"2 - Here is an error message"}
```

aybabtme:
I'd prefer not changing the key of the key-value.

Sirupsen:
Fixed in https://github.com/Sirupsen/logrus/commit/8287db79348cd82752b91ee2b7dd5fd274982c65

PR #143, opened 2015-03-04
Text formatter: Do not quote 9

henrikhodne:
I noticed that some UUIDs in our logs were quoted and some weren't, and tracked that down to whether the number 9 was included or not. It seems this was a bug in the conversion from a regexp to an if-statement in 15b296befc8fa20d93c34f6ca2ba934444d27749, so I changed the code to not quote 9s.


Sirupsen:
Excellent catch.

PR #145, opened 2015-03-06
Add integration test to Airbrake hook

mattbostock:
Add a test for the Airbrake hook to:

a) document how the hook is intended to work
b) test that an XML payload is received with the expected message


aybabtme:
lgtm

PR #146, opened 2015-03-06
[#76800642] Add a new Airbrake hook

mattbostock:
Add a new Airbrake hook in `hooks/airbrake2` which addresses some
frustrations I had with the existing Airbrake hook, namely;

- log entries had to include an 'error' field containing an error type,
  which was redundant if you just want to send your log entry messages
  to Airbrake

- credentials and endpoint settings were not scoped to the hook; this
  new hook needs to be invoked using `NewAirbrakeHook`. For example:

        log.Hooks.Add(airbrake.NewHook("https//endpoint.example.com", "api-key-xyz", "production"))

I chose to write a new hook as I couldn't implement the desired
behaviour without making backwards-incompatible changes to the original
hook.

Limitations:

- This hook blocks when sending Airbrake notifications, like the
original.

- Logrus fields are not sent to Airbrake.

Includes a integration simple test to ensure that messages are being
received as part of an XML payload as expected at the specified
endpoint.

### DIFF
--- a/hooks/airbrake2/airbrake.go
+++ b/hooks/airbrake2/airbrake.go
@@ -1,0 +1,46 @@
+package airbrake
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/tobi/airbrake-go"
+)
+
+// AirbrakeHook to send exceptions to an exception-tracking service compatible
+// with the Airbrake API.
+type airbrakeHook struct {
+	APIKey      string
+	Endpoint    string
+	Environment string
+}
+
+func NewHook(endpoint, apiKey, env string) *airbrakeHook {
+	return &airbrakeHook{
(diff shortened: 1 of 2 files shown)


Sirupsen:
I think we should address this in the current one by just falling back to `event.msg` if the error field is not set.

mattbostock:
Thanks @Sirupsen. How do you feel about the change to the interface? I think it's cleaner to keep the credentials within the hook struct.

Sirupsen:
Yeah, I think that's more consistent with the other hooks. :+1:

Sirupsen:
It's hard to document configuring the client like this, and these settings ideally shouldn't be global either—most other APIs don't do that.

mattbostock:
Thanks for the feedback, I've reworked this here: https://github.com/Sirupsen/logrus/pull/149

PR #147, opened 2015-03-06
Remove misleading comment in Airbrake hook

mattbostock:
As far as I can tell, exceptions are always sent regardless of what
`airbrake.Environment` is set to.


aybabtme:
thanks @mattbostock

PR #149, opened 2015-03-10
[#76800642] Rework the Airbrake hook

mattbostock:
Rework the Airbrake hook to:

a) change the interface so that the Airbrake credentials are stored in
an unexported struct, `airbrakeHook`, which is instantiated using the
`NewHook()` method

b) send log entries where no 'error' field is set to Airbrake, using the
`entry.Message` string as the message sent to Airbrake but continue to
allow the passing of error types using the 'error' field

Update the tests accordingly, assuring that the correct message is
received by the Airbrake server.

Also update the examples in the README, which would not have worked with
the previous implementation of the Airbrake hook.

* * *

Also rename the package since using underscores in package names in discouraged:
https://golang.org/doc/effective_go.html#package-names

Given that this package is in a subdirectory of the logrus package,
the name `airbrake` should be sufficiently descriptive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,43 +164,8 @@ You can add hooks for logging levels. For example to send errors to an exception
 tracking service on `Error`, `Fatal` and `Panic`, info to StatsD or log to
 multiple places simultaneously, e.g. syslog.
 
-```go
-// Not the real implementation of the Airbrake hook. Just a simple sample.
-import (
-  log "github.com/Sirupsen/logrus"
-)
-
-func init() {
-  log.AddHook(new(AirbrakeHook))
-}
-
-type AirbrakeHook struct{}
-
-// `Fire()` takes the entry that the hook is fired for. `entry.Data[]` contains
-// the fields for the entry. See the Fields section of the README.
-func (hook *AirbrakeHook) Fire(entry *logrus.Entry) error {
-  err := airbrake.Notify(entry.Data["error"].(error))
-  if err != nil {
-    log.WithFields(log.Fields{
-      "source":   "airbrake",
-      "endpoint": airbrake.Endpoint,
-    }).Info("Failed to send error to Airbrake")
-  }
-
-  return nil
-}
-
-// `Levels()` returns a slice of `Levels` the hook is fired for.
-func (hook *AirbrakeHook) Levels() []log.Level {
-  return []log.Level{
-    log.ErrorLevel,
-    log.FatalLevel,
-    log.PanicLevel,
-  }
-}
-```
-
-Logrus comes with built-in hooks. Add those, or your custom hook, in `init`:
+Logrus comes with [built-in hooks](hooks/). Add those, or your custom hook, in
+`init`:
 
 ```go
 import (
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ func init() {
 
   // Use the Airbrake hook to report errors that have Error severity or above to
   // an exception tracker. You can create custom hooks, see the Hooks section.
-  log.AddHook(&logrus_airbrake.AirbrakeHook{})
+  log.AddHook(airbrake.NewHook("https://example.com", "xyz", "development"))
 
   // Output to stderr instead of stdout, could also be a file.
   log.SetOutput(os.Stderr)
@@ -211,7 +211,7 @@ import (
 )
 
 func init() {
-  log.AddHook(new(logrus_airbrake.AirbrakeHook))
+  log.AddHook(airbrake.NewHook("https://example.com", "xyz", "development"))
 
   hook, err := logrus_syslog.NewSyslogHook("udp", "localhost:514", syslog.LOG_INFO, "")
   if err != nil {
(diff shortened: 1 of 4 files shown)


hooks/airbrake/airbrake_test.go
```diff
@@ -1,57 +1,133 @@
...
+var (
+	noticeError = make(chan NoticeError, 1)
+)
+
+// TestLogEntryMessageReceived checks if invoking Logrus' log.Error
+// method causes an XML payload containing the log entry message is received
+// by a HTTP server emulating an Airbrake-compatible endpoint.
+func TestLogEntryMessageReceived(t *testing.T) {
+	log := logrus.New()
+	ts := startAirbrakeServer(t)
```
Sirupsen: I like the pattern of 

```go
WithAirbrakeServer(t, function(endpoint) {
# ..
})
```

More.

hooks/airbrake/airbrake_test.go
```diff
@@ -1,57 +1,133 @@
...
+
+var (
+	noticeError = make(chan NoticeError, 1)
+)
+
+// TestLogEntryMessageReceived checks if invoking Logrus' log.Error
+// method causes an XML payload containing the log entry message is received
+// by a HTTP server emulating an Airbrake-compatible endpoint.
+func TestLogEntryMessageReceived(t *testing.T) {
+	log := logrus.New()
```
Sirupsen: You should probably set the output to a discard util?

hooks/airbrake/airbrake.go
```diff
@@ -1,51 +1,51 @@
-package logrus_airbrake
+package airbrake
```
Sirupsen: :+1:

hooks/airbrake/airbrake.go
```diff
@@ -1,51 +1,51 @@
...
 
 import (
+	"errors"
+	"fmt"
+
 	"github.com/Sirupsen/logrus"
 	"github.com/tobi/airbrake-go"
```
Sirupsen: Does this get confusing when they both use `package airbrake`? Should we import this under a different name?

Sirupsen:
This is looking pretty great.

hooks/airbrake/airbrake.go
mattbostock: I considered that, but found it wasn't a problem. When referring to `tobi`'s package, it's clear that we're referring to an external package.

Also, `tobi`'s package is no longer imported when using the Airbrake hook since the Airbrake credentials are now scoped to the `AirbrakeHook` struct (which could also have been confusing).

hooks/airbrake/airbrake_test.go
mattbostock: Do you mean so that stdout is discarded during the tests? I found that output is only shown if the tests failed, which was quite useful in determining what might have caused the failure.

I couldn't see any prior art for this in other tests in `logrus`, though I'm open to your suggestions.

hooks/airbrake/airbrake_test.go
mattbostock: I'm not sure I agree. The syntax above looks more like Ruby/Rspec and would entail unnecessary nesting of code, which is doesn't feel like idiomatic Go IMHO.

mattbostock:
Rebased and force pushed to fix the test failure (in the `examples/` directory).

README.md
```diff
@@ -164,43 +164,8 @@ You can add hooks for logging levels. For example to send errors to an exception
...
-func (hook *AirbrakeHook) Levels() []log.Level {
-  return []log.Level{
-    log.ErrorLevel,
-    log.FatalLevel,
-    log.PanicLevel,
-  }
-}
-```
-
-Logrus comes with built-in hooks. Add those, or your custom hook, in `init`:
```
Sirupsen: Did you remove those because there are now so many other examples and this example is no longer very good?

hooks/airbrake/airbrake.go
```diff
@@ -1,51 +1,51 @@
...
-		}).Warn("Exceptions sent to Airbrake must have an 'error' key with the error")
-		return nil
+// with the Airbrake API.
+type airbrakeHook struct {
+	APIKey      string
+	Endpoint    string
+	Environment string
+}
+
+func NewHook(endpoint, apiKey, env string) *airbrakeHook {
```
burke: Not that I think `golint` should be the ultimate source of truth on good practice, because this usecase does make sense, but I'd like to point out that `golint` will complain about returning a private type from a public function here.

mattbostock: Interesting point. It did occur to me that perhaps `airbrakeHook` should be exported but I like that it enforces use the `NewHook()` method.

README.md
mattbostock: I removed it for two reasons:

a) it seemed unnecessary to repeat the code for the Airbrake hook in the README, which could quickly become obsolete. The comment "Not the real implementation of the Airbrake hook. Just a simple sample." was a red flag here, it seems to make more sense just to link to the actual code.

b) this PR removed the old implementation shown in the `README` and so the example would be obsolete

PR #152, opened 2015-03-16
hooks: Add BugSnag hook

burke:
r: @Sirupsen 
cc: @graemej

No rush on this, worry about it when you're back from SRECon.

### DIFF
--- a/hooks/bugsnag/bugsnag.go
+++ b/hooks/bugsnag/bugsnag.go
@@ -1,0 +1,53 @@
+package logrus_bugsnag
+
+import (
+	"github.com/Sirupsen/logrus"
+	"github.com/bugsnag/bugsnag-go"
+)
+
+// BugsnagHook sends exceptions to an exception-tracking service compatible
+// with the Bugsnag API. Before using this hook, you must call
+// bugsnag.Configure().
+//
+// Entries that trigger an Error, Fatal or Panic should now include an "error"
+// field to send to Bugsnag
+type BugsnagHook struct{}
+
+// Fire forwards an error to Bugsnag. Given a logrus.Entry, it extracts the
+// implicitly-required "error" field and sends it off.
+func (hook *BugsnagHook) Fire(entry *logrus.Entry) error {
+	if entry.Data["error"] == nil {
+		entry.Logger.WithFields(logrus.Fields{
+			"source": "bugsnag",
+		}).Warn("Exceptions sent to Bugsnag must have an 'error' key with the error")
+		return nil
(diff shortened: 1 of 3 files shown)


graemej:
LGTM, as @Sirupsen pointed out on the original it assumes that the Bugsnag config is using the async delivery model (which it does by default) so that it doesn't time out.

hooks/bugsnag/bugsnag.go
```diff
@@ -0,0 +1,53 @@
+package logrus_bugsnag
```
Sirupsen: Just call it `bugsnag`, we need to rename the other ones.

hooks/bugsnag/bugsnag.go
```diff
@@ -0,0 +1,53 @@
...
+func (hook *BugsnagHook) Fire(entry *logrus.Entry) error {
+	if entry.Data["error"] == nil {
+		entry.Logger.WithFields(logrus.Fields{
+			"source": "bugsnag",
+		}).Warn("Exceptions sent to Bugsnag must have an 'error' key with the error")
+		return nil
+	}
+
+	err, ok := entry.Data["error"].(error)
+	if !ok {
```
Sirupsen: Can you duplicate the logic from the new Airbrake hook (https://github.com/Sirupsen/logrus/pull/149) so it defaults to `msg` if `error` is not there?

Sirupsen:
Two comments otherwise LGTM

Sirupsen:
Can you add this to the README as well?

burke:
Sure. Redoing based on the new airbrake hook PR.

hooks/bugsnag/bugsnag.go
burke: I'm not so sure. It works in the `airbrake` hook because `tobi/airbrake-go` only has three fields to set and the hook can own that configuration, but `bugsnag/bugsnag-go` supports a lot more configuration that the calling code should own. That means the caller already has to import a package named `bugsnag`. If we make this one `bugsnag` too, they'll have to choose one to rename.

burke: I mean, I could remap `bugsnag.Configure()` and `bugsnag.Configuration` into this namespace, but that feels pretty ugly.

burke:
Made a few changes.

hooks/bugsnag/bugsnag.go
Sirupsen: Ok, then this is fine :+1:

PR #154, opened 2015-03-19
improved: move level check from entry to logger and bail out faster

imkira:
Hi there.

Thank you for logrus, it's a wonderful logger.
I was benchmarking loggers, as mentioned in https://github.com/imkira/go-loggers-bench
and I found I could make logrus bail-out faster (but more importantly, lesser allocations) when the baseline log-level is higher than the level being logged.
This is a modest attempt to improve it.

I didn't do it in this PR, but I also tried to do the same with WithFields, but given the current API it is very difficult to do it.
My idea was to invert

log.WithFields(...).Info(...)

and allow something like:

log.Info(...).WithFields(...)

or alternatively something like

log.WithLevel(Info).WithFields(...).Log(...)

Using this kind of API it would be possible to return "blackhole" entries/loggers when the level does not satisfy the level condition.
What do you think of it?

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -65,11 +65,15 @@ func (logger *Logger) WithFields(fields Fields) *Entry {
 }
 
 func (logger *Logger) Debugf(format string, args ...interface{}) {
-	NewEntry(logger).Debugf(format, args...)
+	if logger.Level >= DebugLevel {
+		NewEntry(logger).Debugf(format, args...)
+	}
 }
 
 func (logger *Logger) Infof(format string, args ...interface{}) {
-	NewEntry(logger).Infof(format, args...)
+	if logger.Level >= InfoLevel {
+		NewEntry(logger).Infof(format, args...)
+	}
 }
 
 func (logger *Logger) Printf(format string, args ...interface{}) {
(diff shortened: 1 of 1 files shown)


entry.go
```diff
@@ -110,6 +110,10 @@ func (entry *Entry) Debug(args ...interface{}) {
 	}
 }
 
+func (entry *Entry) debug(args ...interface{}) {
+	entry.log(DebugLevel, fmt.Sprint(args...))
+}
```
Sirupsen: Aren't these methods only added to avoid another integer comparison?

imkira: Yeah. I could remove them and call the original (publicly exported) ones. What do you think?

Sirupsen: Yeah, I don't think this extra code is necessary but I'm happy to accept the other part of the patch. Please rebase when you do that as well so it's just one commit. Thanks!

imkira:
@Sirupsen done. Let me know if its good now

Sirupsen:
Thank you! It'll be in 0.7.1

Sirupsen:
You should re-run the benchmarks :) I'll happily accept other performance patches like this that don't break the API.

imkira:
Thanks!
https://github.com/imkira/go-loggers-bench
Updated. Still sucks for JSON though but let's think this later.

PR #155, opened 2015-03-19
Add example for setting log level on New() logger

### LINKED ISSUE #153: SetLevel is only defined for the default logger, not on Instances
The Readme suggest this should be possible.
```
You can set the logging level on a `Logger`, then it will only log entries with
that severity or anything above it:
```
...but the code shows  SetLevel to be an exported function without a receiver.
```
exported.go
30:// SetLevel sets the standard logger level.
31:func SetLevel(level Level) {
```

Logging levels are missing from the stdlib log which is why I picked up logrus, am I missing something?

GuyMorgan:
closes #153

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,14 @@ that severity or anything above it:
 ```go
 // Will log anything that is info or above (warn, error, fatal, panic). Default.
 log.SetLevel(log.InfoLevel)
+
+// Set the Level directly for a New() logger, you may use the ParseLevel() helper.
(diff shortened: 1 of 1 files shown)


aybabtme:
Thanks @GuyMorgan and sorry for the long delay here. I think this is fairly straightforward to anyone reading the docs, and the README covers the important basics. So I'll close for now. Feel free to reopen if you feel strongly about.t

PR #157, opened 2015-03-25
Make airbrake hook non-blocking

gravis:
Logging should not wait for airbrake to be notified.

### DIFF
--- a/hooks/airbrake/airbrake.go
+++ b/hooks/airbrake/airbrake.go
@@ -6,7 +6,7 @@ import (
 	"os"
 
 	"github.com/Sirupsen/logrus"
-	"github.com/tobi/airbrake-go"
+	"github.com/airbrake/gobrake"
 )
 
 // Set graylog.BufSize = <value> _before_ calling NewHook
@@ -15,45 +15,40 @@ var BufSize uint = 1024
 // AirbrakeHook to send exceptions to an exception-tracking service compatible
 // with the Airbrake API.
 type airbrakeHook struct {
-	APIKey      string
-	Endpoint    string
-	Environment string
-	entryChan   chan *logrus.Entry
+	Airbrake   *gobrake.Notifier
+	noticeChan chan *gobrake.Notice
 }
 
-func NewHook(endpoint, apiKey, env string) *airbrakeHook {
+func NewHook(projectID int64, apiKey, env string) *airbrakeHook {
+	airbrake := gobrake.NewNotifier(projectID, apiKey)
+	airbrake.SetContext("environment", env)
 	hook := &airbrakeHook{
-		APIKey:      apiKey,
-		Endpoint:    endpoint,
-		Environment: env,
-		entryChan:   make(chan *logrus.Entry, BufSize),
+		Airbrake:   airbrake,
+		noticeChan: make(chan *gobrake.Notice, BufSize),
 	}
 	go hook.fire()
 	return hook
 }
 
 func (hook *airbrakeHook) Fire(entry *logrus.Entry) error {
-	hook.entryChan <- entry
+	var notifyErr error
+	err, ok := entry.Data["error"].(error)
+	if ok {
+		notifyErr = err
+	} else {
+		notifyErr = errors.New(entry.Message)
+	}
+	notice := hook.Airbrake.Notice(notifyErr, nil, 3)
+	hook.noticeChan <- notice
 	return nil
 }
 
 // fire sends errors to airbrake when an entry is available on entryChan
 func (hook *airbrakeHook) fire() {
 	for {
-		entry := <-hook.entryChan
-		airbrake.ApiKey = hook.APIKey
-		airbrake.Endpoint = hook.Endpoint
-		airbrake.Environment = hook.Environment
+		notice := <-hook.noticeChan
 
-		var notifyErr error
-		err, ok := entry.Data["error"].(error)
-		if ok {
-			notifyErr = err
-		} else {
-			notifyErr = errors.New(entry.Message)
-		}
-
-		if err = airbrake.Notify(notifyErr); err != nil {
+		if err := hook.Airbrake.SendNotice(notice); err != nil {
 			fmt.Fprintf(os.Stderr, "Failed to send error to Airbrake: %v\n", err)
 		}
 
--- a/hooks/airbrake/airbrake.go
+++ b/hooks/airbrake/airbrake.go
@@ -25,6 +25,11 @@ func NewHook(endpoint, apiKey, env string) *airbrakeHook {
 }
 
 func (hook *airbrakeHook) Fire(entry *logrus.Entry) error {
+	go hook.fire(entry) // Don't block logging
+	return nil
+}
+
+func (hook *airbrakeHook) fire(entry *logrus.Entry) error {
 	airbrake.ApiKey = hook.APIKey
 	airbrake.Endpoint = hook.Endpoint
 	airbrake.Environment = hook.Environment
--- a/hooks/airbrake/airbrake.go
+++ b/hooks/airbrake/airbrake.go
@@ -3,10 +3,14 @@ package airbrake
 import (
 	"errors"
 	"fmt"
+	"os"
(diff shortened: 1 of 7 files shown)


gravis:
I'm not sure it's the right solution btw. Having a AddAsyncHook is probably better since we leave the decision to the user.

hooks/airbrake/airbrake.go
```diff
@@ -25,6 +25,11 @@ func NewHook(endpoint, apiKey, env string) *airbrakeHook {
 }
 
 func (hook *airbrakeHook) Fire(entry *logrus.Entry) error {
+	go hook.fire(entry) // Don't block logging
```
Sirupsen: I think a better way to do this is that `NewHook` spawns a go routine that reads off a channel, avoiding spawning an unlimited number of goroutines. The channel should be buffered, and if `fire()` can't write to it it prints an error like the current one.

gravis: Good point!
I'll update that today.
Thanks

gravis:
Done!
Please note that tobi/airbrake-go **never** times-out on sending the error to airbrake, so I think this is a timebomb.
The official github.com/airbrake/gobrake has a default timeout of 10s, which is a lot better I think.
I strongly suggest to use the official repo instead.

Sirupsen:
Ya, let's switch.

gravis:
The api is slightly different:

```go
airbrake.ApiKey = hook.APIKey
airbrake.Endpoint = hook.Endpoint
airbrake.Environment = hook.Environment
``` 

from github.com/airbrake/gobrake

```go
airbrake = gobrake.NewNotifier(projectId, apiKey)
airbrake.SetContext("environment", "production")
```

There 2 options:

* extract the `projectId` from the current `endpoint` param. (API stays the same, but the end isn't "gobrake" idiomatic, and the user could wonder why the endpoint is not working as expected)
* modify the API of NewHook() to take a projectId directly. (API changes dangerously, as we still accept a string, so previous code with endpoint will still compile but won't work)

Any ideas here?

gravis:
The api is slightly different:

```go
airbrake.ApiKey = hook.APIKey
airbrake.Endpoint = hook.Endpoint
airbrake.Environment = hook.Environment
``` 

from github.com/airbrake/gobrake

```go
airbrake = gobrake.NewNotifier(projectId, apiKey)
airbrake.SetContext("environment", "production")
```

There 2 options:

* extract the `projectId` from the current `endpoint` param. (API stays the same, but the end isn't "gobrake" idiomatic, and the user could wonder why the endpoint is not working as expected)
* modify the API of NewHook() to take a projectId directly. (API changes dangerously, as we still accept a string, so previous code with endpoint will still compile but won't work)

Any ideas here?

gravis:
for the record, tobi's version is also using a old endpoint by default:

    Endpoint    = "https://api.airbrake.io/notifier_api/v2/notices"

whereas gobrake is using:

    "https://airbrake.io/api/v3/projects/%d/notices?key=%s",

Sirupsen:
Let's break the API. It's for the better long-term. Please update the README as well :+1:

gravis:
Thanks for the insight, will do that today.
I also agree it's better on the long term, especially before a 1.0 release.

gravis:
Okkkk, so I just discover gobrake only works with `http.Request`:

```go
func (n *Notifier) Notify(e interface{}, req *http.Request) error {
	notice := n.Notice(e, req, 3)
	if err := n.SendNotice(notice); err != nil {
		log.Printf("gobrake failed (%s) reporting error: %v", err, e)
		return err
	}
	return nil
}
```

That's lame... So we'll stick to tobi's version, and I'll create a PR in @tobi 's repo to make sure http requests don't stack to death if airbrake api is unreachable.

Sorry for the mess.

Sirupsen:
Can't you just pass it `nil`?

gravis:
sounds like we can:

https://github.com/airbrake/gobrake/blob/master/notice.go#L55

will try that and let you know

gravis:
There you go!

gravis:
The build is failing because of the import path:

```
# github.com/Sirupsen/logrus/hooks/airbrake
hooks/airbrake/airbrake_test.go:128: cannot use notice.Errors[0] (type gobrake.Error) as type *gobrake.Error in send
```

But I can ensure they are working:

```
ok      github.com/Sirupsen/logrus      0.021s
?       github.com/Sirupsen/logrus/examples/basic       [no test files]
?       github.com/Sirupsen/logrus/examples/hook        [no test files]
ok      github.com/Sirupsen/logrus/formatters/logstash  0.030s
ok      github.com/Sirupsen/logrus/hooks/airbrake       0.022s
ok      github.com/Sirupsen/logrus/hooks/bugsnag        0.027s
ok      github.com/Sirupsen/logrus/hooks/papertrail     0.017s
ok      github.com/Sirupsen/logrus/hooks/sentry 0.019s
ok      github.com/Sirupsen/logrus/hooks/syslog 0.016s
> echo $?
0
```

mattbostock:
:+1: to making the hook non-blocking, I think this is important.

I too like the idea of using the official Airbrake client, however that presents a problem in that:

- it hardcodes the Airbrake endpoint which means that it doesn't work with API-compatible products such as [Errbit](http://errbit.github.io/errbit/)

- only supports version 3 of the Airbrake API (Errbit supports version 2)

gravis:
I don't have a solution for this.
Only possibility is to create a pull request in tobi's repo to use a different http client (like the official gobrake). I have no idea if @tobi would accept something like that?
https://github.com/airbrake/gobrake/blob/master/notifier.go#L17

Thanks

tobi:
Sure. Whatever is best

On Thursday, April 9, 2015, Philippe Lafoucrière <notifications@github.com>
wrote:

> I don't have a solution for this.
> Only possibility is to create a pull request in tobi's repo to use a
> different http client (like the official gobrake). I have no idea if @tobi
> <https://github.com/tobi> would accept something like that?
> https://github.com/airbrake/gobrake/blob/master/notifier.go#L17
>
> Thanks
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Sirupsen/logrus/pull/157#issuecomment-91243654>.
>


-- 

- tobi
CEO Shopify

gravis:
Thanks for the input, I'll propose a pull request then in the coming days.

Sirupsen:
@gravis do you have an update on this?

gravis:
I'm fighting emergencies currently, and didn't find time to update tobi's version yet :(

Sirupsen:
@gravis ping :)

gravis:
@Sirupsen, we're waiting for tobi to merge:
https://github.com/tobi/airbrake-go/pull/16
I would suggest to use my original version with the official airbrake (gobrake) implementation if you want to be solved faster.
Thanks

Sirupsen:
I'll send him an email and ask him to add me as contributor.

Sirupsen:
On the other hand, internally we don't have any use for a custom endpoints anymore going forward.

How about you rebase this and we get it in?

Sirupsen:
Also are you interested in taking over maintenance of the Airbrake hook under your personal Github to keep as much as possible out of core?

gravis:
Sure. How would you see that? We're merging this, and then we create the new repo?

gravis:
Ok, will do that today or tomorrow then.

Sirupsen:
We'll close this PR and merge a change that points to your repository in the README :)

tobi:
I'm a crappy dependency for software projects these days. Happy to add someone else to that repo so you don't have to wait for me.

- tobi
CEO Shopify

(mobile)

Sirupsen:
@tobi feel free to add me. We may go ahead with something else though.

gravis:
Ok, the change has been merged in tobi's repo.
We can close this issue, since the official gobrake isn't going to be used. 
I wonder now if we won't create 2 separate pluging repos:
- one legacy airbrake (the current version)
- one official, using airbrake v3 api. 
We'll propose some pull requests this week.

PR #158, opened 2015-03-26
allow custom time format string in TextFormatter

scarletmeow:
- fixes examples in README.md that incorrectly state usage of
  RFC3339Nano format instead of RFC3339

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,11 +37,13 @@ attached, the output is compatible with the
 [logfmt](http://godoc.org/github.com/kr/logfmt) format:
 
 ```text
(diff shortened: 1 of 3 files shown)


Sirupsen:
Nice, thank you!

scarletmeow:
I'll like to use this to fix an upstream issue at https://github.com/docker/docker/issues/11563, could we cut  a new release of logrus?

Sirupsen:
Definitely

Sirupsen:
v0.7.2 tagged

PR #160, opened 2015-03-27
Add graylog hook to Readme

gravis:
I have updated the list to be a table instead. It's much more readable than the previous version, and used in a lot of places (ex: https://github.com/codegangsta/negroni/blob/master/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,31 +189,18 @@ func init() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


gravis:
Don't trust travis, it's drunk

Sirupsen:
A+ thanks

PR #162, opened 2015-04-04
Allow configuration of timestamp layout

### LINKED ISSUE #161: time formats should use nanosecond resolution
The log formatters, such as json and text, are using the RFC3339. For example, from `JSONFormatter`:

```
data["time"] = entry.Time.Format(time.RFC3339)
```

This only provides 1 second granularity on timestamps. These formatter should either use `RFC3339Nano` by default or allow the user to configure it.

Something like this would be acceptable:

```
type JSONFormatter struct {
    TimestampLayout string
}
```

Let me know if this is okay and I can drop a PR.

stevvooe:
This just allows one to configure the timestamp layout and defaults to using `time.RFC3339Nano` by default.

Closes #161.

Signed-off-by: Stephen J Day <stephen.day@docker.com>

### DIFF
--- a/formatter.go
+++ b/formatter.go
@@ -1,5 +1,9 @@
 package logrus
 
+import "time"
+
+const DefaultTimestampFormat = time.RFC3339
+
 // The Formatter interface is used to implement a custom Formatter. It takes an
 // `Entry`. It exposes all the fields, including the default ones:
 //
--- a/formatters/logstash/logstash.go
+++ b/formatters/logstash/logstash.go
@@ -3,19 +3,27 @@ package logstash
 import (
 	"encoding/json"
 	"fmt"
+
 	"github.com/Sirupsen/logrus"
-	"time"
 )
 
 // Formatter generates json in logstash format.
 // Logstash site: http://logstash.net/
 type LogstashFormatter struct {
 	Type string // if not empty use for logstash type field.
+
+	// TimestampFormat sets the format used for timestamps.
+	TimestampFormat string
 }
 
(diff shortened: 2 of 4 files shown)


Sirupsen:
There's already `TimestampFormat` on the `text_formatter`, can you rebase and rename the other fields to have consistent naming? I think we should keep the default to not be nano, but overridable on each formatter.

stevvooe:
@Sirupsen Updated to comply with feedback.

Small nit: the standard library refers to these as "layouts" (see http://golang.org/pkg/time/#Time.Format), since they are used for parsing and formatting. The PR follows the logrus standards but just wanted to make sure the break with the language from the standard library was conscious.

Sirupsen:
It's definitely not, but do you think it's worth breaking for?

stevvooe:
@Sirupsen It looks like it has only been out for a week or so. We vendor the library, since it is marked as unstable but this may affect others. Now would be the time to fix it. Let me know and we can make the move in this PR.

As an aside, I am totally sorry I missed this other change. I did a search before submitting the ticket and PR and didn't find anything.

stevvooe:
@Sirupsen Ping!

If possible, I'd like to get this change in the release of docker/distribution. It would be great if we could avoid a fork. Let me know if you want those changes or not and we can get his merged.

Sirupsen:
A+ thank you

Sirupsen:
Part of 0.7.3

Sirupsen:
(Sorry for the response time, I've been away to container summit in NYC)

PR #168, opened 2015-04-17
Added Raygun hook.

squirkle:
Hook providing [Raygun](http://raygun.io) support for logrus.

I tried to follow the patterns in the Airbrake hook/Airbrake hook tests as closely as possible.


squirkle:
A [dependent library that I manage](https://github.com/SDITools/goraygun) had not been updated to work with the hook above.  Re-running the build should cause tests to pass.

Sirupsen:
@squirkle I'm trying to move in a direction where the hooks are maintained by people who actually use them rather than me—which meant out of this repository. Do you mind creating a repository under your own user to maintain this, and then add a README entry for it here?

Sirupsen:
bump

squirkle:
Thanks for the reminder :)

Moved this into its own repo and added the README here.

Sirupsen:
With adding README, I meant removing the `raygun` directory and just adding it to the table in the [main README](https://github.com/sirupsen/logrus#hooks)

squirkle:
This should be what you need.  Let me know if any further changes are needed.

PR #170, opened 2015-04-22
default logs to stderr

aybabtme:
This makes me rage always, I think it's worth fixing.
@Sirupsen 

Pretty good rationale on when to use stdout/stderr by @jstorimer:
http://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.8
+
+logrus: defaults to stderr instead of stdout
+
 # 0.7.3
 
 formatter/\*: allow configuration of timestamp layout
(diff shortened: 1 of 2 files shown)


Sirupsen:
I definitely agree with this, and it's been brought up multiple times. I'm unsure whether or not to ship this before or after 1.0

aybabtme:
I've bumped the version to `0.8` for that purpose. Any reason why you would prefer making the change in `1.0`?

Sirupsen:
@stevvooe @icecrime would this have any heartbreaking implications for Docker?

stevvooe:
@Sirupsen For the registry (docker/distribution), this would change the output location of the logs. This may affect infrastructure consuming those log messages if they don't also consume `stderr`. I don't see an issue with this. Logs should go to `stderr`.

PR #171, opened 2015-04-23
assertify changed behavior and consider float64 != int

aybabtme:
@Sirupsen all the tests and PRs are broken because of this. ima merge right away since it's just a fix to the tests


Sirupsen:
Thx

PR #183, opened 2015-05-22
Added special field for *http.Request to Sentry hook

evalphobia:
This PR will attempt to add a special field for http.Request, and send detailed request information.
(see: http://sentry.readthedocs.org/en/latest/developer/interfaces/index.html#sentry.interfaces.http.Http)

When add *http.Request to `http_request` field of logrus data, request information is set to the data.

### DIFF
--- a/hooks/sentry/sentry_test.go
+++ b/hooks/sentry/sentry_test.go
@@ -61,9 +61,12 @@ func TestSpecialFields(t *testing.T) {
 			t.Fatal(err.Error())
 		}
 		logger.Hooks.Add(hook)
+
+		req, _ := http.NewRequest("GET", "url", nil)
 		logger.WithFields(logrus.Fields{
-			"server_name": server_name,
-			"logger":      logger_name,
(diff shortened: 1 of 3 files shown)


hooks/sentry/sentry_test.go
```diff
@@ -61,9 +61,12 @@ func TestSpecialFields(t *testing.T) {
...
 		logger.Hooks.Add(hook)
+
+		req, _ := http.NewRequest("GET", "url", nil)
 		logger.WithFields(logrus.Fields{
-			"server_name": server_name,
-			"logger":      logger_name,
+			"server_name":  server_name,
+			"logger":       logger_name,
+			"http_request": req,
 		}).Error(message)
```
evalphobia: `http_request` is added in this test, but not asserted in the below part.
Its attribute is `sentry.interfaces.Http` in the json data, not like other flat fields.
So `raven.Packet{}` can not include it as `raven.Packet.Interface`, and the data is removed.

PR #186, opened 2015-05-26
Fix: Remove extraneous log-level checks in certain log functions

mapuri:
This change makes calls to Fatalf() behave in a similar manner as call to Fatal() when Logger level is set to `PanicLevel` by causing it to exit in both cases. It also simplifies the code in other calls a bit by removing some extra log-level checks and hence having them at fewer places.

I hope this is helpful and I didn't miss something.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -153,15 +153,11 @@ func (entry *Entry) Panic(args ...interface{}) {
 // Entry Printf family functions
 
 func (entry *Entry) Debugf(format string, args ...interface{}) {
-	if entry.Logger.Level >= DebugLevel {
-		entry.Debug(fmt.Sprintf(format, args...))
-	}
+	entry.Debug(fmt.Sprintf(format, args...))
 }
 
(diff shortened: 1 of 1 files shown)


Sirupsen:
Thanks for your PR! However, the extra checks are there to avoid unnecessary `fmt.Sprint*`-allocations when the logging level is higher than the log statement.

mapuri:
I see, thanks for clarifying. But then the underlying issue still remains i.e. Fatalf() will not exit while Fatal() will if log level is set to PanicLevel.

Would it make sense to do one of the following:
- add os.Exit(1) in Fatalf() as well but then we need to take care of Fataln(), Panicf() etc
- Or, replace fmt.Sprint* calls with just args and let eventual function calls take care of the args?

I can update the PR or open a new one if this makes sense.

Sirupsen:
Yeah, please create another PR that fixes that issue alone. Thanks!

mapuri:
ok, I created #187 to just fix Fatal* routines. PTAL

PR #187, opened 2015-05-26
Fix Fatalf() and Fataln() to exit irrespective of log level

mapuri:
Fixing Fatalf() and Fataln() to behave similar to Fatal() based on discussion in #186


Sirupsen:
Thanks!

PR #192, opened 2015-06-02
Update Sentry hook to allow capture of stacktraces at set Logrus Level

miloconway:
This change would allow users of the Sentry hook to selectively enable stacktraces for a given logging level.
- Update hook to allow optional stacktrace capture
- Update hook test to handle response when raven client compresses / base64 encode the payload
- Implement packet object in hook test to allow for json deserialization of stacktrace

### DIFF
--- a/hooks/sentry/sentry.go
+++ b/hooks/sentry/sentry.go
@@ -58,11 +58,27 @@ type SentryHook struct {
 	// Timeout sets the time to wait for a delivery error from the sentry server.
 	// If this is set to zero the server will not wait for any response and will
 	// consider the message correctly sent
-	Timeout time.Duration
+	Timeout                 time.Duration
+	StacktraceConfiguration stacktraceConfiguration
 
-	client          *raven.Client
-	levels          []logrus.Level
-	stacktraceLevel *logrus.Level
+	client *raven.Client
+	levels []logrus.Level
+}
+
+// stacktraceConfiguration allows for configuring stacktraces
+type stacktraceConfiguration struct {
+	// whether stacktraces should be enabled
+	Enable bool
+	// the level at which to start capturing stacktraces
+	Level logrus.Level
+	// how many stack frames to skip before stacktrace starts recording
+	Skip int
+	// the number of lines to include around a stack frame for context
+	Context int
+	// the prefixes that will be matched against the stack frame.
+	// if the stack frame's package matches one of these prefixes
+	// sentry will identify the stack frame as "in_app"
+	InAppPrefixes []string
 }
 
 // NewSentryHook creates a hook to be added to an instance of logger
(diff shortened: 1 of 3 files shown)


hooks/sentry/sentry.go
```diff
@@ -98,6 +104,9 @@ func (hook *SentryHook) Fire(entry *logrus.Entry) error {
 	if req, ok := getAndDelRequest(d, "http_request"); ok {
 		packet.Interfaces = append(packet.Interfaces, raven.NewHttp(req))
 	}
+	if hook.stacktraceLevel != nil && entry.Level <= *hook.stacktraceLevel {
+		packet.Interfaces = append(packet.Interfaces, raven.NewStacktrace(5, 0, nil))
```
Sirupsen: What are these hard coded numbers?

hooks/sentry/sentry.go
```diff
@@ -60,8 +60,9 @@ type SentryHook struct {
 	// consider the message correctly sent
 	Timeout time.Duration
 
-	client *raven.Client
-	levels []logrus.Level
+	client          *raven.Client
+	levels          []logrus.Level
+	stacktraceLevel *logrus.Level
```
Sirupsen: Should this really be a pointer?

hooks/sentry/sentry.go
miloconway: I'll abstract them out to configurable properties and add comments.

hooks/sentry/sentry.go
miloconway: I wanted a way to allow people to skip enabling stacktrace without introducing another variable... will rethink the approach.

miloconway:
@Sirupsen would you mind seeing if there are other concerns? Thanks

jrdnull:
This would be very useful, is there anything holding this back from being merged?

Sirupsen:
Sorry for the delay here. Maintenance of all the hooks which I usually don't use is not very productive. Is any of you interested in putting it under a personal repository and maintain it?

@gigaroby @evalphobia @awonak @miloconway <-- people who have commits on it.

evalphobia:
@Sirupsen @gigaroby @awonak @miloconway 
I usually use sentry, so if you guys okay I gonna maintain it.

I made the repository here.
https://github.com/evalphobia/logrus_sentry

Sirupsen:
If everyone is OK with that, let's move this PR there—and then submit another PR to logrus to remove the Sentry hook and add an entry to the README.

miloconway:
sounds good

Sirupsen:
@evalphobia can you own orchestrating that? :)

Sirupsen:
\cc @henrikbjorn

PR #193, opened 2015-06-08
Hook for providing default fields to be added to all logged events

redmoxie:
This hook allows initializing default Fields that are added to every logged event. This is especially useful where you are aggregating logs from multiple applications (logstash).

### DIFF
--- a/hooks/defaultfields/README.md
+++ b/hooks/defaultfields/README.md
@@ -1,0 +1,31 @@
+# Default Fields Hooks for Logrus
+
+Add default fields to be logged with every event. Useful when your logs are aggregated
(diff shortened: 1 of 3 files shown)


Sirupsen:
You can use `WithFields` for this to set them on all future entries:

```go
logger := logrus.WithFields(logrus.Fields{"default": "field"})
```

I don't accept hooks and formatters upstream anymore, but encourage people to maintain them out of tree. If you still think it has value, you can convert this PR into adding it to one of the tables in the README. Thanks!

redmoxie:
Thx. Not sure how I missed that this was in the API already.

Issue #185, opened 2015-05-26
DisableColors: true in TextFormatter does more than disable colors


jsyrjala:
```
package main

import (
	"github.com/Sirupsen/logrus"
)

var log = logrus.New()

func main() {
  log.Formatter = &logrus.TextFormatter{FullTimestamp: true, DisableColors: true}
  log.Info("has no colors")
  log.Formatter = &logrus.TextFormatter{FullTimestamp: true}
  log.Info("has colors")
}
```

Prints this, the first line has no color, the second line has color.
```
time="2015-05-26T21:19:08+03:00" level=info msg="has no colors" 
INFO[2015-05-26T21:19:08+03:00] has colors  
```

I was expecting the output text stay the same and only colors to be missing on the first line

PR #195, opened 2015-06-10
Made format consistent when DisableColors=true (issue #185)

russmack:
Fixes issue #185, making the formatting consistent when DisableColors has been set to true.


aybabtme:
the 'colorless' format is in machine readable logfmt, which is why they differ. this PR can't be merged as it will break users of this feature.

PR #198, opened 2015-06-18
Add trace log level 

Xiol:
Hi,

I've added a 'trace' log level for when debug just isn't enough. When hooking into systems that don't support a trace level it'll push them out to debug instead (e.g. syslog, sentry).

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,9 @@ With `log.Formatter = new(logrus.JSONFormatter)`, for easy parsing by logstash
 or Splunk:
 
 ```json
+{"level":"trace","msg":"The tiniest detail","planck_length":
+1.61619926e-35,"time":"2014-03-10 09:15:29.562264112 -0400 EDT"}
+
 {"animal":"walrus","level":"info","msg":"A group of walrus emerges from the
 ocean","size":10,"time":"2014-03-10 19:57:38.562264131 -0400 EDT"}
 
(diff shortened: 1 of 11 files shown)


Sirupsen:
What is your use-case? This is not something that comes up often and for the majority of people would be confusing.

JoeBruner:
TL/DR; Trace is for finding the problem, DEBUG is for fixing the problem

The trace log level can be very useful, especially since error messages generally don't have a stack trace associated with them.  I constantly find error messages where I'm like "Well, that's great, but where did that come from?".  That usually happens when the error is thrown from an exterior library where I don't have control over what's in the message.  

Even more useful though, is when you have a program running in the enterprise and you're getting spruious results.  If you have a bug in your program that you didn't happen to have a logging statement around, it can be nearly impossible to hunt down where exactly a problem might be happening.  Because of configuration mgmt, uptime,  and other fun corporate things, it might be prohibitive if not impossible to recompile the binary to add debugging statements in.  By having a trace log level, you can add a log statement (at a minimum) to the beginning/end of every function.  In normal use, you have this turned off so you aren't littering your logs with useless messages.  If a problem pops up and you don't know where it could be coming from, you can simply change the logging level to TRACE which doesn't require a recompile.  You can then sift through the logs and narrow down where the problem might be happening.

This is different than simply doing the same thing with DEBUG instead of TRACE.  Let's say that you knew whereabouts the problem might be happening and you've added DEBUG statements to try and figure out what's going wrong.  If you marked the prolog/eplilog of every function with DEBUG statements, it would make it a real pain in the rear to sift through the logs to find the debugging statements that you cared about.

Xiol:
@JoeBruner has put it very well, thanks.

In my case, I have an application which does a lot of text-based network traffic. Sometimes I want to see the traffic that is being sent to/from the network. Having this always enabled at the debug level can make it impossible to find useful information from the logs because of the sheer amount of data being generated.

I'd like to be able to dump the network data to the 'trace' level, so that I can enable this extra level of debugging when necessary (or rather, the people debugging my application can do so). At the moment I'm just using a compile-time flag in the library that turns on a load of fmt.Printf() statements - I could obviously improve this situation with a little more effort but I'd like the functionality to be present in the logging library rather than re-implementing portions of my log setup to provide this.

I'd already submitted this PR after I noticed one of Logrus' goals was to be 'completely API compatible with the standard library logger', so I'm not sure if adding this extra level is in line with that goal or not.

zyio:
Wading in to say that I'm on board with the above two commenters, TRACE has come in handy in the past and would be appreciated here.

aybabtme:
Don't think we'll be moving forward with this. Sorry for the long delay.

PR #200, opened 2015-06-20
Don't emit a msg if there is none to emit

freeformz:
This cleans up the case where everything is already encoded in fields.

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -84,7 +84,9 @@ func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
 			f.appendKeyValue(b, "time", entry.Time.Format(timestampFormat))
 		}
 		f.appendKeyValue(b, "level", entry.Level.String())
-		f.appendKeyValue(b, "msg", entry.Message)
+		if entry.Message != "" {
+			f.appendKeyValue(b, "msg", entry.Message)
+		}
 		for _, key := range keys {
 			f.appendKeyValue(b, key, entry.Data[key])
 		}


Sirupsen:
For the sake of simplicity in parsers I'd like to adhere to a fixed schema here, especially I believe that's key in the JSON formatter. If you need this functionality, you can create a custom formatter.

freeformz:
@Sirupsen I can see that for the JSON formatter, but for the text formatter it's just noise that gets in the way (IMO). Maybe you would accept a patch just against the text formatter ?

Sirupsen:
can you show an example of some situations where you like it and what it'd look like? with the default formatting it can look pretty weird.

freeformz:
@Sirupsen The most common grumble I have when I tell people to "just use logrus" (at least internally) is  when we construct a set of Fields that contain all the key/value data and we just want to log that. This is probably the most common case / way to use logrus for us as all of our logging is structured logfmt style logging.

A somewhat contrived example:

```go
func FooBar(v int) {
  ctx := logrus.Fields{"fn":"FooBar", "v": v, "start": time.Now()}
  defer func() {   ctx.WithField("end", time.Now()}.Info() }()
  ctx.Info()
  if err := Foozle(v); err != nil {
    ctx.WithField("err",err).Warn("Error calling Foozle")
  }
}
```

All but the Warn() will add a blank `msg=` to their lines via the text formatter. I could write a custom formatter but it would be a copy of the default with the above patch removing the `msg=` part applied to it. IMO it's just noise for the text formatter when the msg is empty and I'd removing it to be the default.

Sirupsen:
@freeformz that's a very valid argument, can you rebase this with a commit just doing it for the text formatter and I will merge?

freeformz:
@Sirupsen rebased and only for the text formatter now.

Thanks!

Issue #112, opened 2015-01-16
Cross compiling doesn't work


haarts:
I'm using [goxc](https://github.com/laher/goxc) to cross compile my project. Doing so now results in:

```
../../Sirupsen/logrus/text_formatter.go:28: undefined: IsTerminal
[goxc:xc] 2015/01/16 20:50:30 'go' returned error: exit status 2
[goxc:xc] 2015/01/16 20:50:30 Error: exit status 2
[goxc:xc] 2015/01/16 20:50:30 Have you run `goxc -t` for this platform (amd64p32,nacl)???
```

This error is repeated for every platform.

I'm using c6a969a0de65e41b666c5dd015b3988b47cd940b and Go 1.4 on Ubuntu 14.04.
Running `goxc` in the `examples/basic` directory is enough to reproduce the problem. I have ran `goxc -t` of course and I know this worked previous to introducing logrus.

Sirupsen:
What are you cross compiling to?

haarts:
All of them. :) Goxc compiles to the craziest platforms. But none of them succeed at the moment.

Sirupsen:
Did you ever find a solution to this?

haarts:
I have had some luck with:
```
$ goxc -bc="windows,linux,darwin"
```
This specifies 'build constraints'. And yielded binaries for the three platforms. I'm unsure why constraining the platforms works since an unconstrained builds builds for these platforms as well.

The bug seems more `goxc` related, this issue I'd close.

labkode:
The cross compilation only fails for these architectures.
 ```
[goxc:xc] 2015/04/07 21:32:25 Have you run `goxc -t` for this platform (386,netbsd)???
[goxc:xc] 2015/04/07 21:32:25 Have you run `goxc -t` for this platform (amd64,netbsd)???
[goxc:xc] 2015/04/07 21:32:25 Have you run `goxc -t` for this platform (arm,netbsd)???
[goxc:xc] 2015/04/07 21:32:25 Have you run `goxc -t` for this platform (386,dragonfly)???
[goxc:xc] 2015/04/07 21:32:25 Have you run `goxc -t` for this platform (amd64,dragonfly)???
[goxc:xc] 2015/04/07 21:32:25 Have you run `goxc -t` for this platform (386,plan9)???
[goxc:xc] 2015/04/07 21:32:25 Have you run `goxc -t` for this platform (amd64,solaris)???
[...]

PR #203, opened 2015-06-26
Add support for DragonflyBSD and NetBSD.

cgag:
This merges all the bsd (including darwin) support into terminal_bsd.go

Cross-compiling works for all of them in that they all compile, but I've only
actually tested the binary on dragonflybsd.  I don't see why this wouldn't
work on the others since I don't think they'd compile if they didn't
support TIOCGETA and Termios, but just a heads up.


cgag:
Fixes some issues in https://github.com/Sirupsen/logrus/issues/112.  My comment there explains why I didn't add nacl and plan9.

Sirupsen:
Released in 0.8.3

PR #204, opened 2015-06-27
[TextFormatter] Speed up (~40%). Fprintf is changed to buffer.Write*

noxiouz:
As Fprintf is slower than buffer.WriteString, it's replaced to faster call.

Before:
```
BenchmarkSmallTextFormatter	  300000	      5841 ns/op	  14.89 MB/s
BenchmarkLargeTextFormatter	   50000	     24215 ns/op	  11.85 MB/s
BenchmarkSmallColoredTextFormatter	  300000	      5071 ns/op	  27.02 MB/s
BenchmarkLargeColoredTextFormatter	   50000	     27192 ns/op	  20.34 MB/s
```
Now:
```
BenchmarkSmallTextFormatter	  300000	      3883 ns/op	  22.40 MB/s
BenchmarkLargeTextFormatter	  100000	     14305 ns/op	  20.06 MB/s
BenchmarkSmallColoredTextFormatter	  300000	      5114 ns/op	  26.79 MB/s
BenchmarkLargeColoredTextFormatter	   50000	     26918 ns/op	  20.54 MB/s
```

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -131,21 +131,28 @@ func needsQuoting(text string) bool {
 	return true
 }
 
-func (f *TextFormatter) appendKeyValue(b *bytes.Buffer, key, value interface{}) {
-	switch value.(type) {
+func (f *TextFormatter) appendKeyValue(b *bytes.Buffer, key string, value interface{}) {
+
+	b.WriteString(key)
+	b.WriteByte('=')
+
+	switch value := value.(type) {
 	case string:
-		if needsQuoting(value.(string)) {
-			fmt.Fprintf(b, "%v=%s ", key, value)
+		if needsQuoting(value) {
+			b.WriteString(value)
 		} else {
-			fmt.Fprintf(b, "%v=%q ", key, value)
+			fmt.Fprintf(b, "%q", value)
 		}
 	case error:
-		if needsQuoting(value.(error).Error()) {
-			fmt.Fprintf(b, "%v=%s ", key, value)
+		errmsg := value.Error()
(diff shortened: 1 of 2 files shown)


noxiouz:
if it's ok, I can apply the same optimization to the colored formatter.

text_formatter.go
```diff
@@ -131,21 +131,28 @@ func needsQuoting(text string) bool {
 	return true
 }
 
-func (f *TextFormatter) appendKeyValue(b *bytes.Buffer, key, value interface{}) {
+func (f *TextFormatter) appendKeyValue(b *bytes.Buffer, key string, value interface{}) {
+
+	b.WriteString(key)
+	b.WriteByte('=')
+
 	switch value.(type) {
```
aybabtme: if you switch this to:

```
switch v := value.(type) {
case string:
    // in this 'case', `v` is a string so you don't need any more `.(string)`
}
```

aybabtme:
The idea LGTM, try with the switch I suggest, I guess avoiding the type case will squeeze out a bit more performance.

noxiouz:
@aybabtme updated. The result is improved (+ ~8%):
```
BenchmarkSmallTextFormatter	  500000	      3540 ns/op	  24.57 MB/s
BenchmarkLargeTextFormatter	  100000	     13763 ns/op	  20.85 MB/s
```

Sirupsen:
Thanks @noxiouz!

Sirupsen:
Released in 0.8.3

PR #209, opened 2015-07-05
Print struct keys for struct log fields

pdf:
Can be useful for debugging, saves a lot of referring to source code


Sirupsen:
What's the properties for strings, numbers, maps etc? Does +v only affect structs?

pdf:
Yeah, maps already print keys, no change for basic types either.  Godoc is [here](https://golang.org/pkg/fmt/).

Sirupsen:
Released in 0.8.3

PR #215, opened 2015-07-09
Allow sentry hook to be created using an initialized raven client

allanglen:
Allows using a initialized raven client to create a logrus hook.  Avoids having to create an additional raven client when wanting to use raven directly (ex. raven.CapturePanic).

### DIFF
--- a/hooks/sentry/README.md
+++ b/hooks/sentry/README.md
@@ -46,6 +46,39 @@ hook, err := logrus_sentry.NewWithTagsSentryHook(YOUR_DSN, tags, levels)
 
 ```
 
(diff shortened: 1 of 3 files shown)


Sirupsen:
@allanglen @henrikbjorn would any of you be interested in taking over maintenance of this hook out of tree? I'm trying to push all the hooks out to make people who actually use them take over maintenance. :)

henrikbjorn:
Sure i could do that.

Sirupsen:
@henrikbjorn cool. Create a repo and submit a PR here to remove it and add it to the README. Thanks! :)

Issue #217, opened 2015-07-17
Data race at github.com/Sirupsen/logrus.(*TextFormatter).Format()


pengzhai:
Similar to the issue reported in https://github.com/Sirupsen/logrus/issues/184

WARNING: DATA RACE
Read by goroutine 13:
  github.com/Sirupsen/logrus.(*TextFormatter).Format()
      /mnt/jenkins/workspace/test-driver-status-log/Godeps/_workspace/src/github.com/Sirupsen/logrus/text_formatter.go:74 +0x424
  github.com/Sirupsen/logrus.(*Entry).Reader()
      /mnt/jenkins/workspace/test-driver-status-log/Godeps/_workspace/src/github.com/Sirupsen/logrus/entry.go:41 +0xaa
  github.com/Sirupsen/logrus.(*Entry).log()
      /mnt/jenkins/workspace/test-driver-status-log/Godeps/_workspace/src/github.com/Sirupsen/logrus/entry.go:84 +0x373
  github.com/Sirupsen/logrus.(*Entry).Error()
      /mnt/jenkins/workspace/test-driver-status-log/Godeps/_workspace/src/github.com/Sirupsen/logrus/entry.go:135 +0xb0
  github.com/Sirupsen/logrus.(*Entry).Errorf()

PR #218, opened 2015-07-17
Fix data race issue in TextFormatter. Fix for https://github.com/Siru…

pengzhai:
…psen/logrus/issues/217.

Fix for the issue we just reported in https://github.com/Sirupsen/logrus/issues/217

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -78,7 +78,7 @@ func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
 		timestampFormat = DefaultTimestampFormat
 	}
 	if isColored {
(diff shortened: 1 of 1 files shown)


pengzhai:
We have verified this fix on our internal CI system.

noxiouz:
@pengzhai printColored uses timestamp format too. The signature should be changed to pass timestampFormat, otherwise it will use an empty f.timestampFormat

pengzhai:
Good catch, @noxiouz! Fixed.

pengzhai:
Any additional comment, anyone?

Sirupsen:
sorry been on vacation

PR #224, opened 2015-07-28
Add function for creating a Sentry hook when you already have a client

henrikbjorn:
In our case we configure sentry with different contexts etc. by using its default client. it would be great if we could use that with our logrus setup aswell.


henrikbjorn:
Closing as #215 already have this.

PR #225, opened 2015-07-31
Fix Entry log level

Arkan:
This PR comes to fix #208 and #214. 

#208 fixed an issue with log levels but introduced an issue that I tried to fix in #214.
However the fix involved too many objects creations, and therefore was refused.

This new fix tries to mitigate the impact of objects creations by only creating them when the log level changes. This shouldn't impact in any way the basic usage of logrus and should also fix contextual loggers.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -32,7 +32,8 @@ func NewEntry(logger *Logger) *Entry {
 	return &Entry{
 		Logger: logger,
 		// Default is three fields, give a little extra room
-		Data: make(Fields, 5),
+		Data:  make(Fields, 5),
+		Level: logger.Level,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


aybabtme:
@Arkan can you fix the conflicts?

Arkan:
Yes, sure. Conflicts are now solved.

entry.go
```diff
@@ -67,33 +68,42 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
...
 // This function is not declared with a pointer value because otherwise
 // race conditions will occur when using multiple goroutines
 func (entry Entry) log(level Level, msg string) {
-	entry.Time = time.Now()
-	entry.Level = level
-	entry.Message = msg
+	var cxtEntry *Entry = &entry
+
+	// If the log level if different from the current stored log level
+	// Then we will create a copy of Entry to avoid messing with the original log level
```
aybabtme: `entry Entry` is a non-pointer-receiver so it's already a copy and won't change the original one

entry.go
```diff
@@ -67,33 +68,42 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
...
+	return &Entry{Logger: entry.Logger, Data: data, Level: entry.Level}
 }
 
 // This function is not declared with a pointer value because otherwise
 // race conditions will occur when using multiple goroutines
 func (entry Entry) log(level Level, msg string) {
-	entry.Time = time.Now()
-	entry.Level = level
-	entry.Message = msg
+	var cxtEntry *Entry = &entry
```
aybabtme: AFAIK all the changes in this func are meant to accomodate the unnecessary copy, so unless I'm missing something, the func should be left intact.

aybabtme:
other than that, lgtm

aybabtme:
closing due to inactivity.

PR #226, opened 2015-08-06
Fix race condition in entry log method

### LINKED ISSUE #216: Date race
Hi,

Just using logrus in my project and running with "go run -race" I see this output:

```
==================
WARNING: DATA RACE
Read by goroutine 8:
  github.com/Sirupsen/logrus.(*TextFormatter).Format()
      /home/angel/go/projects/src/github.com/Sirupsen/logrus/text_formatter.go:76 +0x42c
  github.com/Sirupsen/logrus.(*Entry).Reader()
      /home/angel/go/projects/src/github.com/Sirupsen/logrus/entry.go:41 +0xaa
  github.com/Sirupsen/logrus.(*Entry).log()
      /home/angel/go/projects/src/github.com/Sirupsen/logrus/entry.go:84 +0x373
  github.com/Sirupsen/logrus.(*Entry).Info()
      /home/angel/go/projects/src/github.com/Sirupsen/logrus/entry.go:119 +0xb0
  github.com/Sirupsen/logrus.(*Logger).Info()
      /home/angel/go/projects/src/github.com/Sirupsen/logrus/logger.go:122 +0x148
  bitbucket.org/soyangel/pricealert/log.Infof()
      /home/angel/go/projects/src/bitbucket.org/soyangel/pricealert/log/log.go:23 +0x162
  bitbucket.org/soyangel/pricealert/scraper.UpdateProductWithNotifications()
      /home/angel/go/projects/src/bitbucket.org/soyangel/pricealert/scraper/scraper.go:44 +0x26e
  main.func·001()
      /home/angel/go/projects/src/bitbucket.org/soyangel/pricealert/cmd/scrape/main.go:56 +0xa5

Previous write by goroutine 7:
  github.com/Sirupsen/logrus.(*TextFormatter).Format()
      /home/angel/go/projects/src/github.com/Sirupsen/logrus/text_formatter.go:77 +0x44c
  github.com/Sirupsen/logrus.(*Entry).Reader()
      /home/angel/go/projects/src/github.com/Sirupsen/logrus/entry.go:41 +0xaa
  github.com/Sirupsen/logrus.(*Entry).log()
      /home/angel/go/projects/src/github.com/Sirupsen/logrus/entry.go:84 +0x373
  github.com/Sirupsen/logrus.(*Entry).Info()
      /home/angel/go/projects/src/github.com/Sirupsen/logrus/entry.go:119 +0xb0
  github.com/Sirupsen/logrus.(*Logger).Info()
      /home/angel/go/projects/src/github.com/Sirupsen/logrus/logger.go:122 +0x148
  bitbucket.org/soyangel/pricealert/log.Infof()
      /home/angel/go/projects/src/bitbucket.org/soyangel/pricealert/log/log.go:23 +0x162
  bitbucket.org/soyangel/pricealert/scraper.UpdateProductWithNotifications()
      /home/angel/go/projects/src/bitbucket.org/soyangel/pricealert/scraper/scraper.go:44 +0x26e
  main.func·001()
      /home/angel/go/projects/src/bitbucket.org/soyangel/pricealert/cmd/scrape/main.go:56 +0xa5
```
In my file log.go loggers are created like this:

```go
 var log = logrus.New()
 var errorLog = logrus.New()
     
 func init() {
     log.Out = os.Stdout
     errorLog.Out = os.Stderr
 } 
```

marcosnils:
Use a copy of the Entry structure to avoid possible race conditions.

Fixes #216

Signed-off-by: Marcos Lilljedahl <marcosnils@gmail.com>

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -100,7 +100,7 @@ func (entry Entry) log(level Level, msg string) {
 	// panic() to use in Entry#Panic(), we avoid the allocation by checking
 	// directly here.
 	if level <= PanicLevel {
-		panic(entry)
+		panic(&entry)
 	}
 }
 
--- a/entry.go
+++ b/entry.go
@@ -70,6 +70,8 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 	return &Entry{Logger: entry.Logger, Data: data}
 }
 
+// This function is not declared with a pointer value because otherwise
+// race conditions will occur when using multiple goroutines
 func (entry Entry) log(level Level, msg string) {
 	entry.Time = time.Now()
 	entry.Level = level
--- a/entry.go
+++ b/entry.go
@@ -70,12 +70,12 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 	return &Entry{Logger: entry.Logger, Data: data}
 }
 
-func (entry *Entry) log(level Level, msg string) {
+func (entry Entry) log(level Level, msg string) {
 	entry.Time = time.Now()
 	entry.Level = level
 	entry.Message = msg
 
-	if err := entry.Logger.Hooks.Fire(level, entry); err != nil {
+	if err := entry.Logger.Hooks.Fire(level, &entry); err != nil {
 		entry.Logger.mu.Lock()
 		fmt.Fprintf(os.Stderr, "Failed to fire hook: %v\n", err)
 		entry.Logger.mu.Unlock()


noxiouz:
I think it's better to solve it by redesigning the architecture. For example, don't pass Timestamp, level, message via Entry. Unfortunately, it would break API. 
The current solution decreases performance. Protecting of Entry fields by Logger mutex looks like an overhead. May be it's better to add a mutex inside Entry struct and use it to protect Entry?

marcosnils:
@noxiouz I agree, I didn't want to introduce a change which would break backwards compatibility. I don't think performance would be much affected by this change however, as previously almost all the `log` method was synchronized anyways. 

I'd hope this change to be merged soon so race problems can be fixed and then thing about a new architecture about how to accomplish this.

noxiouz:
@marcosnils the heaviest part of this method is formatting inside ```entry.Reader()```. Right now it's impossible to format two unrelated entries at the same time. Also writting to a file would block formatting of entries too. Of course, only the benchmark can prove it :) I don't insist on a benchmark as I'm not a collaborator, but I think it's important enough to be investigated. 
This logger is widely used, and it's a great responsibility to maintain a performance level. As the changes could affect Docker performance for example (and lots of projects).

noxiouz:
@marcosnils  Anothe thought that both formatters (json and text) copy Entry during formatting. So passing a copy to them can solve the race. And the formatters would get rid of copying Entry. It doesn't affect the API, I think.

marcosnils:
@noxiouz that would solve the problem because the race detector is complaining about multiple goroutines accessing to the same memory space while setting Entry variables in the `log` method

```
 	entry.Time = time.Now()
 	entry.Level = level
 	entry.Message = msg
```

noxiouz:
@marcosnils yes. So we can just copy the Entry at thebeginning of Log method, set Time, level, msg and remove a redundant copy inside the formatters.

marcosnils:
@noxiouz just rebased with your propositions. What do you think?

noxiouz:
@marcosnils I think it's good.

freeformz:
@marcosnils You may want to add a comment as to why log takes an Entry vs an *Entry.

marcosnils:
@freeformz done!.

aybabtme:
thanks @marcosnils

PR #231, opened 2015-08-10
Allow hooks to suppress messages

joshk0:
By returning a canned error that's exported in the logrus package, hooks
can choose to suppress messages entirely.

This creates two considerations:

- If your hooks are publishing to some upstream service, the order in
  which hooks are added matters. An upstream hook can squelch a downstream
  hook.

- Looking at this from the opposite perspective, hooks upstream of the
  one that choose not to suppress will be fully fired before a
  downstream hook which chooses to squelch the message, suppressing the
  overall output as well as further downstream hooks.

In general, this means that the best practice here is that if you have a
hook that may suppress a log message, you should add it as early as
possible, ahead of the hooks that don't care about suppressing messages.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -2,11 +2,14 @@ package logrus
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"time"
 )
+
+var Suppress = errors.New("suppressed log message")
 
 // An entry is the final or intermediate Logrus logging entry. It contains all
 // the fields passed with WithField{,s}. It's finally logged when Debug, Info,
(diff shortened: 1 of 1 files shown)


Sirupsen:
Can you explain your use-case?

Sirupsen:
Closing due to lack of activity.

joshk0:
Sorry for the lack of communication here. We used this to implement a log throttling hook. Which means that we insert it first, and if the same message gets logged above a threshold number of times, we reject all similar messages with the same msg value. Then, when we observe a new message being logged, we log that we throttled that msg value a certain number of times and reset the counter.

PR #233, opened 2015-08-11
Make Entry.log panic with the message rather than the opaque entry

### LINKED ISSUE #232: Panicf() loses the error message
Calling the Panicf() function results in a panic with unhelpful message along the lines of `panic: (*logrus.Entry) (0xa2cda0,0xc20812a040)`. This seems to be coming from https://github.com/Sirupsen/logrus/blob/master/entry.go#L105 which omits the msg argument presumably carrying the actual error message.

mkobetic:
Fixes #232


mkobetic:
Hm, I guess that's intentional. It seems if you want to use the Panicf API, you better use the hooks as well to send panics to external services.

PR #239, opened 2015-08-31
Add logger wide context field handling

joelacrisp:
This small patch provides the ability to add "Fields" at the logger level which will be inherited by any Entry created by that logger. Essentially it just pre-populates the Entry's fields with the global ones. This is useful for things like overall context such as "environment":"dev" etc

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -28,6 +28,8 @@ type Logger struct {
 	Level Level
 	// Used to sync writing to the log.
 	mu sync.Mutex
+	// Contains all the fields set by the user.
+	Data Fields
 }
 
 // Creates a new logger. Configuration should be set by changing `Formatter`,
@@ -48,6 +50,7 @@ func New() *Logger {
 		Formatter: new(TextFormatter),
 		Hooks:     make(LevelHooks),
 		Level:     InfoLevel,
+		Data: make(Fields, 5),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


logger.go
```diff
@@ -204,3 +207,20 @@ func (logger *Logger) Panicln(args ...interface{}) {
...
+
+// Add a single field to the global log context
+func (logger *Logger) SetContextField(key string, value interface{}) {
+	logger.SetContextFields(Fields{key: value})
+}
+
+// Add a map of fields to the global log context
+func (logger *Logger) SetContextFields(fields Fields) {
+	data := Fields{}
+	for k, v := range logger.Data {
```
noxiouz: Why don't you do like this?
```go
for k, v := range fields {
	logger.Data[k] = v
}
```

joelacrisp: Ah, yes, I was pre-setting the size of the combined map. For some reason I removed that. Which do you think is better : data:=make(Fields,len(fields)+len(logger.Data)) then my original code or the loop above?

Sirupsen:
You can use an Entry just as well as a Logger. Just do:

```go
contextLogger := logrus.WithFields(..)
contextLogger.WithFields(..).Info("Logs with all the fields")
```

joelacrisp:
This is different, it lets you set the context for the logger, and pass it around as type Log not type Entry, so that you can have a consistent set of broader context for each *Log* not each Entry

Sirupsen:
Are you talking about passing the logger around in your own application and the interface of an Entry and Log being different?

joelacrisp:
Yes, particularly for 3rd party products. We do use the pattern you suggest above extensively in our own code (it's in the examples, after all) but for other code which expects something exporting the log API I added the global context. 

The other reason we pass a Logger not an entry is so that some sub-systems may add and remove hooks, something else which I don't believe is available on Entry

PR #244, opened 2015-09-04
The HasFields Interface

akutz:
The patch adds a new type named HasFields, an interface that enables a
type to control the manner in which it is formatted as a Logrus field
value.

For example, take the following type:

    type Person struct {
        Name  string
        Alias string
    }

    p := &Person{"Bruce", "Batman"}

If p were given to Logrus as a field value with an associated key of
"hero" then it would be formatted as such (given a text formatter):

    hero="&{Bruce Batman}"

However, the Person type may know how it wants to be formatted when it
comes to be logged as a field value, and this interface will provide
that advantage.

    func (p *Person) Fields() map[string]interface{} {
        return map[string]interface{} {"name" : p.Name, "alias" : p.Alias}
    }

Now when a Person instance is provided to Logrus as a field value with
an associated key of hero it would be formatted as such (given a text
formatter):

    hero.name=Bruce hero.alias=Batman...

The JSON formatter will not nest field names as there is no precedent
for such behavior and consumers of the data expect a direct
serialization of a series of key/value pairs without additional object
structure. Given then the same example above, the JSON formatter would
produce the following output:

    { "hero.name" : "bruce", "hero.alias" : "batman" ... }

The values in the map returned by the Fields function are also
considered with regards to whether they implement the HasFields
interface, enabling multiple levels of field data to find its way into
the log statement in an ordered fashion.

There may be cases where it is not desireable to have all keys in a
possible key path emitted along with a value. To that end this
interface defines the Flatten function. If Flatten returns a true
value, it indicates that the top-level key should be removed when
emitting the field data. For example:

    func (p *Person) Flatten() {
        return true
    }

Because the Flatten function returns true, if the Person instance p is
logged as a field value with an associated key of "hero" this is what
will actually be emitted by the text formatter:

    name=Bruce alias=Batman...

The key supplied along with a value is stripped from the emitted field
names when the Flatten function returns true.

Additionally, it's possible to bypass the explicit naming of fields
altogether. The HasFields interface requires types to implement a
function called UseTypeFields that returns a flag indicating whether
or not a type's public fields should be used instead of any explicit
key/value pairs returned by the Fields function.

    func (p *Person) Fields() map[string]interface{} {
        return nil
    }

    func (p *Person) UseTypeFields() bool {
        return true
    }

With the above redefinition of the Fields function for the Person type,
and the UseTypeFields function returning true, the text and JSON
formatters will produce the same output as the earlier examples without
having to specify any explicit fields.

Finally, the HasFields interface defines one more function,
ExceptFields. This function returns an array of strings -- a list of
field names to not allow to be emitted as part of the log message.

    func (p *Person) ExceptFields() []string {
        return []string {"name"}
    }

Because the ExceptFields function returns an array containing the field
called "name," this is what will be emitted by the text formatter:

    alias=Batman...

It doesn't matter whether the field is discovered because UseTypeFields
returns true or if the field is explicitly returned via the Fields
function. If the name of the field (case insensitive) is in the array
returned by the ExceptFields function the field's value will not be
emitted.

### DIFF
--- a/has_fields.go
+++ b/has_fields.go
@@ -1,0 +1,182 @@
+package logrus
+
+import (
+	"fmt"
+	"reflect"
+	"regexp"
+	"strings"
+)
+
+// HasFields is an interface that enables a type to control the manner in
+// which it is formatted as a Logrus field value.
+//
+// For example, take the following type:
+//
+//     type Person struct {
+//         Name  string
+//         Alias string
+//     }
+//
+//     p := &Person{"Bruce", "Batman"}
+//
+// If p were given to Logrus as a field value with an associated key of "hero"
+// then it would be formatted as such (given a text formatter):
+//
+//     hero="&{Bruce Batman}"
+//
+// However, the Person type may know how it wants to be formatted when it comes
+// to be logged as a field value, and this interface will provide that
+// advantage.
+//
+//     func (p *Person) Fields() map[string]interface{} {
+//         return map[string]interface{} {"name" : p.Name, "alias" : p.Alias}
+//     }
+//
+// Now when a Person instance is provided to Logrus as a field value with an
+// associated key of hero it would be formatted as such (given a text
+// formatter):
+//
+//     hero.name=Bruce hero.alias=Batman...
+//
+// The JSON formatter will not nest field names as there is no precedent for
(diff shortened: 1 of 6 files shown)


akutz:
Closing to reopen as different PR.

PR #245, opened 2015-09-07
Godoc Documentation

wallclockbuilder:
Add godoc documentation.
Before:
Before

After
After


aybabtme:
thanks @wallclockbuilder !

Issue #243, opened 2015-09-04
Hooks should probably fire asynchronously.


dradtke:
Especially if an application is using a hook that posts to a remote service, the act of posting to that service shouldn't slow down response times. This would require refactoring the hook firing mechanism to return a channel of potential errors rather than an error, and then fire hooks in their own goroutines.

alecha:
+1

Sirupsen:
I'm of the opinion this is up to the hook. This is too surprising to be a default.

Sirupsen:
I'm going to close this for now and watch the discussion on this closely and possibly bring it up again.

PR #246, opened 2015-09-07
Experimental implementation of async hooks.

dradtke:
This is my first crack at implementing a solution for #243. The API is certainly up for discussion, but I designed it to be fairly unobtrusive and very straightforward.

The idea behind my approach is that each hook is considered either synchronous, or asynchronous. By default, any value implementing `Hook` is considered synchronous, and if it *also* implements `AsyncHook` (effectively a no-op interface), then the hook is considered asynchronous instead and its `Fire()` method will be run in a separate goroutine.

A new `HooksDone` channel is added to `Entry` as a notification mechanism for knowing that all hooks have finished processing. It will always provide a value, but if you only have synchronous hooks registered, or none at all, then it will provide a value almost immediately. Errors are logged directly to standard error for both synchronous and asynchronous hooks.

As an aside, my original idea was to make *all* hooks run asynchronously, but that broke the hook's ability to modify the entry's data, so any hook that wants to do that must remain synchronous.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -29,6 +29,10 @@ type Entry struct {
 
 	// Message passed to Debug, Info, Warn, Error, Fatal or Panic
 	Message string
+
+	// A receive-only notification channel for knowing when all hooks have
+	// finished firing.
+	HooksDone <-chan struct{}
 }
 
 func NewEntry(logger *Logger) *Entry {
@@ -87,17 +91,13 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 }
 
 // This function is not declared with a pointer value because otherwise
-// race conditions will occur when using multiple goroutines
-func (entry Entry) log(level Level, msg string) {
+// race conditions will occur when using multiple goroutines.
+func (entry Entry) log(level Level, msg string, hooksDone chan struct{}) {
 	entry.Time = time.Now()
 	entry.Level = level
 	entry.Message = msg
 
-	if err := entry.Logger.Hooks.Fire(level, &entry); err != nil {
-		entry.Logger.mu.Lock()
-		fmt.Fprintf(os.Stderr, "Failed to fire hook: %v\n", err)
-		entry.Logger.mu.Unlock()
-	}
+	entry.Logger.Hooks.Fire(level, &entry, hooksDone)
 
 	reader, err := entry.Reader()
 	if err != nil {
@@ -124,7 +124,9 @@ func (entry Entry) log(level Level, msg string) {
 
 func (entry *Entry) Debug(args ...interface{}) *Entry {
 	if entry.Logger.Level >= DebugLevel {
-		entry.log(DebugLevel, fmt.Sprint(args...))
+		hooksDone := make(chan struct{}, 1)
+		entry.HooksDone = hooksDone
+		entry.log(DebugLevel, fmt.Sprint(args...), hooksDone)
 	}
 	return entry
 }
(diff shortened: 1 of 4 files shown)


logger.go
```diff
@@ -64,38 +64,43 @@ func (logger *Logger) WithFields(fields Fields) *Entry {
 	return NewEntry(logger).WithFields(fields)
 }
 
-func (logger *Logger) Debugf(format string, args ...interface{}) {
+func (logger *Logger) Debugf(format string, args ...interface{}) *Entry {
 	if logger.Level >= DebugLevel {
-		NewEntry(logger).Debugf(format, args...)
+		return NewEntry(logger).Debugf(format, args...)
```
aybabtme: this break the API guarantees of respecting the `StdLogger` interface

entry.go
```diff
@@ -29,6 +29,10 @@ type Entry struct {
 
 	// Message passed to Debug, Info, Warn, Error, Fatal or Panic
 	Message string
+
+	// A receive-only notification channel for knowing when all hooks have
+	// finished firing.
+	HooksDone <-chan struct{}
```
aybabtme: I think a `sync.WaitGroup` would be more appropriate here. Also can you explain the purpose of exposing this? It's not clear to me what this is supposed to be used for. Hooks don't have an ID, so you could have many hooks being fired, one of them yours, and you'd be notified that one of those many hooks finished?

entry.go
```diff
@@ -124,7 +124,9 @@ func (entry Entry) log(level Level, msg string) {
 
 func (entry *Entry) Debug(args ...interface{}) *Entry {
 	if entry.Logger.Level >= DebugLevel {
-		entry.log(DebugLevel, fmt.Sprint(args...))
+		hooksDone := make(chan struct{}, 1)
+		entry.HooksDone = hooksDone
```
aybabtme: This is racy, multiple goroutines can invoke a `Debug/Info/Warn/...` function on the `*Entry` at the same time.

aybabtme:
I like the idea, but the implementation needs some work. In general, avoid modifying the interface of exported symbols. Also, avoid adding new exported types (I think all hooks should be async, don't need a new type).

entry.go
dradtke: The main purpose was to ensure it could be tested properly, so it could easily be made hidden and used only for testing.

aybabtme: If it's only for testing, I'd suggest doing the instrumentation in tests and not adding it to the type itself.

dradtke:
Thanks for the feedback. I agree with most of it, though I think there is one big reason to keep some hooks synchronous: asynchronous hooks break the hook's ability to modify log entry data. Since `log()` returns before the asynchronous hooks run, the value is passed back to the caller before anything can be modified. It could be fixed by having it take a pointer receiver, but as the comment above the method states, doing so could introduce a race condition.

Sirupsen:
You can make your hook run async by just running it in a goroutine, so I don't quite understand the value of this, except if you have a use-case where you want to make sure all the hooks are completed at the same time (which is still possible if you own all the hooks without this PR)?

dradtke:
I suppose this functionality could be moved to the individual hooks, but when I looked through the ones listed on the README, none of them actually did that.

If nothing else, the documentation for hooks should call out the fact that logrus fires them synchronously, so anything posting to a remote service should execute that post in a new goroutine. Right now it seems like most users assume that hooks are asynchronous by default, which is not the case.

aybabtme:
I agree with @dradtke on that point, I think it'd be a saner default that every hook be fired asynchronously. It's a bit off putting when you think about it that hooks be blocking events in someone's code, especially since they're configured quite far away (in `main` usually) from where they get triggered.

Sirupsen:
The simplest Logrus core comes from core having no assumptions about when they're run. The hooks should be implemented properly as asynchronous if that is required, instead of polluting core with that logic. I'm going to keep tracking this issue within the community, but for now it won't make it upstream.

PR #247, opened 2015-09-08
Revert api change

aybabtme:
Reverts the change in API from #179, add a compile check to ensure this doesn't happen again.

@Sirupsen @yawn @gelraen


Sirupsen:
@aybabtme thank you! That was definitely a mistake.

Issue #207, opened 2015-07-02
Make logrus.Fields an implementable interface


Volox:
I think it would be better if `log.WithFields` accept as input an interface so everyone can implement such interface if needed.

For example by now if I have a simple structure and i want to use it with the `WithFields` method I must create a _wrapper_ that returns a `logrus.Fields` struct or write the attributes manually.

Sample Options struct:

``` go
type Options struct {
    Paginate bool
    Count    uint
    MaxPages uint
}

// Wrapper function can became the implementation of the interface
func (o *Options) ToFields() (fields log.Fields) {
    fields = make(log.Fields)
    fields["paginate"] = o.Paginate
    fields["count"] = o.Count
    fields["maxpages"] = o.MaxPages
    return
}
```

Given that struct we need to do either:

``` go
// With the wrapper
log.WithFields(options.ToFields()).Info( "Test" )
```

OR

``` go
// Manual entries
log.WithFields(log.Fields{
    "paginate": options.Paginate,
    "count": options.Count,
    "maxpages": options.MaxPages,
}).Info( "Test" )
```

But if we have the interface we can simply write:

``` go
log.WithFields(options).Info( "Test" )
```

Sirupsen:
This is a great idea and solves a bunch of problems people have asked for in the past at the right level of abstraction.

ryanfaerman:
@Sirupsen Do you mind if I take a shot at tackling this?

Sirupsen:
@ryanfaerman sure!

alecha:
I'm also interested in this feature, how can we do to don't break the existing function calls?

noxiouz:
@alecha  we can add implementation of Fields interface to logrus.Fields and change signature. It would not break compatibility, I think.

ryanfaerman:
That's the approach I'm taking. Creating an interface and then having Fields be a default implementation of said interface.

alecha:
I've created a simple interface in my PR, let me know if this is a good approach ;)

taybin:
Any updates on this?

dgsb:
This issue is 5 years old. No plan for logrus 2 yet. Closing.

PR #248, opened 2015-09-08
Make Fields an implementable interface with a ToFields() method, issue

alecha:
Following the discussion started for issue #207, I’ve created a simple way to have the Fields type as implementable interface with a ToFields method.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -67,12 +67,12 @@ func (entry *Entry) WithField(key string, value interface{}) *Entry {
 }
 
 // Add a map of fields to the Entry.
-func (entry *Entry) WithFields(fields Fields) *Entry {
+func (entry *Entry) WithFields(fields FieldsConverter) *Entry {
 	data := Fields{}
 	for k, v := range entry.Data {
 		data[k] = v
 	}
-	for k, v := range fields {
+	for k, v := range fields.ToFields() {
 		data[k] = v
 	}
 	return &Entry{Logger: entry.Logger, Data: data}
--- a/exported.go
+++ b/exported.go
@@ -68,7 +68,7 @@ func WithField(key string, value interface{}) *Entry {
 //
 // Note that it doesn't log until you call Debug, Print, Info, Warn, Fatal
 // or Panic on the Entry it returns.
-func WithFields(fields Fields) *Entry {
+func WithFields(fields FieldsConverter) *Entry {
 	return std.WithFields(fields)
 }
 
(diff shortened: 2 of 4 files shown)


Sirupsen:
Is `ToFields()` really the best name? This makes a ton of sense in the context of Logrus, but may not for someone reading code who're not familiar with Logrus. I don't have a great idea.

An alternative to this approach showed up her: https://github.com/Sirupsen/logrus/pull/249 What do you think about this?

Tests and documentation would be needed when this proposal would be ready to be accepted.

@aybabtme

aybabtme:
I think this would work better if we had a call that takes varargs, alike to the way `go-kit/log` does it:

```go
log.With(keyvals ...interface{})

// log.With("key","val", "other_key", "other_val").Info("hello")
```
problem with ^ is that proper usage of the logger isn't a compile-time check:
````
func (l *log) With(keyvals ...interface{}) log {
   if len(keyvals) % 2 != 0 {
      keyvals = append(keyvals, "<MISSING VALUE>")
   }
   for i := 0; i < len(keyvals); i+=2 {
      k := keyvals[i]
      v := keyvals[i+1]
      // do something with k,v
   }
}
```

The advantage of this, however, is that since you have `interface{}`, you can type-check to see if the value implements a `log.Valuer` interface:

```go
for _, kv := range keyvals {
   if valuer, ok := kv.(log.Valuer); ok {
      k := valuer.Key()
      v := valuer.Value()
      // do something with k,v
      // remove valuer from keyvals
   }
   // continue with k=kv[i], v=kv[i+1]
}
```

~but, since we don't do that, we can't really do a type-check like that. =P

* * *

So I think the idea is valuable, but I my take on this is that I wouldn't change the current API to take a different type of argument. While this would theoretically work with people using the logger as is, it wouldn't work for people who wrap logrus to make their own customized logger, which is a lot of people.

So hum, I'm not sure what the right course of action here is. Having a way to 'hint' that a type knows how to split in 'key-values' would be nice, but I don't think this is the right solution for now.

alecha:
I can understand, actually I'd rather not add a new dependency (even if Golf is nice and widely used, so if the decision is to add a new dependency this may be the better choice) and I don't like so much the varargs solution because I tend not to use interface{} in my go programs, but it's probably not always possible ;)

Regarding the ToFields() name, because of is an interface declared in logrus, I think is reasonable if the name has a meaning who is related to logrus.

noxiouz:
Will be any updates here?

alecha:
I think that we can think about a better name instead using "ToFields()",  I don't know if @Sirupsen prefers to include the other PR which adds the Golf dependency.

aybabtme:
Sorry for the long delay here. This will break existing users that may have aliased the `Fields` type to something else, without implementing the `ToFields` interface. Because this breaks backward compatibility, I'll close. 

I think it would be a good idea for a 1.0 rewrite, however, to make the Fields an interface. So I'll tag this as 1.0.

PR #257, opened 2015-09-26
Add WithError support to logstash

creack:
Before this patch, when `WithError` is used, the value is empty.

EDIT: updated to convert all errors to string. If a field has an error type, it would not be marhsalled properly.


Sirupsen:
@stevvooe

stevvooe:
It should attempt JSON serialization then fall back to string conversion on error.

PR #249, opened 2015-09-08
Golf Support

akutz:
This patch adds support for the Golf package, a library that enables
types to define the fields they want returned about themselves or by way
of reflection.

For example, take the following type:

    type Person struct {
        Name  string
        Alias string
    }

    p := &Person{"Bruce", "Batman"}

If p were given to Logrus as a field value with an associated key of
"hero" then it would be formatted as such (given a text formatter):

    hero="&{Bruce Batman}"

However, the Person type may know how it wants to be formatted when it
comes to be logged as a field value, and this interface will provide
that advantage.

    func (p *Person) PlayGolf() bool {
        return true
    }

Now when a Person instance is provided to Logrus as a field value with
an associated key of hero it would be formatted as such (given a text
formatter):

    hero.Name=Bruce hero.Alias=Batman...

For more information on how the Golf package works, please see
https://github.com/akutz/golf. Thank you!

PR #260, opened 2015-10-05
Updated README with Golf Formatters

akutz:
This patch updates the README with mention of the third-party formatters from the [golf project](https://akutz.github.io/golf) per @Sirupsen's request in PR #249.


aybabtme:
This is an interesting project, but I don't think its relevant enough to `logrus` to come in the README, which is already quite large.  For this reason, I'll close the PR.

PR #261, opened 2015-10-06
Move airbrake

gravis:
The current implementation is using an old implementation of the
Airbrake v2 api. Airbrake has release v2 since, but this package is
necessary for in-house solutions like errbit.
The number of errbit users don't justify to continue pushing this
version as the official.
That's why the official hook is now using gobrake (the official
package), and the one coming from logrus is now named "legacy".

Note that I suggest to remove completely the [examples](https://github.com/Sirupsen/logrus/blob/master/examples) folder , which is redundant with all the examples in the [README](https://github.com/Sirupsen/logrus/blob/master/README.md)


Sirupsen:
Thank you so much @gravis.

gravis:
That was quick :)
Now I expect all the logrus users using airbrake to yell at me because their programs don't compile anymore :p

gravis:
forgot to update the changelog :(

Sirupsen:
Ya, I did that in master :)

Logrus will abuse the pre 1.0 tag until we settle on an API. For 1.0 for example I plan to have hooks and formatters use the same API. This work is in preparation for 1.0 and to cut down on maintenance burden.

PR #267, opened 2015-10-06
Missed the canonical import with Airbrake

apriendeau:


apriendeau:
So this is bigger problem. I will dig into it more

PR #269, opened 2015-10-07
Move sentry hook to external repository

evalphobia:
To move out sentry hook to external repository,

- remove hooks/sentry dir
- update README to change the link.

See discussion in #192.


Sirupsen:
Amazing, thank you!

Issue #263, opened 2015-10-06
Move Bugsnag out of core


Sirupsen:
I'm working on moving all 3rd party hooks and formatters out of core to keep core as small as possible.

@ConradIrwin @burke @BlakeMesdag do one of you mind claiming this under a personal / company repository (in the case of Bugsnag who it'd make sense maintained this) and submitting a PR to logrus to remove it and reference it in the README?

csfrancis:
I'll own this.

csfrancis:
I've moved Bugsnag out: https://github.com/Shopify/logrus-bugsnag.

BlakeMesdag:
Thanks @csfrancis, I hadn't had a chance to do anything with this yet.

PR #279, opened 2015-10-22
Generate better callstacks for Bugsnag hook

csfrancis:
@Sirupsen /cc @burke 

#### Problem

Call stacks in the Bugsnag UI include Logrus frames in the stack, which is less than ideal:
image

#### Solution

Use `bugsnag-go/errors` to generate a stack trace that omits the four frames of Logrus internals before logging the error.


Sirupsen:
Can you own https://github.com/Sirupsen/logrus/issues/263 @csfrancis ? :heart_eyes:

csfrancis:
Replaced with https://github.com/Shopify/logrus-bugsnag/pull/1.

PR #285, opened 2015-10-27
Added Build Rule

GJRTimmer:
Add build rule, to only build on Linux platform, reason for adding this: This allows the usage of the logrus package within a Windows based project. Current logrus project fails to import due to the fact the syslog is missing on Windows.

Important Note:
The additional empty newline between the // +build and the package declaration is mandatory for GoLang to detect the build command.

Current change is tested on Windows, It allows an import of Logrus without errors


Sirupsen:
A+ thank you

PR #289, opened 2015-11-03
Cross compiling for Solaris-like OS

0xE0F:
As described in issue #112 not all OS could be built. I've added support for Solaris OS (illumos kernel and SmartOs as well).

Since golang doesn't define SYS_IOCTL for Solaris, I've found nice extension "golang.org/x/sys" which has built-in IoctlGetTermios function.

I've seen projects who solve issue with absence of SYS_IOCTL by adding it manually, but for me using golang.org/x/sys seem more portable way.

### DIFF
--- a/terminal_solaris.go
+++ b/terminal_solaris.go
@@ -1,0 +1,15 @@
+// +build solaris
+
+package logrus
+
+import (
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// IsTerminal returns true if the given file descriptor is a terminal.
+func IsTerminal() bool {
+	_, err := unix.IoctlGetTermios(int(os.Stdout.Fd()), unix.TCGETA)
+	return err == nil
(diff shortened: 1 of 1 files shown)


Sirupsen:
I'm not hugely in favour of adding a dependency to Logrus, being as it's a library included in a large amount of contexts. Will it really require vendoring very much code to get just this piece of functionality for Solaris derived OSes?

0xE0F:
Since an internal syscall package was frozen and there is no SYS_IOCTL for Solaris, the package "gaoling.org/x/sys" seems to me is only way to get access to low-level function.

0xE0F:
Besides, extra dependency is actual only for Solaris platform and I suggest this pull-request because I need gorethink driver for RethinkDB working under Solaris.

Sirupsen:
Thanks for your patch :)

PR #291, opened 2015-11-09
Added ProvidesLogMessageAndData interface

akutz:
This patch introduces the `ProvidesLogMessageAndData` interface. This interfaces enables types to directly influence the log message and data provided to a log entry when a user simply invokes `log.Info(obj)`. This simplified syntax will make things like error handling and other log-aware types so much easier on the end-developer and result in cleaner code. 

@Sirupsen, I am specifically needing this functionality for the error handling package, [Goof](https://github.com/akutz/goof), I created. If you take a look at that package I think you'll be convinced of the value of this patch. It's more than I can do via a TextFormatter.

Please see the file `provides_test.go` for examples.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -116,7 +116,9 @@ func (entry Entry) log(level Level, msg string) {
 
 func (entry *Entry) Debug(args ...interface{}) {
 	if entry.Logger.Level >= DebugLevel {
-		entry.log(DebugLevel, fmt.Sprint(args...))
+		if !entry.logMsgAndData(DebugLevel, args...) {
+			entry.log(DebugLevel, fmt.Sprint(args...))
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


Sirupsen:
Does https://github.com/Sirupsen/logrus/pull/248 resolve this for you?

akutz:
Hi @Sirupsen,

No, it does not, because the idea is that supplying an `error` object  by itself, or any other object that provides message and data, to any of the functions `Debug`, `Info`, etc. should be formatted accordingly. 

This is to greatly simplify how logging occurs and to minimize code requirements to do so. Why should one always have to do `WithError` or `WithFields` when the log functions can detect if a type supplies a message and/or data?

Besides, I specifically want to replace the log message with the error's message, and since custom formatters do away with the logged argument by the time an entry is received, all I have is the formatted argument along with any fields that may have been passed.

Thus the argument is checked to see if it supplies a message and data, and if so, it negates the need for additional, chained calls.

akutz:
Hi @Sirupsen,

Also, be sure to check out the example from the Goof project (https://github.com/akutz/goof).

Sirupsen:
This is going to be considered for the design of 1.0 which will be much more modular, which I anticipate having a prototype for by the end of this year.

aybabtme:
I'll close this PR for now, since we won't merge it in the current state of the logger, but we'll keep track of the idea (it's a good idea) for @Sirupsen's 1.0 rewrite.

PR #294, opened 2015-11-17
Entry times as UTC with new TimestampInUTC formatting option

markcampanelli-wf:
To ensure consistency across systems, use UTC for all Entry times.

Add a new formatting option, the boolean TimestampInUTC, in both the TextFormatter and the JSONFormatter. This boolean defaults to false, giving the existing behavior using local times.

One JSONFormatter unit test had to be changed slightly by setting TimestampInUTC: true.

It would be good to add unit testing of the TimestampInUTC setting, but I'm not sure the best way to do this.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -21,7 +21,7 @@ type Entry struct {
 	// Contains all the fields set by the user.
 	Data Fields
 
-	// Time at which the log entry was created
+	// Time, in UTC, at which the log entry was created
 	Time time.Time
 
 	// Level the log entry was logged at: Debug, Info, Warn, Error, Fatal or Panic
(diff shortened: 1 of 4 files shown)


Sirupsen:
This breaks backwards compatibility a little too hard :/ Why are your system clocks not UTC?

markcampanelli-wf:
That is a good question, and it appears that the clocks where we deploy to are in UTC. However, most (all?) of our localhost development uses local time, which makes it a bit more difficult to integrate local instances with deployed microservices. TimestampInUTC would conveniently make this more consistent.

aybabtme:
This can be done by a custom formatter, which like - you proposed here - would print times in UTC. Otherwise, it's a change that will likely confuse unexpected users as their local logs now print in another timezone. 

I'll be closing this PR, since there's no loss of capabilities here (you can get the behavior you want with a custom formatter) and it breaks backward compatibility.

PR #298, opened 2015-11-21
formatter.go: simplify prefixFieldClashes(Fields)

dolmen:
Minor refactoring: simplify the implementation of `prefixFieldClashes(Fields)` in formatter.go

### DIFF
--- a/formatter.go
+++ b/formatter.go
@@ -31,18 +31,15 @@ type Formatter interface {
 // It's not exported because it's still using Data in an opinionated way. It's to
 // avoid code duplication between the two default formatters.
 func prefixFieldClashes(data Fields) {
-	_, ok := data["time"]
-	if ok {
-		data["fields.time"] = data["time"]
+	if t, ok := data["time"]; ok {
+		data["fields.time"] = t
 	}
 
-	_, ok = data["msg"]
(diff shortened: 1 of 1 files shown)


noxiouz:
It disallows optimization for ```_, ok```. So if the fields does not collade very often, the code becomes slower.

dolmen:
Interesting. Do you have some links about those optimisations?

noxiouz:
Look at this bench:
http://play.golang.org/p/O5lKruoQmY

Results on my laptop:
```
PASS
BenchmarkKandOK-4         	100000000	        11.2 ns/op
BenchmarkOkOnly-4         	300000000	         4.97 ns/op
BenchmarkKandOKNotExists-4	100000000	        16.4 ns/op
BenchmarkOkOnlyNotExists-4	100000000	        11.7 ns/op
ok  	github.com/mapbench	6.012s
```
You can look at asm code, which is generated in both cases with:
```
 go tool compile -S m_test.go
```

aybabtme:
I think this reads better, and the performance difference due to compiler optimization doesn't bother me a lot. Especially, compiler optimizations like this are very hit-and-miss and I expect the SSA backend to improve things a lot.

PR #303, opened 2015-12-03
Detect TTY based on stderr, not stdout

illicitonion:
We actually write to stderr by default, so:

bin >/dev/null

currently weirdly prints non-colorized output, whereas:

bin 2>log

weirdly prints colorized output to a file.


Sirupsen:
Nice catch, thanks.

PR #307, opened 2015-12-16
implemented IsTerminal() on Solaris

fazalmajid:
This is a Solaris implementation of IsTerminal(), based on https://github.com/golang/go/issues/13085

### DIFF
--- a/terminal_solaris.go
+++ b/terminal_solaris.go
@@ -1,15 +1,22 @@
+// Based on ssh/terminal/util_solaris.go:
+// Copyright 2015 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 // +build solaris
 
 package logrus
 
 import (
-	"os"
-
 	"golang.org/x/sys/unix"
+	"syscall"
 )
 
 // IsTerminal returns true if the given file descriptor is a terminal.
+// see: http://src.illumos.org/source/xref/illumos-gate/usr/src/lib/libbc/libc/gen/common/isatty.c
 func IsTerminal() bool {
-	_, err := unix.IoctlGetTermios(int(os.Stdout.Fd()), unix.TCGETA)
+	fd := syscall.Stderr
+	var termio unix.Termio
+	err := unix.IoctlSetTermio(fd, unix.TCGETA, &termio)
 	return err == nil
 }


terminal_solaris.go
```diff
@@ -1,15 +1,22 @@
...
+	"syscall"
 )
 
 // IsTerminal returns true if the given file descriptor is a terminal.
+// see: http://src.illumos.org/source/xref/illumos-gate/usr/src/lib/libbc/libc/gen/common/isatty.c
 func IsTerminal() bool {
-	_, err := unix.IoctlGetTermios(int(os.Stdout.Fd()), unix.TCGETA)
+	fd := syscall.Stderr
+	var termio unix.Termio
+	err := unix.IoctlSetTermio(fd, unix.TCGETA, &termio)
```
aybabtme: won't this mutate the state of the file descriptor? seems like undesirable behavior from a method named `IsTerminal`

binarycrusader: I'm confused as to why this is calling IoctlSetTermio instead of IoctlGetTermio, and I'm also confused as to why this is assuming stderr specifically instead of stdout.

Can you explain that a bit more?

fazalmajid: @aybabtme & @binarycrusader: Good question. The code in question isn't present in the original git://golang.org/x/crypto/ssh/termina/util_solaris.go where I first implemented it.

stderr vs. stdout because I used terminal_notwindows.go as a starting point. Since `IsTerminal()` does not have an argument to specify which fd to test for terminal-ness, The rationale for @illicitonion's choice of stderr is here:
https://github.com/Sirupsen/logrus/commit/05f9567ba3659e78f02d071dd0c23600e0602c1e

I am no longer sure why I submitted this pull request. I suspect what happened is I implemented terminal_solaris.go to get InfluxDB or Prometheus to build on Solaris, and only later merged master back and pushed it, not realizing someone else had already implemented this independently:
https://github.com/Sirupsen/logrus/commit/8cdd4b39f502d10358d92586e020d3535f257ec9

I verified the code in the master branch works on Solaris (well, apart from the stdout vs. stderr issue). Feel free to disregard this pull request.

fazalmajid:
closed by 8cdd4b3

Sirupsen:
Releasing master which will work with Solaris. Thanks @binarycrusader and @fazalmajid

PR #311, opened 2016-01-16
Add file and line number

omidnikta:
This is the way I propose to add File-Line number capability. The function
ShowFileLine(bool) will set the appearance of the file and line number of the source file in logs. If you set 
ShowFileLine(false)
it only contains a function call and a boolean comparison, so performance is not an issue in it.


omidnikta:
New working pull request will be provided.

xild:
is someone working on this? 
I can work in a new PR.

PR #312, opened 2016-01-17
Add caller capability

omidnikta:
This pull request adds the capability of adding file and line number to the logger. Use
ShowCaller(true/false) to enable/disable it.


entry_test.go
```diff
@@ -22,7 +22,7 @@ func TestEntryWithError(t *testing.T) {
 
 	logger := New()
 	logger.Out = &bytes.Buffer{}
-	entry := NewEntry(logger)
+	entry := NewEntry(logger, 1)
```
aybabtme: the test should assert that the right caller is printed

entry.go
```diff
@@ -29,16 +31,23 @@ type Entry struct {
...
 		Logger: logger,
 		// Default is three fields, give a little extra room
-		Data: make(Fields, 5),
+		Data:  make(Fields, 5),
+		Depth: d,
 	}
 }
 
+func (entry *Entry) deepen() {
+	entry.Depth = entry.Depth + 1
```
aybabtme: this mutates the entry, which means multiple calls will result in bogus callers

omidnikta: As far as I know, when you call to log, a new entry is created. So no data race is possible.

entry_test.go
omidnikta: There is no NewEntry(logger) anymore. You need to feed a number to it like NewEntry(logger, 1). When you set SetCaller(false), the caller will have no effect.

entry.go
aybabtme: nha, `Entry` are reusable

entry_test.go
aybabtme: what I'm saying is that the test assertions need to check that the right `callrer: src.go:42` is printed

entry.go
omidnikta: Latest commit fixes the reusing entry bug, if an entry will not be used concurrently.

entry_test.go
omidnikta: This problem is now solved.

entry.go
omidnikta: This problem is solved now.

aybabtme:
closing due to inactivity and failed CI.

apriendeau:
@aybabtme this failed due to travis. WE should consider reopening:

```
Cloning into '/home/travis/gopath/src/github.com/logrus/logrus'...
remote: Invalid username or password.
fatal: Authentication failed for 'https://github.com/logrus/logrus/'
package github.com/logrus/logrus: exit status 128
```

aybabtme:
@apriendeau travis failed because the code is incorrect.

apriendeau:
Well the logs are very vague and not useful. @aybabtme are we okay with the implementation because I will fix the import paths if we are.

aybabtme:
@apriendeau the proposed change in this PR breaks existing users and is larger than strictly necessary. I'm not personally in favor of merging it.

Issue #313, opened 2016-01-18
Event modified by lagstash formatter


goodfoo:
Our app shares a global Event instance pre-populated with fields which are used in logstash parsing (program, environment, host, sha).

When using the logstash formatter the fields.message is populated with the previous message.

This line tries to preserve a message or level set in the fields object
https://github.com/Sirupsen/logrus/blob/master/formatters/logstash/logstash.go#L33

But it actually stores the current message to the fields object, which then gets picked on the next log event.

It would be safer if the logstash formatter did not mutate the Event.

goodfoo:
this PR attempts to avoid the issue by creating a new logrus.Fields and copying values.

https://github.com/Sirupsen/logrus/pull/314

PR #314, opened 2016-01-18
avoid modifying the entry by copying all fields and values

goodfoo:
this PR avoids mutating the passed Event object at the cost of creating a new logrus.Fields and copying values.

see https://github.com/Sirupsen/logrus/issues/313


aybabtme:
lgtm, thanks for the fix!

PR #316, opened 2016-01-26
Add the Logstash hook

ripcurld00d:
Since, `logrus` has a Logstash formatter, why not having a Logstash hook too.


Sirupsen:
We no longer accept hooks in core. Please move this under your own maintained repository, and submit another PR with a link. Thanks!

mpmlj:
+1

PR #320, opened 2016-02-01
Add LogrusLogger interface for Entry and Logger

### LINKED ISSUE #308: Logger vs. Entry: provide interface
Hi

I wanted to have a logger as an argument to a function, but I don't necessarily care whether it is a `Logger` or an `Entry`, as both provide the same functions (`Info`, `Debug`, `WithFields`, ...), and I want the caller to decide the context of the logging.

It is a bit weird to take an `Entry`, and have the caller pass `logrus.NewEntry(logger)` instead of just `logger`.

I expected there to be an interface that can be used in place.

little-arhat:
This make it possible for client code to accept either Logger or Entry.
For example, utility function may accept logger object to inform fatal
errors and it is job of the calling code to provide either generic
top-level logger, or request-bound Entry created using .WithFields.

There are some tests for compilation stage check.

(fixes #308)

### DIFF
--- a/logrus.go
+++ b/logrus.go
@@ -96,3 +96,39 @@ type StdLogger interface {
 	Panicf(string, ...interface{})
 	Panicln(...interface{})
 }
+
+// Logrus logger interface generalizes Entry and Logger types, so you can take any of these
+type LogrusLogger interface {
+	// we can return LogrusLogger here, but this will require many changes and will
+	// possible break backward compatiblity
+	WithField(key string, value interface{}) *Entry
+	WithFields(fields Fields) *Entry
+	WithError(err error) *Entry
+
+	Debugf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Printf(format string, args ...interface{})
+	Warnf(format string, args ...interface{})
+	Warningf(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
+	Fatalf(format string, args ...interface{})
+	Panicf(format string, args ...interface{})
+
+	Debug(args ...interface{})
+	Info(args ...interface{})
+	Print(args ...interface{})
+	Warn(args ...interface{})
+	Warning(args ...interface{})
+	Error(args ...interface{})
+	Fatal(args ...interface{})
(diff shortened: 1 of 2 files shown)


chancez:
This would be super great. I tend to basically pass tons of `*logrus.Entry` around and it can be a bit confusing for those unfamiliar.

little-arhat:
@Sirupsen what do you think?

rschmukler:
I would also love to see this!

ear:
If you pass a `logger.WithField("key", val)` (or `WithFields(…)`) off to a consumer does the code in this PR let the receiver add its own extra field(s) before calling a logging method?

little-arhat:
@ear, sure! This is one of the purpose of the interface from this PR — to hide away difference between `Entry` and `Logger` and to be able to work with just one type (interface) in functions/methods where you want to do the logging.

As you may see in tests, https://github.com/Sirupsen/logrus/pull/320/files#diff-9469334cadef7f7ea2f6a4db193646b7R321 `fn` accepts interface `LogrusLogger` and adds some new fields inside, before actual logging. Caller, in turn, passes `logger.WithField` to `fn`.

ear:
Awesome!

I started using this package very recently (days) and realized only earlier after posting here that I could do this using the `Entry` type.

little-arhat:
@ear sure, you can use `Entry` type, but then you will not be able to pass just `Logger` to such function. Purpose of this PR is to abstract over these two instances, as they should be interchangeable in most cases.

Sirupsen:
I'm renaming it to `FieldLogger` in master, as this shouldn't be unique to Logrus. Thanks for this!

chancez:
YES im so happy about this.

Sirupsen:
Included in 0.10.0

jakesyl:
Is usage documented anywhere?

freeformz:
@jakesyl If [this](https://godoc.org/github.com/sirupsen/logrus#FieldLogger) isn't sufficient (and it *may* not be) please open an issue detailing what you need more info on. Thx!

PR #324, opened 2016-02-14
add keep args support

xujinzheng:
sometimes we need use the orignial params of `args` which passed from `Debug`, `Info`, `Warn`, `Error`, `Fatal` or `Panic`... in hooks or formatters.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -123,7 +123,8 @@ func (entry *Entry) Debug(args ...interface{}) {
 }
 
 func (entry *Entry) Print(args ...interface{}) {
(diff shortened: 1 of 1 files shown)


aybabtme:
I'm not sure why this is needed, I'm not a fan of the idea. Especially, `Printf/Print/Println` style logging with args should be discouraged, in favor of purely structured logging. Exposing this in the Entry encourages an undesired practice.

aybabtme:
I'll close for now, feel free to reopen if you feel strongly about it!

PR #331, opened 2016-03-01
Add WriterLevel() function to the logger

dradtke:
This commit adds a variant of the logger's Writer() function that
accepts a log level. When the variant is used, any messages written to
the returned pipe will be written with the provided level. The original
Writer() function uses the logger's Print() method as it always has.

This makes the logger pluggable into e.g. Gorilla's [recovery handler](http://www.gorillatoolkit.org/pkg/handlers#RecoveryHandler) without adding confusion by reporting panic messages using the default log level used by `Print()`:

```go
r := mux.NewRouter()
r.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
    panic("Unexpected error!")
})

http.ListenAndServe(":1123", handlers.RecoveryHandler(
    // Specify a custom output for the recovery handler.
    // In this case, we want it to utilize Logrus, and all messages should be printed as errors.
    handlers.RecoveryLogger(log.New(logrus.StandardLogger().WriterLevel(logrus.ErrorLevel), "", 0)),
)(r))
```

This change should be 100% backwards-compatible with existing code.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -7,18 +7,40 @@ import (
 )
 
 func (logger *Logger) Writer() *io.PipeWriter {
+	return logger.WriterLevel(255)
+}
+
+func (logger *Logger) WriterLevel(level Level) *io.PipeWriter {
 	reader, writer := io.Pipe()
 
-	go logger.writerScanner(reader)
+	var printFunc func(args ...interface{})
+	switch level {
+	case DebugLevel:
+		printFunc = logger.Debug
+	case InfoLevel:
+		printFunc = logger.Info
+	case WarnLevel:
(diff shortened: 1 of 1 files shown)


writer.go
```diff
@@ -7,18 +7,40 @@ import (
 )
 
 func (logger *Logger) Writer() *io.PipeWriter {
+	return logger.WriterLevel(255)
```
aybabtme: can you change this to be `Info` level?

aybabtme:
@dradtke thanks for the PR, I applied my comment in #347

PR #332, opened 2016-03-10
vendorize dependencies using Godep

lobeck:
tracking dependencies is hard, that's why there is a tool called godep (https://github.com/tools/godep) since go1.5 which takes care of it.

this fork contains the required dependencies tracked in the Godeps.json and additionally everything in vendor/ as suggested by the godep documentation.

- add Godeps file to track dependencies
- commit dependencies in vendor/ folder


mynameismevin:
I love it. +1 for vendoring.

methane:
I don't like it.
go get installs dependencies.  It's enough for libraries.
I use vendoring only for building executable, not libraries.

lobeck:
i'm also still not sure about this, as i had installation issues when i installed it first, but it worked after removing everything and `go get` it again.

Also the point regarding vendoring of executables vs. libraries sounds valid.

MarkSonghurst:
Presumably this would require pushing the use of godep on users of logrus?

I use Glide, an alternative to godep.
I'd sooner not have to keep 2 different vendoring tools up to date.

I agree with the stance on only vendoring executables.
Perhaps if dependency versions are critical for logrus releases then they could somehow be pinned down in a Makefile?

Sirupsen:
10,000s of clones are done of this repo every day. Each clone would now include all these test-related dependencies and deploy them to their production environments. None of these are run-time requirements.

Thanks for the suggestion though.

PR #333, opened 2016-03-11
Add Typetalk hook

dragon3:
I wrote a Typetalk hook for logrus.
Typetalk is a chat app like Slack and Hipchat ( https://www.typetalk.in/ )

I'd just like to add the link in the description of hooks.


dragon3:
Thanks!

Issue #310, opened 2016-01-14
AppEngine integration


9uuso:
Hi,

I tried to use a database driver with my application today, which relies on this logger. I wanted to have my app run on Google App Engine, but I noticed that the only thing which is standing on my way is the use of `syscall` package. Looking at the code, I noticed that the original `terminal_linux.go` file is based on Go's official `ssh/terminal` repo. However, now that I took a look at `ssh/terminal`, they have updated it to use [hard-coded constants for linux](https://github.com/golang/crypto/blob/c8b9e6388ef638d5a8a9d865c634befdc46a6784/ssh/terminal/util_linux.go). In the end, I was able to get logrus build on appengine with [some modifications](https://github.com/9uuso/logrus/commit/28716abd1d1bcca1ce279e04ef63aede463352c8).

Declaration of `IsTerminal()` in `terminal_appengine.go` is clearly a hack and I don't have any idea whether I may app may have side-effects from it.

Since the tests pass and the program compiles on Darwin and AppEngine, I was wondering if you could give me an opinion whether there could be chance for some hack like this to get into the main project? I'd imagine that at least some people on upstream projects could find the hack welcome.

If you are fine with the idea, I can make a PR of it.

Sirupsen:
Fine with me!

sirupsen:
https://github.com/sirupsen/logrus/issues/343

PR #343, opened 2016-04-02
terminal: Include appengine tag to compile for GAE

kaneshin:
Hi!

I tried to deploy an application to PaaS, which is Google App Engine.
The application includes `github.com/Sirupsen/logrus` for logging, but I noticed logrus containes `syscall` package which is prohibited using on GAE. So I hacked `terminal_[arch].go` referencing 
https://github.com/Sirupsen/logrus/issues/310 issue.

If you are fine with the PR, please merge it!

Thank you.

### DIFF
--- a/terminal_bsd.go
+++ b/terminal_bsd.go
@@ -1,4 +1,5 @@
 // +build darwin freebsd openbsd netbsd dragonfly
+// +build !appengine
 
 package logrus
 
--- a/terminal_appengine.go
+++ b/terminal_appengine.go
@@ -1,0 +1,8 @@
+// +build appengine
+
+package logrus
+
+// IsTerminal returns true if stderr's file descriptor is a terminal.
+func IsTerminal() bool {
+	return true
+}
--- a/terminal_linux.go
+++ b/terminal_linux.go
@@ -3,6 +3,8 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// +build !appengine
+
 package logrus
 
 import "syscall"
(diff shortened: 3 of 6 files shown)


terminal_bsd.go
```diff
@@ -1,4 +1,5 @@
 // +build darwin freebsd openbsd netbsd dragonfly
+// +build !appengine
```
kaneshin: It means, `(darwin OR freebsd OR openbsd OR netbsd OR dragonfly) AND !appengine`.

aybabtme:
im ok with this, @Sirupsen any objection?

kaneshin:
👀

kaneshin:
👀

Sirupsen:
Should we always emit TTY colours on GCE?

kkirsche:
@Sirupsen I haven't been able to find anything about GCE using TTY colors in their documentation. I haven't done testing myself though

aarongreenlee:
@kkirsche My hunch is that we should not emit colors for GCE.

kaneshin:
Why did you talk about GCE?
It makes no colors output if you build or run application by google app engine binary.
Not only GCE. It should be happened on OS X as well if the PR isn't applied.

abelaska:
@Sirupsen please merge. This is really important for us. It blocks our affords to deploy to Google App Engine since this package is also used by other important packages like https://github.com/dancannon/gorethink

Thank you!

andrewhare:
👍

andrewhare:
@Sirupsen Thanks :)

PR #346, opened 2016-04-15
race in logstashformatter.go

e-max:
I've got warning about date race logstashformat and fixed it. 

But there are another places which looks dangerous. For example  Entry.Data is public so there is a possibility that some code outside  will write in this map in the same moment as you read from it in Format method.


aybabtme:
thanks!

PR #356, opened 2016-05-06
Fix comment

tevino:


aarongreenlee:
@tevino Thanks for the typo correction.

PR #358, opened 2016-05-17
Add new hook for Logmatic.io

gpolaert:
Logmatic.io is Saas-based log management solution.
We develop a simple hook in order to send your Logrus logs straight to Logmatic.io


aarongreenlee:
Thank you for this contribution, @gpolaert.

PR #359, opened 2016-05-18
Terminal detection removed, use single flag for set colored output

chapsuk:
This #289  include bigger dependency(~4Mb) and make `logrus` more fat.
Terminal detection used only for make colored output, what about use single field `logrus.TextFormatter.UseColors` for enable colored output?


Sirupsen:
@chapsuk the dependency for https://github.com/Sirupsen/logrus/pull/289 is pulled in on all platforms? Is your problem the binary size?

Terminal detection is a core feature of Logrus. We want some magic here to make development as nice as possible.

chapsuk:
Problem is `vendor/` size, `+6Mb` for `2Mb` project is too big. Terminal detection is nice feature, but used  only for colored output. Set flag more simple way for make this.
Thanks!

Sirupsen:
@chapsuk maybe it's possible to create a compile flag that removes these dependencies instead? I'm not on top of the new flag changes in newer versions of Go.

PR #361, opened 2016-05-24
Added func to log with a level as imput

Defaultuser0001:


Sirupsen:
This is not something we're currently interested in supporting. You can likely implement something with a custom hook and a `WithFields` or a formatter.

Issue #151, opened 2015-03-13
Extending JSONFormatter data JSON serialization (Patch advice)


cep21:
I want to write a patch to solve the below problem: Would this be accepted?

I have a structure with a function pointer field.  Whenever I run

    logrus.WithField("item", item)....

It fails using JSONFormatter with

    Failed to obtain reader, Failed to marshal fields to JSON, json: unsupported type: datum.LifetimeFunc

This is because it cannot be serialized into JSON.  This is a common object that I would like to log in multiple places.  It would be great if JSONFormatter supported something like the following

type LogFormattable interface {
  LogFormat() interface{}
}

Then, inside Format() for JSONFormatter, it could check if the object implements LogFormattable and if so, use LogFormat() rather than the object itself in the JSON log.

Does this sound reasonable?  If so, I could get a pull request up shortly.

cep21:
As an extra issue, it would be great if I could trigger some kind of callback on log formatting issues inside Entry.log() rather than them going to stderr.  This has caused issues in production code where untracked error logs were getting dropped.

Sirupsen:
What about if you pass a function as a value to a field, it gets executed? That makes sense in general, because expensive values (e.g. for debugging or lower debugging levels) would then only be computed when needed.

cep21:
I really like that.  It requires reflection packages though, right?  Would reflection on every field slow down logrus too much, or would that be acceptable?

Sirupsen:
We're already checking types, which in my experience is fairly cheap. I've wanted this in Logrus for some time, because it seems expected and would solve a number of people's performance problems (and being able to do crazy, fancy things). It shouldn't be hard to do, but it would have to come with a benchmark to proof it doesn't slow down Logrus extensively. I don't think it would, because it wouldn't do extra allocations. Let me know if you want to work on this.

cep21:
I'll work on that patch.  It requires changing both the json and stdout formatter.  You feel trying to cast into a ```func() string``` is better than implementing a type?

cep21:
Hi,

Before I work on this I wanted to clarify the use of func() rather than interface?  One option is something like this:


```go
type Loggable interface {
    Logged() interface {}
}

type F func() interface{}

func (f F) Logged() interface {} {
  return f()
}

...
```

Then code can look like this:
```go
log.WithField("complex", complexItem).Info("Hi")
```

And I don't need to remember to cast out complex item.  People can add inline functions with:

```go
[...]

PR #365, opened 2016-06-02
field values can be the result of a function call made at time of logging

thwarted:
Looking for feedback on this, I'm not married to this implementation, but I've gotten use out it for the use case described below.

This change lets you specify a function to be called _at the time the log entry is generated_ to set the value of a field.  This way you can set up a `logrus.Entry` with a bunch of context that may not be known yet, but pass that `logrus.Entry` around to other routines that can use it directly without having to create a new one with a call to `.WithFields`.

The function must take no arguments and return a string, and must be cast to `logrus.Fn`.  I didn't want to use an interface for this because that would require creating a new type in the simple case just to define a function on it.  As it stands now, you can use a inline function definition/closure, a pointer-receiver method reference, or a function name reference, so it's pretty flexible.

My exact use case is that I create an object and set up a logrus instance on it, but it needs to finish building itself out at some point in the future, or I want to log the current state of the object whenever it might emit a log line.  

A simplified, contrived version of how I'm using this:

``` go
type obj struct {
   log *logrus.Entry
   id  string
   amt int
}

func (o *obj) logIdent() string { return o.id }
func (o *obj) logAmt() string { return strconv.Itoa(o.amt) }

o := &obj{}
o.log = logrus.WithFields(logrus.Fields{
   "id":  logrus.Fn(o.logIdent),
   "amt": logrus.Fn(o.logAmt),
})
… // do something with o that modifies id and amt and does logging
o.log.Info("event occurred") // log with amt=0
o.amt -= 100
if o.amt < 0 {
  o.log.Error("amt below threshold") // log with amt=-100
}
```

I couldn't find a way to do this with the current code, other than to create new `logrus.Entry` every time the value could change: I had originally achieved this functionality by assigning a new instance of `logrus.Entry` (by calling `logrus.WithFields`) to `o.log` whenever I might change the value of `id` or `amt`, but this was error prone and didn't feel clean because, in my case, I'm populating the instance from various sources (unmarshalling from database or network, or manually building, etc) or various business rules change `amt`, and it was easy to miss having to touch logging on every one.

Suggestions for tests that would be appropriate for this change are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,26 @@ In general, with Logrus using any of the `printf`-family functions should be
 seen as a hint you should add a field, however, you can still use the
 `printf`-family functions with Logrus.
 
+Usually, you'll use simple types, like string, as the values of the the fields.
+If you need to calculate the value of the field at run time, but want to
+centralize the definition of the logger with a set of fields, a function can
+be specified as the value of the field:
+
+```go
+customLogger = log.WithFields(log.Fields{
+  "static":  100,
+  "dynamic": log.Fn(myFunc),
+})
+...
+customLogger.Fatal("failed to send event")
+```
+
+This will generate a field named `dynamic` with the string value that `myFunc`
+returns *at the time the log entry is generated*.  The function must take no
+arguments and return a string, and must be cast to `logrus.Fn`.  It can be
+defined inline (`logus.Fn(func() string { ... })`), be an interface method,
+or a plain function referenced by name.
+
 #### Hooks
 
 You can add hooks for logging levels. For example to send errors to an exception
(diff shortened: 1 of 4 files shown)


thwarted:
Ah, I see this was already suggested in #151 (to be included with #207 ?).  No activity in a while though.

dgsb:
Up

dgsb:
@thwarted I like the simplicity of using a closure/function pointer to compute a dynamic value to add the trace at the time the trace is issued.
Regarding the interface, I'd rather add a new method to the Logger and Entry, something like WithFunc.
wdyt ?
I know this PR has been opened quite a long time, if you're not interested anymore I will complete the work to make it mergeable into master

thwarted:
A `WithFunc` method might be a little cleaner, it might make it easier to define the function inline without a cast.

I'm no longer using logrus, so I don't have a strong opinion.  Do as you will!

freeformz:
Thank you, but in order to clean up older PRs and issues (this PR hasn't had any activity in the last ~ year), I'm going to close this PR. Feel free to re-open with an updated branch if this is still of interest. PS: I'm sorry this sat for so long.

PR #369, opened 2016-06-18
Optimise buffer allocation

frostyplanet:
Previously bytes.Buffer allocate twice in Entry.Reader() and Formater.Format() and it's not reused, we use sync.Pool to reduce the allocation.

Removed entry.Reader() because it seams unnecessary.

Added a benchmark case for logger.

Before:
```

$ go test -run=xxx  -bench NULLLogger -cpu=1 .
PASS
BenchmarkNULLLogger	  300000	      4344 ns/op
ok  	_/home/plan/projects/qingcloud/logrus	1.354s

plan@plan-tp:~/projects/qingcloud/logrus$ go test -run=xxx  -bench NULLLogger -cpu=8 .
PASS
BenchmarkNULLLogger-8	  500000	      2828 ns/op
ok  	_/home/plan/projects/qingcloud/logrus	1.449s
```
After: 
```
$ go test -run=xxx  -bench NULLLogger -cpu=1 .
PASS
BenchmarkNULLLogger	  500000	      3581 ns/op
ok  	_/home/plan/projects/qingcloud/logrus	1.833s

$ go test -run=xxx  -bench NULLLogger -cpu=8 .
PASS
BenchmarkNULLLogger-8	 1000000	      2195 ns/op
ok  	_/home/plan/projects/qingcloud/logrus	2.224s
```

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -3,10 +3,20 @@ package logrus
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"os"
+	"sync"
 	"time"
 )
+
+var bufferPool *sync.Pool
+
+func init() {
+	bufferPool = &sync.Pool{
+		New: func() interface{} {
+			return new(bytes.Buffer)
+		},
+	}
+}
 
 // Defines the key when adding errors using WithError.
 var ErrorKey = "error"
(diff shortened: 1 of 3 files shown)


frostyplanet:
re-open with  different branch

PR #370, opened 2016-06-18
Optimise speed

### LINKED ISSUE #376: High level of contention in Logger mutexes
We use logrus in a handful of high traffic services and find ourselves having to be extremely careful about  where we make logger calls otherwise the services can end up blocking inside [Entry#log](https://github.com/Sirupsen/logrus/blob/4b6ea7319e214d98c938f12692336f7ca9348d6b/entry.go#L101)and causing some grief.  Given we're just emitting them to the docker host over UDP, this contention often ends up needlessly blocking as 

I figure there are multiple ways of fixing this, so let's ask before sending patches.  

Ideally we could have our own implementation of `Logger` which maintained its own internal mutexes, or used buffered channels, ring buffers or ... something similar.  Under critical I'd rather drop messages than drop requests, though that's probably not true of all services.

However if there was another approach you'd prefer we take I'm happy to take a look.

frostyplanet:
- fixes https://github.com/sirupsen/logrus/issues/376

Previously bytes.Buffer allocate twice in Entry.Reader() and Formater.Format() and it's not reused, we use sync.Pool to reduce the allocation.

Removed entry.Reader() because it seams unnecessary.

*Entry can be reuse when calling function directly from Logger.

Added a benchmark case for logger.

Before:

```

$ go test -run=xxx  -bench DummyLogger -cpu=1 .
PASS
BenchmarkDummyLogger      300000          4344 ns/op
ok      _/home/plan/projects/qingcloud/logrus   1.354s

$ go test -run=xxx  -bench DummyLogger -cpu=8 .
PASS
BenchmarkDummyLogger-8    500000          2828 ns/op
ok      _/home/plan/projects/qingcloud/logrus   1.449s
```

After: 

```
$ go test -run=xxx  -bench DummyLogger -cpu=1 .
PASS
BenchmarkDummyLogger      500000          3581 ns/op
ok      _/home/plan/projects/qingcloud/logrus   1.833s

$ go test -run=xxx  -bench DummyLogger -cpu=8 .
PASS
BenchmarkDummyLogger-8   1000000          2195 ns/op
ok      _/home/plan/projects/qingcloud/logrus   2.224s
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,3 +400,20 @@ handler := func() {
 logrus.RegisterExitHandler(handler)
 ...
 ```
+
+#### Thread safty
+
+By default Logger is protected by mutex for concurrent writes, this mutex is invoked when calling hooks and writing logs.
+If you are sure such locking is not needed, you can call logger.SetNoLock() to disable the locking.
+
+Situation when locking is not needed includes:
+
+* You have no hooks registered, or hooks calling is already thread-safe.
+
+* Writing to logger.Out is already thread-safe, for example:
+
+  1) logger.Out is protected by locks.
+
+  2) logger.Out is a os.File handler opened with `O_APPEND` flag, and every write is smaller than 4k. (This allow multi-thread/multi-process writing)
+
+     (Refer to http://www.notthewizard.com/2014/06/17/are-files-appends-really-atomic/)
--- a/entry.go
+++ b/entry.go
@@ -3,10 +3,20 @@ package logrus
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"os"
+	"sync"
 	"time"
 )
+
+var bufferPool *sync.Pool
+
+func init() {
+	bufferPool = &sync.Pool{
+		New: func() interface{} {
+			return new(bytes.Buffer)
+		},
+	}
+}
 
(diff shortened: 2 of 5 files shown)


frostyplanet:
I have also the method  Logger.SetNoLock(), in case the user wants to disable mutex locking for faster performance. When log file is opened with append mode, it's safe to write concurrently to a file without locking.

```
$ go test -run=XXX  -bench Dummy -v
PASS
BenchmarkDummyLogger-4      	  500000	      2649 ns/op
BenchmarkDummyLoggerNoLock-4	 1000000	      1876 ns/op
ok  	_/home/plan/projects/qingcloud/logrus	3.258s
```

aarongreenlee:
@frostyplanet Thanks for contributing. Could you please also contribute to the documentation with this pull request to explain the use case and provide an example for disabling locking?

frostyplanet:
@aarongreenlee  Did you mean adding description in README.md ?

frostyplanet:
Document is added. And this branch is rebased to master

aarongreenlee:
Thank you for this contribution, @frostyplanet.

frostyplanet:
Thanks for accepting

PR #371, opened 2016-06-20
Introduce a LevelVar for the flag package

zimbatm:
This makes it easy to switch the log level using the flag package.

### DIFF
--- a/level_var.go
+++ b/level_var.go
@@ -1,0 +1,32 @@
+package logrus
+
+// Can be used with the `flag' package to parse Logrus level from
+// command-line args.
+//
+// Example:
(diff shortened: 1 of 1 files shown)


dmathieu:
Hey,

I am going to close this as it is something which can already be implemented in your own code with `ParseLevel`.

```go
var lvl string
flag.Var(&lvl, "log-level", "Sets the log level")
parsed, _ := logrus.ParseLevel(lvl)
logrus.SetLevel(parsed)
```

Adding this method will therefore not provide anything new.

zimbatm:
Ok no worries. The main advantage is that it handles the error case like any other flag parsing error.

Issue #295, opened 2015-11-18
Data Race in entry.go:128


cosminilie:
There seems to be a data race between github.com/Sirupsen/logrus.(*Entry).Info() at line 128  and the github.com/Sirupsen/logrus.SetLevel() method. I guess this applies to all methods on the Entry (Info, Warn, Error, Debug) at the if statement "if entry.Logger.Level >= InfoLevel {....."

This happens to me, because i want to update the loglevel externally (http endpoint).

Ex:

``` golang
go func() {
        for {
            level := <-logLevel
            log.SetLevel(level)
        }
}()

//Business domain
    go func() {
        log.WithFields(log.Fields{
            "File":   "main.go",
            "Action": "BussinessLogicInit",
        }).Info("Started Main Bussiness Unit")
   ......
}()
```

Sirupsen:
Patch very welcome :+1:

cosminilie:
Will try to send a patch. Thank you

ash2k:
Same thing about all the other fields - `Formatter`, `Hooks`, `Out`.
e.g. `log.SetFormatter(...)` accesses field `Formatter` of the global instance while holding the mutex, but code in `entry.go` is reading from the field not holding the mutex.

rand99:
yep, server down

```
panic: sync: WaitGroup is reused before previous Wait has returned

goroutine 97 [running]:
panic(0x739d80, 0xc848524040)
	/usr/local/go/src/runtime/panic.go:481 +0x3e6
sync.(*WaitGroup).Wait(0xc8202ee968)
	/usr/local/go/src/sync/waitgroup.go:129 +0x114
vendor/github.com/gemnasium/logrus-graylog-hook.(*GraylogHook).Flush(0xc8202ee940)
	vendor/github.com/gemnasium/logrus-graylog-hook/graylog_hook.go:91 +0x2d
vendor/github.com/gemnasium/logrus-graylog-hook.(*GraylogHook).Fire(0xc8202ee940, 0xc8470a2f00, 0x0, 0x0)
	vendor/github.com/gemnasium/logrus-graylog-hook/graylog_hook.go:84 +0xf2
vendor/github.com/Sirupsen/logrus.LevelHooks.Fire(0xc8202f4420, 0xc816a31a02, 0xc8470a2f00, 0x0, 0x0)
	/vendor/github.com/Sirupsen/logrus/hooks.go:28 +0x106
[...]

PR #374, opened 2016-06-23
fix data race on set level and formatter

klizhentas:
fixes Sirupsen/logrus#295

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -27,7 +27,7 @@ type Logger struct {
 	// logged. `logrus.Debug` is useful in
 	Level Level
 	// Used to sync writing to the log.
-	mu sync.Mutex
+	mu sync.RWMutex
 }
 
 // Creates a new logger. Configuration should be set by changing `Formatter`,
@@ -71,13 +71,13 @@ func (logger *Logger) WithError(err error) *Entry {
 }
 
 func (logger *Logger) Debugf(format string, args ...interface{}) {
-	if logger.Level >= DebugLevel {
+	if logger.level() >= DebugLevel {
 		NewEntry(logger).Debugf(format, args...)
 	}
 }
 
 func (logger *Logger) Infof(format string, args ...interface{}) {
-	if logger.Level >= InfoLevel {
+	if logger.level() >= InfoLevel {
 		NewEntry(logger).Infof(format, args...)
 	}
 }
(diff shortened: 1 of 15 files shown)


logger.go
```diff
@@ -181,32 +181,44 @@ func (logger *Logger) Println(args ...interface{}) {
...
+func (logger *Logger) level() Level {
+	logger.mu.RLock()
+	defer logger.mu.RUnlock()
+	return logger.Level
+}
+
+func (logger *Logger) formatter() Formatter {
+	logger.mu.RLock()
+	defer logger.mu.RUnlock()
+	return logger.Formatter
```
ash2k: You may avoid some overhead by not using `defer` as it is not strictly necessary here - reading a field can not panic and `logger` is not `nil` because it was accessed one line above. Same for the `level()` method.

aarongreenlee: If you could make the small adjustment to remove the defer and ping me, I'd be happy to merge this pull request.

Thanks for contributing this improvement!

klizhentas: @aarongreenlee I've addressed the code review comment

klizhentas:
@aarongreenlee ping

aarongreenlee:
Thanks, @klizhentas. This all looks good to me. Can you please resolve these minor conflicts? I'll merge shortly thereafter. 

Thanks for this contribution.

johanbrandhorst:
@klizhentas pinging to have merge conflicts resolved. Good work!

klizhentas:
I don't think it's actually possible to have my changeset to coexist with the master mostly because of this chunk of code that turns off locking:

https://github.com/Sirupsen/logrus/blob/master/logger.go#L53

https://github.com/Sirupsen/logrus/blob/master/logger.go#L306

Moreover I think the new `MutexWrap` introduces one more race condition on it's own - the `disabled` variable read/writes are not atomic and should be actually `atomic` operations. Moreover there's theoretically  could be a condition when you:
```go
Lock()
SetNoLock()
Unlock()
```

getting a deadlock

DmitriyMV:
I think it's partially fixed by #512, but `Formatter` problem remains. The mutex will prevent that but it will also introduce performance penalities in situations where the same log is used by several goroutines in parallel. Additional info here: https://github.com/golang/go/issues/17973 TLDR - it doesn't scale well. 

However, I'm still unconvinced that we really should bother about that in the first place, given the fact the most of the loggers and hooks are IO bound. Any thoughts on the matter? Synthetic benchmarks do not really show the realistic scenarios.

PR #375, opened 2016-06-24
Added Support to Call Handlers on Fatal

aarongreenlee:
This pull request enhances the exported API with two new functions which allow for Exit Handlers to be invoked by any Fatal call.

**New Functions**
```
AddExitHandler( func() { ... } )
Exit(1)
```

While GO offers the ability to recover from panic there is no way to intercept an os.Exit event. To allow graceful shutdown and clean-up or programs which use Logrus to exit using a Fatal call I've borrowed the idea of the `atexit` package and enhanced Logrus to use it's own Exit() function which invokes handlers before quitting.

**Usage**
* When setting up a logger one may optionally call `RegisterExitHandler( func() {...} )` to add a function which will be invoked for any `Fatal` call to the logger before ultimately calling `os.Exit(1)`.

### DIFF
--- a/alt_exit.go
+++ b/alt_exit.go
@@ -1,0 +1,60 @@
+package logrus
+
+// The following code was sourced and modified from the
+// https://bitbucket.org/tebeka/atexit package governed by the following license:
+//
+// Copyright (c) 2012 Miki Tebeka <miki.tebeka@gmail.com>.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to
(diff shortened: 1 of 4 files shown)


aarongreenlee:
I'm not sure how to test this properly without some means of sharing state between the test application and the logrus software under test.

**Question:**
Would you be ok with adding a build tag to the test file I introduced so that the test only runs when disk is available?

aarongreenlee:
Need to update documentation and add examples before merging.

stevvooe:
@aarongreenlee I had some concerns about adding this in https://github.com/sirupsen/logrus/pull/385#discussion_r95447494.

stevvooe:
Specifically, how is support for this orthogonal to hooks?

PR #381, opened 2016-07-06
Allow replacing Exit function

ericklaus-wf:
Adds the ability to replace the exit function called by logrus's Fatal() family. Tests can use this to replace the function with a stub that panics or otherwise stores the exit call without killing the test run.

This does mean that libraries receiving a Logger will be able to circumvent the exit functionality of Fatal calls, but those libraries could already arbitrarily terminate the program, change the logger's output destination, and add hooks, so I don't think it's really a concern.

See Sirupsen#375 is another PR modifying exit behavior. However, that PR still ultimately results in the program terminating.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -150,7 +150,7 @@ func (entry *Entry) Fatal(args ...interface{}) {
 	if entry.Logger.Level >= FatalLevel {
 		entry.log(FatalLevel, fmt.Sprint(args...))
 	}
-	os.Exit(1)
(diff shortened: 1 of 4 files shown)


aarongreenlee:
@ericklaus-wf, can you please restore the build to a passing state and ping me? I'll check back when the tests are passing for code review. 

Thanks for contributing.

ericklaus-wf:
I'll recreate this with the test bug fixed and conflicts resolved.

PR #378, opened 2016-06-30
Check for error types in logstash formatter

jonoradich:
Check for `error` type to correctly serialise errors.

See https://github.com/Sirupsen/logrus/issues/377.

PR #382, opened 2016-07-12
Removing logstash formatter

lpetre:
Moved this formatter and the associated commit history to the https://github.com/bshuster-repo/logrus-logstash-hook repository. As per conversation here: https://github.com/Sirupsen/logrus/pull/378


Sirupsen:
Thanks a lot @lpetre

PR #383, opened 2016-07-18
customize message field with json formatter

### LINKED ISSUE #367: JSONFormatter should be able to customize `msg` field name
Our log parsing infrastructure currently assumes that full-text message have `message` name ( similar to logstash json format )

```
{
  "message"    => "hello world",
  "@version"   => "1",
  "@timestamp" => "2014-04-22T23:03:14.111Z",
  "type"       => "stdin",
  "host"       => "hello.local"
}
```

However the current json formatter assumes that the field is called `msg` without any option to modified ( https://github.com/Sirupsen/logrus/blob/master/json_formatter.go#L33). It would be nice to be able to rename `msg` to `message`.

dqminh:
This add `MessageField` option to JSONFormatter to customize the message field
name instead of the default `msg`.

For example, to change the message field to `message`

```
log.SetFormatter(&logrus.JSONFormatter{MessageField: "message"})
```

Fix #367

### DIFF
--- a/formatter.go
+++ b/formatter.go
@@ -2,7 +2,10 @@ package logrus
 
 import "time"
 
-const DefaultTimestampFormat = time.RFC3339
+const (
+	DefaultTimestampFormat = time.RFC3339
+	DefaultMessageField    = "msg"
+)
 
 // The Formatter interface is used to implement a custom Formatter. It takes an
 // `Entry`. It exposes all the fields, including the default ones:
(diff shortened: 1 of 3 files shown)


aarongreenlee:
Thanks, @dqminh. I'll take a look at this pull request sometime this week.

dqminh:
@aarongreenlee have you had the chance to look at this ?

stevvooe:
Why not allow the parameterization of any of the fields?

stevvooe:
Looks like #396 addresses this, as well.

dqminh:
@stevvooe yep, #396 also does this. I will just close this one since i think #396 is reasonable.

PR #384, opened 2016-07-18
Renaming upper-case 'Sirupsen' to 'sirupsen'

mnzt:
The reason for this is that most other packages use lower-case naming conventions.

The vendoring tool I use (GB) is case sensitive, and can become a massive pain in the :peach: when Logrus is the only package with an upper-case name (as most other packages are in lower case, following https://blog.golang.org/package-names) - this often leads to build failures.

> Why don't you just vendor it in lower case?

Because Logrus has itself as an internal dependency, it vendors two folders, `Sirupsen/logrus` and `sirupsen/logrus`.

### DIFF
--- a/examples/hook/hook.go
+++ b/examples/hook/hook.go
@@ -1,7 +1,7 @@
 // +build ignore
-// Do NOT include the above line in your code. This is a build contrainst used
+// Do NOT include the above line in your code. This is a build constraint used
 // to prevent import loops in the code whilst go get'ting it.
-// Read more about build contrains in golang here:
+// Read more about build constraints in golang here:
 // https://golang.org/pkg/go/build/#hdr-Build_Constraints
 
 package main
--- a/examples/hook/hook.go
+++ b/examples/hook/hook.go
@@ -1,9 +1,6 @@
 package main
 
-import (
-	"github.com/Sirupsen/logrus"
-	"gopkg.in/gemnasium/logrus-airbrake-hook.v2"
-)
+import "github.com/sirupsen/logrus"
 
 var log = logrus.New()
 
--- a/examples/hook/hook.go
+++ b/examples/hook/hook.go
@@ -1,6 +1,9 @@
 package main
 
-import "github.com/sirupsen/logrus"
+import (
+	"github.com/sirupsen/logrus"
+	airbrake "gopkg.in/gemnasium/logrus-airbrake-hook.v2"
+)
 
 var log = logrus.New()
 
--- a/examples/hook/hook.go
+++ b/examples/hook/hook.go
@@ -1,3 +1,9 @@
+// +build ignore
+// Do NOT include the above line in your code. This is a build contrainst used
+// to prevent import loops in the code whilst go get'ting it.
+// Read more about build contrains in golang here:
+// https://golang.org/pkg/go/build/#hdr-Build_Constraints
+
 package main
 
 import (
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Logrus <img src="http://i.imgur.com/hTeVwmJ.png" width="40" height="40" alt=":walrus:" class="emoji" title=":walrus:"/>&nbsp;[![Build Status](https://travis-ci.org/Sirupsen/logrus.svg?branch=master)](https://travis-ci.org/Sirupsen/logrus)&nbsp;[![GoDoc](https://godoc.org/github.com/Sirupsen/logrus?status.svg)](https://godoc.org/github.com/Sirupsen/logrus)
+# Logrus <img src="http://i.imgur.com/hTeVwmJ.png" width="40" height="40" alt=":walrus:" class="emoji" title=":walrus:"/>&nbsp;[![Build Status](https://travis-ci.org/sirupsen/logrus.svg?branch=master)](https://travis-ci.org/sirupsen/logrus)&nbsp;[![GoDoc](https://godoc.org/github.com/sirupsen/logrus?status.svg)](https://godoc.org/github.com/sirupsen/logrus)
 
 Logrus is a structured logger for Go (golang), completely API compatible with
 the standard library logger. [Godoc][godoc]. **Please note the Logrus API is not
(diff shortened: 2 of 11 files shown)


aarongreenlee:
Thank you for your contribution, @mnzt.

Tools like `go get` seem to be case insensitive. I wonder if GB would also benefit by becoming case insensitive.

I don't see any negative side effects to this pull request and will accept it now.

Thanks again.

mnzt:
@aarongreenlee thank you for taking time to review the PR :smile:.

It looks like a linter tool automatically removed the airbrake import, whoops!

I've noticed that logrus examples imports airbreak, which in turn imports logrus - to circumvent this issue I have added a build constraint flag. Hopefully all goes well on the following CI check :smile_cat:

examples/hook/hook.go
```diff
@@ -1,8 +1,14 @@
+// +build ignore
+// Do NOT include the above line in your code. This is a build contrainst used
```
glasser: "constraint"

examples/hook/hook.go
```diff
@@ -1,8 +1,14 @@
...
+// to prevent import loops in the code whilst go get'ting it.
+// Read more about build contrains in golang here:
```
glasser: "constraints"

glasser:
Does it make sense to use Go 1.4 custom import path checking for this? https://docs.google.com/document/d/1jVFkZTcYbNLaTxXD9OcGfn7vYv5hWtPx9--lTx1gPMs/edit

mnzt:
@glasser I think the main issue here was that an import cycle was created in `examples/hook`.

mnzt:
Any progress on this @aarongreenlee?

stevvooe:
@Sirupsen This looks like a real issue but it would be a possibly breaking change.

Any opinion here?

Sirupsen:
How do you think this could break @stevvooe?

stevvooe:
@Sirupsen When this package changes to `github.com/sirupsen/logrus`, won't the builds fail if the only thing found in GOPATH is `github.com/Sirupsen/logrus` on a case sensitive filesystem?

For the most part, I am not against this change. If we are willing to risk possible breakage, it is for the better.

mnzt:
People can still vendor an upper case S in sirupsen, however it's common
practice to keep packages as lower case.

mnzt:
The real issue here is other dependencies vendoring with an upper case S.
Often at work I end up with two vendored lots of logrus - as half use the
common practice, and the over half use what's the in the documentation.

stevvooe:
LGTM

@mnzt Have you seen any breakage?

@Sirupsen If you put your LGTM, I think we should go through with this. We can send out some tweets to let people know.

Sirupsen:
Let's go ahead with this.

Will it work with this if I rename myself to @sirupsen instead of @Sirupsen on Github? It's silly that I have upper-case.

stevvooe:
@Sirupsen Please, do the honors. I gave up my capital S long ago.

sirupsen:
Just renamed myself 🙈

stevvooe:
💥

ahmetalpbalkan:
no more the firstmost dependency on Godeps.json 💥

hilljgo:
Note: this broke our glide dependencies with the following error:
```
[WARN]	Error updating github.com/sirupsen/logrus: The Remote does not match the VCS endpoint
```

changing our import from `github.com/Sirupsen/logrus` to `github.com/sirupsen/logrus` fixed it

stevvooe:
@hilljgo Thanks for the report! Seems like a low impact fix.

hryx:
I observed the same Glide issue as @hilljgo, but the problem ended up being exponentially more painful when vendoring several projects which also use Logrus.

If using the old name, Glide fails due to a mismatched remote URL. If using the new name, Go fails due to a `case-insensitive import collision`.

So the only option I've found, until those vendored projects all start using the lowercase `s`, is to lock to an older version of Logrus and tell Glide the remote URL explicitly:

```yaml
- package: github.com/Sirupsen/logrus
  version: a437dfd2463eaedbec3dfe443e477d3b0a810b3f
  repo:    https://github.com/sirupsen/logrus
```

bbrks:
Just leaving a comment here for anybody Googling errors.

```
case-insensitive import collision: "github.com/Sirupsen/logrus" and "github.com/sirupsen/logrus"
```

The fix is to lowercase your import statement for logrus.

mnzt:
Note on the above @bbrks' comment, for https://getgb.io you'll need to unvendor (`gb vendor delete github.com/Sirupsen/logrus`) the uppercase named dependency and then vendor the lowercase named dependency (`gb vendor fetch github.com/sirupsen/logrus`).

There may be a small teething period of getting everything over to the correct naming - but it'll be worth it in the end

cstockton:
The import path is valid per the [specification](https://golang.org/ref/spec#Import_declarations), if the tool doesn't work then they should fix it, if your builds don't work, you should fix them. Breaking every users build with the assumption they could simply vendor is .. ugh, very narrow sighted. Many people import libraries which use logrus, and don't control the import declarations of those libraries and are unable to update them. Meaning they are broke until the package maintainers react.

sirupsen:
I've needed to revert this. This breaks way too many builds.

https://github.com/sirupsen/logrus/issues/451

sirupsen:
If there remains interest in this, we should do this under `logrus/logrus`.

mnzt:
I believe the underlying issue here is where internal code 'externally' refers to `{S,s}irupsen/logrus`, rather than a local import, ie: `../../something`. This causes issues in the examples folder however, meaning the examples become uncopyable.

A potential solution is to move the examples out of this repo into their own repo, meaning `logrus` as a package can consist exclusively of local imports - meaning users can vendor/go get via any case they desire.

/cc @sirupsen @stevvooe

maxmeyer:
@sirupsen Any near plans about moving the repo?

nazieb:
@sirupsen Is the plan moving the repo still on? If that's so, considering github.com/logrus is already taken, I suggest using github.com/go-logrus so we can get benefit of using something like gopkg.in

I've registered the org by the way, and will happily transfer it to you if you want.

PR #385, opened 2016-07-18
Allow replacing Exit function

ericklaus-wf:
Adds the ability to replace the exit function called by logrus's Fatal() family. Tests can use this to replace the function with a stub that panics or otherwise stores the exit call without killing the test run.

This does mean that libraries receiving a Logger will be able to circumvent the exit functionality of Fatal calls, but those libraries could already arbitrarily terminate the program, change the logger's output destination, and add hooks, so I don't think it's really a concern.

### DIFF
--- a/alt_exit.go
+++ b/alt_exit.go
@@ -51,6 +51,16 @@ func Exit(code int) {
 	os.Exit(code)
 }
 
+// handleExit runs all the Logrus atexit handlers and then terminates the program using the logger's Exit function (or os.Exit() if no exit function is set).
+func (logger *Logger) handleExit(code int) {
+	runHandlers()
+	if logger.Exit != nil {
+		logger.Exit(code)
+	} else {
+		os.Exit(code)
+	}
+}
+
 // RegisterExitHandler adds a Logrus Exit handler, call logrus.Exit to invoke
 // all handlers. The handlers will also be invoked when any Fatal log entry is
 // made.
--- a/entry.go
+++ b/entry.go
@@ -161,7 +161,7 @@ func (entry *Entry) Fatal(args ...interface{}) {
 	if entry.Logger.Level >= FatalLevel {
 		entry.log(FatalLevel, fmt.Sprint(args...))
 	}
-	Exit(1)
+	entry.Logger.handleExit(1)
 }
 
 func (entry *Entry) Panic(args ...interface{}) {
@@ -209,7 +209,7 @@ func (entry *Entry) Fatalf(format string, args ...interface{}) {
 	if entry.Logger.Level >= FatalLevel {
 		entry.Fatal(fmt.Sprintf(format, args...))
 	}
-	Exit(1)
+	entry.Logger.handleExit(1)
 }
 
 func (entry *Entry) Panicf(format string, args ...interface{}) {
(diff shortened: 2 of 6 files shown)


ericklaus-wf:
@aarongreenlee I remade this PR after fixing merge conflicts from the `atexit` PR and resolving the failing tests CI was complaining about.

Sorry about those failing tests!

aarongreenlee:
Thanks, @ericklaus-wf. I'll take a look at this contribution this week.

aarongreenlee:
@ericklaus-wf Some merge conflicts have popped up. Can you please resolve them? I anticipate a quick merge shortly thereafter. Thank you!

ericklaus-wf:
@aarongreenlee I rebased with master and resolved conflicts.

ericklaus-wf:
@aarongreenlee How do you feel about getting this merged in?

alt_exit.go
```diff
@@ -51,6 +51,16 @@ func Exit(code int) {
 	os.Exit(code)
 }
 
+// handleExit runs all the Logrus atexit handlers and then terminates the program using the logger's Exit function (or os.Exit() if no exit function is set).
+func (logger *Logger) handleExit(code int) {
+	runHandlers()
+	if logger.Exit != nil {
+		logger.Exit(code)
```
stevvooe: What if the function doesn't actually exit? Should this panic in that case?

ericklaus-wf: I don't think so. If someone replaces the exit functionality, they're taking matters into their own hands and we should honor their replacement.

stevvooe: That breaks the guarantee that `Fatal` exits.

Why would test which logs are called?

ericklaus-wf: Consider wanting to test a main method that terminates when certain bad input is provided.

```
func main() {
    args, err := parseArgs()
    if args.a < 0 {
        log.Fatal("a was less than zero!")
    }
    executeAwesome(args.a)
}
```

If a dev really wanted to test the `Fatal` branch, a dev would need to create a drop-in replacement for Logger and inject that replacement at unit test time. With this change, a dev can replace the `Exit` method with one that (for example) panics and recover from the panic as part of the unit test.

ericklaus-wf: I agree that this allows a library user to change `Fatal`'s behavior to instead not exit. I don't see that as a bad thing. But if the consensus is that it *is* a bad thing, we should scrap this PR and instead write one to replace a `Fatal`'s exit with a `panic` (probably via a flag).

stevvooe: @ericklaus-wf This kinds of speaks to the odd cases where `log.Fatal` or `log.Panic` come up. They are very awkward to test. I don't really want to get into a debate about testing methodology, but it seems like this would be better covered by integration testing that would ensure that both `log.Fatal` was called and the processed exited accordingly.

Are there any unit tests in the stdlib that test `log.Fatal`? Maybe we can find inspiration here.

My main concern is that something that is meant to be a testing tool will quickly become a way to do an end around on the hooks subsystem.

sirupsen:
Looks like this was done in https://github.com/sirupsen/logrus/pull/375

PR #349, opened 2016-04-18
Remote caller

apriendeau:
This fixes the problems in #312 to address #63.

Issue #63, opened 2014-09-22
Log filename and line number


gust1n:
Both stdlib package log and many other logging frameworks are supporting this (e.g. https://github.com/inconshreveable/log15). Am I missing how to enable it or is it not supported?

cep21:
I solve this in a performant manner by doing the following:

```go
var log *logrus.Logger

func init() {
	log = mylog.NewLogger()
}


func NewLogger() *logrus.Logger {
	_, file, _, ok := runtime.Caller(1)
	if !ok {
		file = "unknown"
	}
	inst := &logInstance{
		innerLogger: ret,
		file:        filepath.Base(file),
	}
	ret.Hooks.Add(inst)
    return ret
}
```

logInstance is a hook that adds "file" to the log entry.  This saves me calling runtime.Caller each time I log a message while still having code that does

```go
  log.WithField()....Info()
```

chreble:
+1 it's a MUST-HAVE, and saying "completely API compatible with the standard library logger" in the README is wrong as we have that feature in the standard Go logger (SetFlags) http://golang.org/pkg/log/#SetFlags

gonzaloserrano:
+1

Sirupsen:
Perhaps someone take a closer look at `SetFlags` implementation and port it to `logrus`? Currently all the implementations I've seen rely on hardcoding number of stack frames which is not very portable and doesn't work in some cases.

chreble:
@Sirupsen,  I can definitely have a look :)

Sirupsen:
@chreble awesome, let me know what you find!

dmlyons:
@Sirupsen unfortunately, the `SetFlags` implementation in the standard library uses a (sort of) hardcoded number of stack frames: http://golang.org/src/log/log.go#L130.  Each exported output function, such as Println, passes `callDepth` like this: http://golang.org/src/log/log.go#L169.

Is this such a big deal to implement in logrus?

Sirupsen:
@dmlyons no it's not, I've just not needed this myself and haven't seen an implementation that hasn't failed on different platforms.

davisford:
Has anyone resolved this in a reasonable way?  I just want the filename and line number.  I'd also prefer to be able to configure it to only print on specific levels (e.g. err, panic, warn)

gonzaloserrano:
[...]

PR #388, opened 2016-07-24
remote caller (another take)

victorhaggqvist:
So, since #349 have not been merged yet, I thought I make another go at implementing remote caller support, addressing #63  (This patch is not based on any prior work)

There are a few differences in this patch compared to #349.
Mainly that here the actual remote caller look up is done in the "core" of logrus if you will, and then stored in the `Entry` as `RemoteCaller`. `RemoteCaller` will then contain some thing like `/full/path/to/file.go:42` after a successful lookup.

It is then up to the formatter if it likes to make use of the info. I have added remote caller as `caller` in both the text and json formatter.

I believe this is closer to how it should be done if logrus should be able to claim that it has support for remote caller. Since this actually add info about the caller, not just a the option that the user would like to show caller which is how it is done in #349 then basically leaving it up to the formatter to actually find the caller.

About compatibility, `ShowCaller` is on by default. Which means that any one that uses logrus out of the box with no additional config will have caller now appear with out any config change. To disable this one can set `logger.ShowCaller = false` on the logger. Any custom formatters output won't be affected, since they now should make use of `entry.RemoteCaller` if they want to show the caller. (Thought if one use a custom formatter not making use of `RemoteCaller` one should probably set `ShowCaller = false` to avoid doing the lookup).

### DIFF
--- a/formatter.go
+++ b/formatter.go
@@ -1,8 +1,16 @@
 package logrus
 
-import "time"
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
 
 const DefaultTimestampFormat = time.RFC3339
+
+// To be used for triming base path of remote caller
+var gosrcpath = filepath.Join(os.Getenv("GOPATH"), "src")
 
 // The Formatter interface is used to implement a custom Formatter. It takes an
 // `Entry`. It exposes all the fields, including the default ones:
@@ -30,7 +38,7 @@ type Formatter interface {
 //
 // It's not exported because it's still using Data in an opinionated way. It's to
 // avoid code duplication between the two default formatters.
-func prefixFieldClashes(data Fields) {
+func prefixFieldClashes(data Fields, hasRemoteCaller bool) {
 	if t, ok := data["time"]; ok {
 		data["fields.time"] = t
 	}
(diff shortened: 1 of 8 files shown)


aarongreenlee:
Thanks for this contribution @victorhaggqvist. I'll take a look at this later in the week.

emilevauge:
Any news on this ?

formatter.go
```diff
@@ -30,7 +38,7 @@ type Formatter interface {
 //
 // It's not exported because it's still using Data in an opinionated way. It's to
 // avoid code duplication between the two default formatters.
-func prefixFieldClashes(data Fields) {
+func prefixFieldClashes(data Fields, hasRemoteCaller bool) {
```
edrex: It seems safe enough to always prefix the caller key, which would avoid the need to change the method signature and the calls in the formatters.

victorhaggqvist:
Just a update from me.

I have done some further work/experimentation that progress on top of this branch at https://github.com/victorhaggqvist/logrus/tree/feat/remote-caller-use. Mostly trying to come up with a reliable way to find the correct depth in the stack, still not quite sure if the way it is implemented there is optimal either, but I have found it more reliable at least.

The other change on that branch is a slightly more opinionated formatter.

If maintainer like, I can push that work to this branch as well.

For other waiting on this feature, feel free to take a test drive of the implementation in https://github.com/victorhaggqvist/logrus/tree/feat/remote-caller-use. Maybe try it out in your toy-projects.

formatter.go
victorhaggqvist: It probably is. Though, In the progression of the branch ([here](https://github.com/victorhaggqvist/logrus/tree/feat/remote-caller-use)), I have actually removed the prefixing entirely, in favour of a more specific name ie. `remote_caller`.

Issue #118, opened 2015-01-29
Provide Output method to Entry


nvcnvn:
Standar log pakcage have the _Output_ method for Logger http://golang.org/pkg/log/#Logger.Output
Maybe Entry should have that method so it interface exactly with std Logger.

maxpersson:
+1. As an example, not having this makes SetLogger() in go-nsq hard to use.

sundarv85:
+1. Same as @maxpersson has said. Want to use logrus with go-nsq and Output method is mandatory for it to be used with [go-nsq's SetLogger](https://github.com/bitly/go-nsq/blob/master/consumer.go#L209).

Sirupsen:
@sundarv85 @maxpersson why not just use `WithFields` in `go-nsq`?  Set your logger where you need those fields on every statement somewhere with:

```go
logger := logrus.WithFields(logrus.Fields{
  "id": r.id,
  "topic": r.topic,
  "channel": r.channel
})
```

And then use it:

```go
logger.Info("OMG messages")
logger.Warn("OOMG too many messages!")
```

That'll log with all those fields. That's a much cleaner pattern in my opinion.

Although `logrus` should implement `Output()` for full backwards compatibility with the `stdlogger`, but I wouldn't encourage its behaviour (unless I'm missing a case).

nvcnvn:
The issue is some libs (in this case NSQ Go client) require the specific
"Output" method for their logger interface.
Example: https://godoc.org/github.com/bitly/go-nsq#Conn.SetLogger
Logrus should implement Output method so it can be replacment for std log
package anyway!

Sirupsen:
@nvcnvn ya, I agree it should be in, I'm just saying I don't see it as a Logrus best practise.

nvcnvn:
Hi Simon,
May be I'm not understand your previewous comment correctly.
The reason why we can't use logrus.WithFields is we're not the author of
go-nsq so it will take time to fork then read all go-nsq code and finally
replace all the log lines with logrus.
That will easier if Logrus have Output method.

Sirupsen:
oooh, that use case makes a lot more sense. Gotcha! :sparkles:

nvcnvn:
Does any one know what is expected behavior of "Output" method Calldepth arg?
Reading out std log package http://golang.org/pkg/log/#Logger.Output say that:
"Calldepth is used to recover the PC and is provided for generality, although at the moment on all pre-defined paths it will be 2."
[...]

PR #390, opened 2016-07-30
Add log.Logger's Output method to Logger&Entry

droyo:
The standard library's *log.Logger type has the method

```
Output(calldepth int, s string) error
```

Because it is so minimal, many libraries chose to target
it as an interface, which users could implement to receive
additional log messages:

```
type Logger {
    Output(int, string) error
}
```

While it is not overwhelmingly used, it is out there. Implementing
this interface allows us to use logrus.Logger and logrus.Entry
types to receive log messages from such libraries. I chose to log
such messages at the "Info" level, copying the Print\* methods. The
"calldepth" argument is ignored.

Addresses issue #118

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -254,6 +254,11 @@ func (entry *Entry) Panicln(args ...interface{}) {
 	}
 }
 
+func (entry *Entry) Output(calldepth int, s string) error {
+	entry.Infoln(s)
+	return nil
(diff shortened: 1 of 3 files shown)


gravis:
Why was this issue closed?
It makes perfect sense to me

Thanks

PR #396, opened 2016-08-15
Added customizable keys to JSON formatter

onetwopunch:
I have found a need for this and I'm sure it would be useful to others as well. If a user needs to customize the keys of the JSON message to fit a particular parsing scheme, they are currently disallowed from changing the key `msg` to `message` for example or `level`, or `time` as they are all hard-coded. This fix allows simple customizability of the keys in the JSON formatter.

### DIFF
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 )
+
 type fieldKey string
 type FieldMap map[fieldKey]string
 
@@ -24,7 +25,17 @@ func (f FieldMap) resolve(key fieldKey) string {
 type JSONFormatter struct {
 	// TimestampFormat sets the format used for marshaling timestamps.
 	TimestampFormat string
-	FieldMap        FieldMap
+
+	// FieldMap allows users to customize the names of keys for various fields.
+	// As an example:
+	// formatter := &JSONFormatter{
+	//   	FieldMap: FieldMap{
+	// 		 FieldKeyTime: "@timestamp",
+	// 		 FieldKeyLevel: "@level",
+	// 		 FieldKeyLevel: "@message",
+	//    },
+	// }
+	FieldMap FieldMap
 }
 
 func (f *JSONFormatter) Format(entry *Entry) ([]byte, error) {
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -8,6 +8,9 @@ import (
 type JSONFormatter struct {
 	// TimestampFormat sets the format used for marshaling timestamps.
 	TimestampFormat string
+	MessageKey      string
+	LevelKey        string
+	TimeKey         string
 }
 
 func (f *JSONFormatter) Format(entry *Entry) ([]byte, error) {
@@ -29,9 +32,24 @@ func (f *JSONFormatter) Format(entry *Entry) ([]byte, error) {
 		timestampFormat = DefaultTimestampFormat
 	}
 
-	data["time"] = entry.Time.Format(timestampFormat)
-	data["msg"] = entry.Message
-	data["level"] = entry.Level.String()
+	timeKey := f.TimeKey
+	if timeKey == "" {
+		timeKey = "time"
+	}
+
+	messageKey := f.MessageKey
+	if messageKey == "" {
+		messageKey = "msg"
+	}
+
+	levelKey := f.LevelKey
+	if levelKey == "" {
+		levelKey = "level"
+	}
+
+	data[timeKey] = entry.Time.Format(timestampFormat)
+	data[messageKey] = entry.Message
+	data[levelKey] = entry.Level.String()
 
 	serialized, err := json.Marshal(data)
 	if err != nil {
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -4,21 +4,27 @@ import (
 	"encoding/json"
 	"fmt"
 )
-
 type fieldKey string
+type FieldMap map[fieldKey]string
 
 const (
-	DefaultKeyMsg   = "msg"
-	DefaultKeyLevel = "level"
-	DefaultKeyTime  = "time"
+	FieldKeyMsg   = "msg"
+	FieldKeyLevel = "level"
+	FieldKeyTime  = "time"
 )
+
+func (f FieldMap) resolve(key fieldKey) string {
+	if k, ok := f[key]; ok {
+		return k
+	}
+
+	return string(key)
+}
 
 type JSONFormatter struct {
 	// TimestampFormat sets the format used for marshaling timestamps.
 	TimestampFormat string
-	MessageKey      string
-	LevelKey        string
-	TimeKey         string
+	FieldMap        FieldMap
 }
 
 func (f *JSONFormatter) Format(entry *Entry) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


onetwopunch:
@Sirupsen it's been a few days since I added this PR, would you or another admin please take a look? I think it could be valuable.

onetwopunch:
@aarongreenlee Would you mind having a look at this? Doesn't seem like anyone else here is actively reviewing PRs.

stevvooe:
Does this address #367?

onetwopunch:
@stevvooe Yes it does, but allows customizable `level` and `time` keys as well.

onetwopunch:
@stevvooe It seems from your comments here and on the related PR that you approve of the changes. Are you planning to merge this any time soon?

stevvooe:
@onetwopunch I am mostly getting my bearings.

To tell you the truth, it might be better to define a key type and let the user provide a map. It would look something like this:

```go
type fieldKey string

const (
  FieldKeyMsg fieldKey = "msg"
  FieldKeyLevel fieldKey = "level"
  FieldKeyTime fieldKey = "time"
)

type FieldMap map[fieldKey]string
```

Then, it can be configured as follows:

```
JSONFormatter{
  FieldMap: FieldMap{
    FieldKeyLevel: "@level",
    FieldKeyTime: "@timestamp",
  }
}
```

This avoids struct bloat as new fields are added and keeps key translation consistent.

onetwopunch:
@stevvooe I appreciate the feedback but I'd like to respectfully disagree that my method causes any more struct bloat than yours. I might agree if we were planning on adding many more keys, but since I've covered all the key types the logger will likely use, I don't see this to be a problem. 

As a nit, memory wise, your method is actually taking up slightly more space since you allocate space on the heap for the constants (even if the user does not want to configure the key) as well as the computational overhead (however small) of allocating a map if you might only want to change one of the keys.

I also believe my change is more in keeping with the current Logrus API since JSONFormatter already has a string attribute. Forcing users to create a map just to pass in values when they already can pass in timestamp seems a bit overly complex. Please let me know if I've missed the mark on my comments, but we've been using my fork of this in production for months now with no issue and the API seems to make sense to me.

stevvooe:
I'm not sure if I correctly described the issue with "struct bloat". While the extra fields do require extra memory (string header per field), the extra top-level fields make understanding their role harder. The issue is that they aren't grouped by purpose. We can usually see this represented by the shared suffix "Key". By calling it a `FieldMap` or something similar, most users can look at it and move on with their without having to understand what three fields do.

> As a nit, memory wise, your method is actually taking up slightly more space since you allocate space on the heap for the constants (even if the user does not want to configure the key) as well as the computational overhead (however small) of allocating a map if you might only want to change one of the keys.
 
This heap space is already taken up by the constants in the method. While the proposed map-based implementation uses less memory without any field mapping, this should only be allocated once per process, so I wouldn't factor this into the design decision.

> I also believe my change is more in keeping with the current Logrus API since JSONFormatter already has a string attribute. Forcing users to create a map just to pass in values when they already can pass in timestamp seems a bit overly complex.

The string argument is a string because it is a string argument. Nothing changes for specifying the formatter:

```go
JSONFormatter{
  TimestampFormat: time.RFC3339Nano,
  FieldMap: FieldMap{
    FieldKeyLevel: "@level",
    FieldKeyTime: "@timestamp",
  }
}
```

Here, we are mapping a set of fields, which, naturally, can be represented by a map.

In addition, the if statements can replaced by the following:

```go
func (f FieldMap) resolve(key fieldKey) string {
  key, ok := f[key]
  if ok {
    return key
  }

  return string(key)
}

// ... later in the function
data[f.FieldMap.resolve(FieldKeyLevel)] = entry.Level.String()
```

Another benefit is that one define commonly used field mappings as defaults:

```go
var LogstashFieldMap = FieldMap{
  FieldKeyLevel: "@level",
  FieldKeyTimestamp: "@timestamp",
  FieldKeyMessage: "@message",
}

var CompactFieldMap = FieldMap{
  FieldKeyLevel: "l",
  FieldKeyTimestamp: "t",
  FieldKeyMessage: "m",
}

jr := &JSONFormatter{
  FieldMap: CompactFieldMap,
}
```

Furthermore, any addition of new keys will automatically be supported due to the strong typing and map-based approach.

onetwopunch:
I agree that the if statements can be cleaned up a bit. I tend to follow the rule of 4: if you repeat yourself 4 times that's when you should abstract things away, but not before. Trying to make things DRY tends to complicate things. But I have no issue changing that. 

TBH, I still feel like forcing the user to create a FieldMap seems needlessly complex-looking but at this point it seems like bike-shedding over a very simple fix. I think your ideas are good, but they may be an over-design considering there's not likely going to be other keys for the logger. Again, if you're set on that particular implementation, I'll change it to get this PR merged.

onetwopunch:
@stevvooe I updated the code to include the resolveKey method which cleans up the method a lot. Since there's no reason to have more than those 3 keys, it would seem that the added complexity of a field map where the default key is implicit as the key itself, is not only difficult to understand upon reading but requires more levels of nesting for our users to call.

I'd argue that an explicit API like the one I've written is much simpler, cleaner and easier to understand. Please let me know if what you think about the latest changes.

stevvooe:
> Since there's no reason to have more than those 3 keys, it would seem that the added complexity of a field map where the default key is implicit as the key itself, is not only difficult to understand upon reading but requires more levels of nesting for our users to call.

Please, give users more credit. This is simple stuff and basic Go.

It is a very strong statement to say that there is "no reason" since you cannot predict the future. You're assuming that there will always be three keys or this field map will only ever operate on the default fields. This proposed approach allows future extension and future fields without making the data structure more complex or breaking compatibility.

At this point, I may just carry this changeset myself, as I don't really have time to debate this.

onetwopunch:
@stevvooe Ok if you feel that strongly about it, I'll make the changes to your specifications. It doesn't really make that big of a deal to me.

onetwopunch:
@stevvooe The latest commit has addressed all the changes you wanted.

json_formatter.go
```diff
@@ -4,10 +4,27 @@ import (
...
+		return k
+	}
+
+	return string(key)
+}
 
 type JSONFormatter struct {
 	// TimestampFormat sets the format used for marshaling timestamps.
 	TimestampFormat string
+	FieldMap        FieldMap
```
stevvooe: Let's add a documentation comment for this field.

onetwopunch: Good call

stevvooe:
LGTM after adding a documentation comment.

> I also apologize if I came across argumentative, that wasn't my intention

No need to apologize. We disagreed. We stated our positions. Coming across argumentative is par for the course when effectively discussing engineering matters. In fact, I appreciate the discourse, when pursued in a civil manner.

I do have to apologize about my tone in saying "I don't really have time to debate this." That could be better said as, "This discussion has ran its course", as in all arguments have been presented.

And to be honest, I agree with many sentiments at the base of your arguments ("Trying to make things DRY tends to complicate things", in particular) but I simply disagree with their application in this particular case.

Thank you for pursuing this change! Based upon the number of open issues and PRs related to this, it seems like this is a much needed feature!

Issue #364, opened 2016-05-31
Concurrent map errors


tejasmanohar:
I've been running into this error while using logrus, especially with larger log payloads. Not sure if this has the same root cause of #295 or not.

```
fatal error: concurrent map read and map writeINFO[0035]
​
goroutine 44 [running]:
runtime.throw(0x92f2a0, 0x21)
  /usr/local/Cellar/go/1.6.2/libexec/src/runtime/panic.go:547 +0x90 fp=0xc82066f7e0 sp=0xc82066f7c8
runtime.mapaccess1_faststr(0x6e6de0, 0xc820149350, 0x888940, 0xa, 0xc82066f980)
  /usr/local/Cellar/go/1.6.2/libexec/src/runtime/hashmap_fast.go:202 +0x5b fp=0xc82066f840 sp=0xc82066f7e0
github.com/Sirupsen/logrus.(*TextFormatter).printColored(0xc82011cac0, 0xc8204abc70, 0xc82041bcc0, 0xc820896c80, 0x5, 0x5, 0x910460, 0x19)
  .../Godeps/_workspace/src/github.com/Sirupsen/logrus/text_formatter.go:120 +0x439 fp=0xc82066fa38 sp=0xc82066f840
github.com/Sirupsen/logrus.(*TextFormatter).Format(0xc82011cac0, 0xc82041bcc0, 0x0, 0x0, 0x0, 0x0, 0x0)
  .../Godeps/_workspace/src/github.com/Sirupsen/logrus/text_formatter.go:81 +0x566 fp=0xc82066fbd8 sp=0xc82066fa38
github.com/Sirupsen/logrus.(*Entry).Reader(0xc82041bcc0, 0xc81c54f402, 0x0, 0x0)
  .../Godeps/_workspace/src/github.com/Sirupsen/logrus/entry.go:44 +0x67 fp=0xc82066fc70 sp=0xc82066fbd8
github.com/Sirupsen/logrus.Entry.log(0xc820131400, 0xc820149350, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x22102, 0xc8208c2ea0, ...)
  .../Godeps/_workspace/src/github.com/Sirupsen/logrus/entry.go:94 +0x296 fp=0xc82066fd98 sp=0xc82066fc70
github.com/Sirupsen/logrus.(*Entry).Error(0xc820119440, 0xc82066fea0, 0x1, 0x1)
[...]

PR #398, opened 2016-08-19
Fix race condition in hooks that modify entry.Data by guarding access

iancmcc:
Fixes Sirupsen/logrus#364
I'm not sold on the aesthetics of the implementation, but it's the only way I can think of doing it without breaking existing hooks.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -42,6 +42,8 @@ type Entry struct {
 
 	// When formatter is called in entry.log(), an Buffer may be set to entry
 	Buffer *bytes.Buffer
+
+	mu *sync.RWMutex
 }
 
 func NewEntry(logger *Logger) *Entry {
@@ -49,6 +51,7 @@ func NewEntry(logger *Logger) *Entry {
 		Logger: logger,
 		// Default is three fields, give a little extra room
 		Data: make(Fields, 5),
+		mu:   &sync.RWMutex{},
 	}
 }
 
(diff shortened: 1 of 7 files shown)


iancmcc:
We can't guarantee hooks won't deadlock, unfortunately, so I've changed it to include a func called `entry.SetField(k , v)` that sets values in a protected way, for use in hooks.

DsXack:
👍

stevvooe:
Wouldn't it be better to treat `Entry.Data` as immutable. If a hook wants to change a value, it should copy the whole map and add an entry. It will be a lot safer in practice and basic operations won't be encumbered by locks.

DmitriyMV:
Relevant #529

danielzheng:
same issue in v1.3.0, panic: 
fatal error: concurrent map iteration and map write
github.com/sirupsen/logrus/entry.go:108

dgsb:
@danielzheng can you provide any hints on how to reproduce thie error ?
This PR is quite old and some mutex protection has been added since in the master branch, I wonder if this PR would still fix the issues you've found.

danielzheng:
> @danielzheng can you provide any hints on how to reproduce your error ?
> This PR is quite old and some mutex protection has been added since in the master branch, I wonder if this PR would still fix the issues you've found.

We found it's a bug in middleware, it's OK now.

freeformz:
There are many conflicts and it's very old so I'm going to close this.

But, if this is something that still needs to be pursued, please resolve the conflicts and re-open the PR. Thanks!

PR #399, opened 2016-08-23
TextFormatter: brighten up the blue

pkieltyka:
Logrus is amazing, but I find the current blue color used by the TextFormatter to be near impossible to read. I've updated it to a bit more of a cyan, but I think it can pass as a blue. See screenshots below.

Before:
screen shot 2016-08-23 at 9 56 08 am

After:
screen shot 2016-08-23 at 9 56 29 am

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -14,7 +14,7 @@ const (
 	red     = 31
 	green   = 32
 	yellow  = 33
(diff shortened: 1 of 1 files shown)


dmathieu:
Thank you. This does make the text easier to read.

PR #402, opened 2016-08-26
Enable override of location for time field

nelz9999:
This is a fully backwards-compatible update that allows users to define which `time.Location` (https://golang.org/pkg/time/#Location) to log the timestamps in.

Example: even though my box is running in Pacific Time, I want my log lines to always be in UTC. I can do this by creating the formatter like this: `&JSONFormatter{Location: time.UTC}`

### DIFF
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -40,6 +40,8 @@ type JSONFormatter struct {
 	//    },
 	// }
 	FieldMap FieldMap
+
+	// Location overrides the default timestamp location (roughly: timezone)
 	Location *time.Location
 }
 
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -54,6 +54,9 @@ type TextFormatter struct {
 	// that log extremely frequently and don't use the JSON formatter this may not
 	// be desired.
 	DisableSorting bool
+
+	// Location overrides the default timestamp location (roughly: timezone)
+	Location *time.Location
 }
 
 func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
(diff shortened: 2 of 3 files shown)


nelz9999:
Travis failures seem to be around downloading of some dependencies...?

stevvooe:
@nelz9999 Rebase and it should pull in some build fixes.

It would be good to make this change across the TextFormatter and the JSONFormatter, alike.

nelz9999:
@stevvooe Updated, and applied to TextFormatter too!

stevvooe:
LGTM

cc @sirupsen

sirupsen:
Why `Location` over `Timezone`? 👀

nelz9999:
@sirupsen Because that is the type of the underlying object (https://golang.org/pkg/time/#Location).

stevvooe:
@nelz9999 @sirupsen Maybe `TimestampLocation` and group it with the format field?

mzuneska:
I have a custom `JSONFormatter` that is exactly the same except the timestamp is always set to UTC and I'd like to get rid of it. Can I do anything to help push this forward?

stevvooe:
> Why Location over Timezone? :eyes:

Not sure if I addressed this before. The problem is that what we call a timezone is actually a regional location. Taking `America/Los Angeles`, the switching behavior of the timezone, is actually defined by the governments that cover that region. They'll set decrees for the actual timezones like PST shall be this date and PDT will be this date. These dates might differ from what is in `America/Vancouver`.

Go's `time` package diffentiates this by having the concept of a [`FixedZone`](https://golang.org/pkg/time/#FixedZone) and a location. Matching this is probably good practice.

nelz9999:
I no longer have the desire or motivation to shepherd this along. If anyone else would like to take on this cause, feel free.

thechriswalker:
I was looking for this functionality, specifically forcing log to be UTC without having to remember to set or adjust in code the `TZ` environment variable, which is probably not portable anyway.

Then I realised it can easily be done with a Higher Order `logrus.Formatter` like this: https://gist.github.com/thechriswalker/ae6cbd163219934f52f36e896c7170c9
Then I don't have to implement it in each Formatter I use, it does modify the `log.Entry` but the function `time.UTC()` idempotent and if you want UTC times anyway...

Just sharing in case it is of use to anyone and the `EnforceUTC` / `EnforceLocation` option doesn't make it into the core.

PR #405, opened 2016-08-29
Ignore example hook to avoid failing test when Go1.5 or lower

evalphobia:
Since `gopkg.in/airbrake/gobrake.v2` updated to use new Go version's feature, tests is failing  in Go1.5 or lower.
So skip `go get` in old go version to add build tags.


evalphobia:
Check this PR first for other PR's CI testing. 🍒

PR #407, opened 2016-08-31
Added option to disable level text truncation in default text formatter

roganartu:


fermin-silva:
Looking forward to this 👍 

Normally people grep for "ERROR" in log files, so printing "ERRO" or "FATA" is not ideal.
I'd add the possibility to justify all log levels as well, so the log lines are alligned (i.e. adding one more space to INFO), but it's just a nice to have.

stevvooe:
Would it be better to change the default truncation to 5 characters?

```
INFO_
FATAL
ERROR
DEBUG
```

Would this break anyone?

fermin-silva:
My TOC would trigger if this ends up showing `WARNI` for warnings. If it prints `WARN_` then it's probably fine

roganartu:
I would prefer to not change the current default behaviour, and instead hide this change behind a configurable option.

Changing the truncation to 5 characters is conceptually simpler, but presents several problems:
1. How to handle padding out `INFO`. The underscore isn't very nice imo, and padding with a space probably makes more sense.
2. Truncate or pad `WARNING`? Either truncate to `WARN` with padding or truncate to WARNI. The former requires extra logic, and the latter just replaces `ERROR` -> `ERRO` with `WARNING` -> `WARNI`.
3. By just changing the default truncation there is no recourse for users to change back to the existing behaviour if they want. I've been on the receiving end of changes like that before and it kinda sucks.

Since this proposal was for a configuration change I'm going to update the PR to include a section in the README too.

roganartu:
Fixed the conflicts since this has been sitting for a while.
@sirupsen is there anything I need to do to get this merged? I'd like to get this closed out so I can stop maintaining a fork just for having this in my own projects.

stevvooe:
> How to handle padding out INFO. The underscore isn't very nice imo, and padding with a space probably makes more sense.

The underscore was just to demonstrate spacing. Just leave at 4, with a space.

Either way, LGTM

lynncyrin:
Hey @roganartu @fermin-silva @stevvooe I'm invested in following up this PR with https://github.com/sirupsen/logrus/issues/959, let me know if you're in favor 👍 ?

PR #409, opened 2016-08-31
Add additional fields to logger that will be present on all new entries

stillinbeta:
We have a bunch of things we want to go into every log line: App name, region app is running in, server name, etc. The existing solution is to Create an Entry and distribute that throughout your code in place of a logger. Howere, that requires all methods than previously took a Logger to take an Entry, which is non-intuitive (at least to me)

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -1,6 +1,7 @@
 package logrus
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"sync"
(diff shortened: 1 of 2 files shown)


nelz9999:
👍  I've nosing around a bit to see if something like this already existed. Thanks for submitting it!

stillinbeta:
Those failing checks don't appear to be related to my change, some problem with the airbrake dependency maybe?

PR #419, opened 2016-09-28
Fix formatting of wrapped errors when colors are used

aaronlehmann:
There are two different code paths for rendering a key/value pair. The
non-color version uses a type switch that handles specific types such as
"error", and the color version uses the %+v printf format specifier.
This causes an inconsistency between the two formats. In particular,
errors created using the github.com/pkg/errors package will include a
stack trace of where the error was created when printed to the terminal,
but not to a file. Printing the stack trace as part of the log field is
probably not the right behavior.

The output is also inconsistent between the two forms because strings
are not quoted/escaped when colors are used. This can make log output
unparseable.

Fix this by making both code paths use the type switch and escaping
rules. Fix the escaping code to pass the error value to Fprintf, not the
error itself, which seems to be necessary to avoid blank output with
errors created by github.com/pkg/errors.

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -122,7 +122,8 @@ func (f *TextFormatter) printColored(b *bytes.Buffer, entry *Entry, keys []strin
 	}
 	for _, k := range keys {
 		v := entry.Data[k]
-		fmt.Fprintf(b, " \x1b[%dm%s\x1b[0m=%+v", levelColor, k, v)
+		fmt.Fprintf(b, " \x1b[%dm%s\x1b[0m=", levelColor, k)
+		f.appendValue(b, v)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @Sirupsen could you have a peek at this one? :heart:

Sirupsen:
@aarongreenlee

thaJeztah:
ping @aarongreenlee (sorry for the repeated ping, but would be great if this was merged before our docker 1.13 code freeze)

Sirupsen:
Released as part of 0.11.0

thaJeztah:
Thanks @Sirupsen ❤️ !

Sirupsen:
@thaJeztah @aluzzardi is anyone from Docker Inc interested in helping out with some ownership of Logrus? 🎉

thaJeztah:
@Sirupsen I can discuss the option 👍

Sirupsen:
@thaJeztah awesome, let me know! I feel so guilty for not being able to respond quicker here.

stevvooe:
@Sirupsen I can take it on.

Sirupsen:
@stevvooe <3 <3 <3 <3 <3 <3 <3

Sirupsen:
I've added you both as collaborators.

PR #420, opened 2016-10-03
fix error with fields not being prefixed if it clashes with time, level and msg

mattias-lundell:


dgsb:
No longer valid. This bug has already been fixed in another commit.

PR #423, opened 2016-10-11
"Print" function should not be the alias of "Info" function.

usk81:


stevvooe:
What problem does this address?

PR #426, opened 2016-10-12
[messed up] Update to stop printing timestamp if disabled (in color)

puddingfactory:
Please see #427


puddingfactory:
Please see #427 instead

PR #400, opened 2016-08-24
Allow disabling timestamps with colored output

carolynvs:
When using colored output, `DisableTimestamps` doesn't work.

**Sample Application**

```
package main

import "github.com/Sirupsen/logrus"

func main() {
    l := logrus.New()
    l.Formatter = &logrus.TextFormatter{
        DisableTimestamp: true,
        ForceColors:      true,
    }

    l.Info("I'm a little teapot")
}
```

**Expected Output**

```
INFO I'm a little teapot
```

**Actual Output**

```
INFO[0000] I'm a little teapot
```

PR #427, opened 2016-10-12
Update to stop printing timestamp if disabled and in color

puddingfactory:
## Goal

I don't want the timestamp to be printed, but do want to keep the excellent color/formatting.
## Problem

While there is an option for `DisableTimestamp` in `logrus.TextFormatter` struct, I found that this will **only** exclude the timestamp if printing **without color**. 
## Solution

Add one other check for `DisableTimestamp` within `TextFormatter.printColored` method.


puddingfactory:
This is a duplicate of #400 which I really should've noticed before. Sorry for submitting a duplicate PR

PR #430, opened 2016-10-18
Flip stdout and stderr

hatchan:
The default was changed to os.Stderr in f8f08842cce7ed05c296430852f60f287458c529

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,8 +81,8 @@ func init() {
   // Log as JSON instead of the default ASCII formatter.
   log.SetFormatter(&log.JSONFormatter{})
 
-  // Output to stderr instead of stdout, could also be a file.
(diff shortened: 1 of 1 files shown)


stevvooe:
Are you saying that because the default is `os.Stderr`, the example should show how to change it to `os.Stdout`?

hatchan:
@stevvooe Yes indeed. The default is `os.Stderr` but the readme hints that it is actually `os.Stdout`.

hatchan:
We could the text  to explicitly say that `os.Stderr` is the default. Something like:

> // Output to stdout instead of stderr (default), could also be a file.

stevvooe:
LGTM

sirupsen:
Fixed directly in master

PR #432, opened 2016-10-19
Remove TTY checking from text formatter

### LINKED ISSUE #431: Characters are escaped based on TTY in TextFormatter
The `TextFormatter` will escape `\t` and `\n` if `Out` is not a TTY.  The only way to avoid this and maintain the formatting is to set `ForceColors`, but this will result in non TTY Out instances to include the escaped color color codes.  

I have a change that splits this escaping and terminal checking out of `Format()` and allows the user to set the appropriate values directly or use the helper to create a Formatter bases on the terminal with the appropriate values set.

fdawg4l:
Moved TTY checking and default setting for the TextFormatter to the
NewTextFormatter() method.  This will set the `Color` bool in the
formatter to emit colors.  Likewise all bools can now be toggled from
the TextFormatter struct directy regardless of Out being
a TTY or not.

Fixes #431

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,7 @@ ocean","size":10,"time":"2014-03-10 19:57:38.562264131 -0400 EDT"}
 "time":"2014-03-10 19:57:38.562543128 -0400 EDT"}
 ```
 
-With the default `log.SetFormatter(&log.TextFormatter{})` when a TTY is not
-attached, the output is compatible with the
+With `log.SetFormatter(&log.TextFormatter{Colors: false})`, the output is compatible with the
 [logfmt](http://godoc.org/github.com/kr/logfmt) format:
 
 ```text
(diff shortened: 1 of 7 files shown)


fdawg4l:
Looks like the Travis failure is due to `go get` failing for previous `Go` versions.  I'm not sure that's due to this change but please ping if I borked something inadvertently.

stevvooe:
This PR represents a lot of backwards incompatible changes won't be mergeable.

fdawg4l:
@stevvooe Can you be more specific?  What parts break backwards compatibility?  (Also, what does backwards compatibility mean here?)

stevvooe:
@fdawg4l You've changed several existing fields names, such that it will break users' builds. For example, if my code was using the `FullTimestamp` field, it would now break with this change.

fdawg4l:
Hmm...  Good point.  But some of the variables as-is are redundant.  For instance-
```
ForceColors bool
DisableColors bool

DisableTimestamp bool
FullTimestamp bool
TimestampFormat string
```

I can split up the chance and leave out the simplification for now.  The nonescapable characters conflated with tty handling was really the only thing that matters in this change.  Will close this and repush.

PR #433, opened 2016-10-19
Update README.md

rybit:
add the nats hook that I wrote


sirupsen:
Added directly in master due to merge conflict

PR #442, opened 2016-11-05
Cleaned up README

johnlonganecker:
```
- split up contextLogger example
- added missing import to example
- Alphabatized list of hooks
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ application, you can also create an instance of the `logrus` Logger:
 package main
 
 import (
+  "os"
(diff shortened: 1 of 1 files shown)


johnlonganecker:
I am not sure why this failed. In that commit I modified one line in a README.md file.

stevvooe:
@johnlonganecker Looks like it failed on an upstream dependency when building with Go <= 1.5. Filed https://github.com/Sirupsen/logrus/issues/445.

PR #450, opened 2016-11-30
Add optional log field with calling method

dclendenan:
To simplify tracing of similar error messages, it can be useful to have the calling method of a log function be logged.  This adds a flag which enables that.  Off by default, to avoid impact to existing users who may be using the 'method' field in logging for other purposes..


entry.go
```diff
@@ -85,14 +96,46 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
...
+
 // This function is not declared with a pointer value because otherwise
 // race conditions will occur when using multiple goroutines
 func (entry Entry) log(level Level, msg string) {
 	var buffer *bytes.Buffer
 	entry.Time = time.Now()
 	entry.Level = level
 	entry.Message = msg
-
+	if ReportCaller() {
```
stevvooe: This looks up the attribute on the std logger, which may be an incorrect assumption for this particular entry.

entry.go
```diff
@@ -85,14 +96,46 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
...
+			if fullFuncName == "main.main" {
+				return "main"
+			} else {
+				return fullFuncName
+			}
+		}
+	}
+
+	// if we got here, we failed to find the caller's context
+	return "UNKNOWN_CALLER"
```
stevvooe: Just use empty string or an error value to signal this.

entry.go
```diff
@@ -85,14 +96,46 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 	return &Entry{Logger: entry.Logger, Data: data}
 }
 
+// getCaller retrieves the name of the first non-logrus calling function
+func getCaller() (method string) {
+	// Restrict the lookback to 25 frames - if it's further than that, report UNKNOWN
+	pcs := make([]uintptr, 25)
+
+	// the first non-logrus caller is at least three frames away
```
stevvooe: This will fail if the calling package is also called `logrus`, which might be common in a local project wrapper. It might be better to parameterize the number of calls from the `getCaller()` function.

entry.go
```diff
@@ -85,14 +96,46 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
...
+	depth := runtime.Callers(3, pcs)
+	for i := 0; i < depth; i++ {
+		fullFuncName := runtime.FuncForPC(pcs[i]).Name()
+		idx := strings.LastIndex(fullFuncName, "/") + 1
```
stevvooe: Don't mangle the name result. Just use it directly. It contains useful information, like the actual package path and possibly the receiver.

stevvooe:
We implemented a similar feature for the Docker Registry in https://github.com/docker/distribution/blob/master/context/trace.go#L46. It is very similar so there might be some tidbits that can help in this implementation.

It might be good to have a benchmark for this to measure overhead of enabling source tracing.

entry.go
dclendenan: that adds significant log volume - perhaps making the truncation optional would better suit?

stevvooe: @dclendenan Indeed, but logging the function name is significant enough.

My main issue here is that this is very ad-hoc. There might be a more canonical way to decompose into function and local package.

dclendenan:
@stevvooe do the changes since your last review satisfy your concerns?
thanks!

json_formatter.go
```diff
@@ -9,9 +9,10 @@ type fieldKey string
 type FieldMap map[fieldKey]string
 
 const (
-	FieldKeyMsg   = "msg"
-	FieldKeyLevel = "level"
-	FieldKeyTime  = "time"
+	FieldKeyMsg    = "msg"
+	FieldKeyLevel  = "level"
+	FieldKeyTime   = "time"
+	FieldKeyCaller = "method"
```
stevvooe: This should be `FieldKeyFunc = "func"`. The caller may not always be a method.

stevvooe:
@dclendenan This looking almost there, but need to make sure the implementation is correct. I have a few more nits to fix up, in the mean time.

@sirupsen Any concerns on this one?

entry.go
```diff
@@ -85,14 +102,70 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
...
+		minimumCallerDepth = knownLogrusFrames
+	}
+
+	for i := 0; i < depth; i++ {
+		fullFuncName := runtime.FuncForPC(pcs[i]).Name()
+		pkg := getPackageName(fullFuncName)
+
+		// If the caller isn't part of this package, we're done
+		if pkg != logrusPackage {
+			if fullFuncName == "main.main" {
```
stevvooe: Again, let's avoid these kind of rewrites. This makes the output inconsistent, since the function may be `main` in the `main` package.

formatter.go
```diff
@@ -30,7 +31,7 @@ type Formatter interface {
 //
 // It's not exported because it's still using Data in an opinionated way. It's to
 // avoid code duplication between the two default formatters.
-func prefixFieldClashes(data Fields) {
+func prefixFieldClashes(data Fields, reportCaller bool) {
```
dclendenan: fieldKey only exists in the JSON formatter right now, but even if that was ported into the text formatter the Fields passed in don't include whether the calling function will be added later - it just has the fields passed via WithField(s).  

This remapping actually looks like it may have issues if the field-name override is used, since the protection added is only for the original names and not the overrides.  I'd be open to looking at bringing the field mapping into the text formatter and then making prefixFieldClashes() handle this case, but it doesn't seem like that should be conflated with my current change.

The condition here was to ensure that if current users are making use of the field we use for the calling function (now 'func') and are not setting ReportCaller, then their logging isn't changed.  Refactors to accomplish this without adding the boolean all seemed much more more significant, and it didn't seem that they were justified.  If you see something I'm missing, please call it out.

dclendenan:
@stevvooe @sirupsen I've caught this up with other changes, what's the next step here?

Thanks!

dclendenan:
Just checking in here - @stevvooe, @sirupsen any further issues with this PR?

stevvooe:
@dclendenan I'm sorry, I've been a little distracted. I'll defer to @sirupsen on this one.

In the mean time, do you have an example log trace with this enabled in a few different calling paths? For example, with interface methods and function pointers, etc. to make sure the formatting looks good.

dclendenan:
@stevvooe my turn to get distracted, sorry for lag here.  I've added a few more tests, but in general the formatting will match that added to the README: https://github.com/dclendenan/logrus#logging-method-name

If your package structure is more complex then the referenced method path will get more complex to  match, eg;
func="my.github.host/org/package/submodule.FunctionA"
func="my.github.host/org/package/submodule.(*Object).FunctionB"

stevvooe:
@dclendenan Thanks! That looks fine.

I'll defer to @sirupsen for a merge on this one.

dclendenan:
Thanks @stevvooe 

@sirupsen do you have any further comments or are we good to merge this?

dclendenan:
caught up with other upstream changes - @sirupsen ready for merge?

Cheers!

Dave

dclendenan:
@stevvooe regarding discussion in https://github.com/sirupsen/logrus/issues/63 - if this PR looks like a good direction then should we merge it and then I can look at extending what I've done with the additional information?  I'm a little leery of continuing to extend a PR that's been open as long as this one has :)

cc: @sirupsen

kotuk81:
2 years for a merge?

stevvooe:
@dclendenan I have merge rights here but this is not my primary job. My github notifications are pretty much overrun and I'm sorry I let it rot. If you want to try one more time, just tweet at me and I'll take a look to get this merged.

dgsb:
up

dgsb:
@dclendenan thanks for your contribution. There are some conflict with master, I will take care of the merge.

dclendenan:
@dgsb Thanks!  Sorting out that conflict kept getting pushed down my todo list, but it'd be great to see this finally merge!

dgsb:
merge resolution has been done on #850

PR #453, opened 2016-12-08
Do not add 'FieldKey*'  field when they are void.

KerwinKoo:
Set the default field key to "", eg. :

        logrus.FieldKeyLevel:   ""

    it will create an error format JSON string
    as "":"info". So add FieldKEy length checking before do json.Marshal(data)

    It alse can using th default key-name that defined in previous:

    const (
        FieldKeyMsg   = "msg"
        FieldKeyLevel = "level"
        FieldKeyTime  = "time"
    )

    But I think developer who set these fields to "" wants to hide or conceal the key-value that definde in default JSONFormatter.

    Sorry for my poor English :D

### DIFF
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -60,12 +60,19 @@ func (f *JSONFormatter) Format(entry *Entry) ([]byte, error) {
 		timestampFormat = DefaultTimestampFormat
 	}
 
-	if !f.DisableTimestamp {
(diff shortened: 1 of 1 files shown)


KerwinKoo:
I'm sorry for the debug info in source code that pushed. 

Merge request later.

PR #454, opened 2016-12-08
Do not add 'FieldKey*'  field when they are void.

KerwinKoo:
Set the default field key to "", eg. :

```
logrus.FieldKeyLevel:   ""
```

it will create an error format JSON string as `"":"info"`. So add `FieldKey` length checking before do json.Marshal(data).

It alse can using the default key-name that defined in previous:

```
        const (
            FieldKeyMsg   = "msg"
            FieldKeyLevel = "level"
            FieldKeyTime  = "time"
        )
```

But I think developer who set these fields to "" wants to hide or conceal the key-value that definde in default JSONFormatter.

Sorry for my poor English :D

### DIFF
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -60,11 +60,17 @@ func (f *JSONFormatter) Format(entry *Entry) ([]byte, error) {
 		timestampFormat = DefaultTimestampFormat
 	}
 
-	if !f.DisableTimestamp {
+	if !f.DisableTimestamp && len(f.FieldMap[FieldKeyTime]) > 0 {
 		data[f.FieldMap.resolve(FieldKeyTime)] = entry.Time.Format(timestampFormat)
 	}
-	data[f.FieldMap.resolve(FieldKeyMsg)] = entry.Message
-	data[f.FieldMap.resolve(FieldKeyLevel)] = entry.Level.String()
+
+	if len(f.FieldMap[FieldKeyMsg]) > 0 {
+		data[f.FieldMap.resolve(FieldKeyMsg)] = entry.Message
+	}
+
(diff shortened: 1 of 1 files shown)


json_formatter.go
```diff
@@ -60,11 +60,17 @@ func (f *JSONFormatter) Format(entry *Entry) ([]byte, error) {
...
 	}
 
-	if !f.DisableTimestamp {
+	if !f.DisableTimestamp && len(f.FieldMap[FieldKeyTime]) > 0 {
 		data[f.FieldMap.resolve(FieldKeyTime)] = entry.Time.Format(timestampFormat)
 	}
-	data[f.FieldMap.resolve(FieldKeyMsg)] = entry.Message
-	data[f.FieldMap.resolve(FieldKeyLevel)] = entry.Level.String()
+
+	if len(f.FieldMap[FieldKeyMsg]) > 0 {
```
stevvooe: Just compare with `""`:

```
if key := f.FieldMap.resolve(FieldKeyMsg); key != "" {
  data[key] = entry.Message
}
```

Also, note that you need to use `resolve` to get the actual value for the key.

dgsb:
These FieldKey* are constants these values can not be changed. I don't think we need this fix.

gguridi:
If you really need something soooo specific I think the way to go is to create your own Formatter for your specific needs. In this PR we are linking the logic between showing/not showing certain logrus-populated properties with the fact that the user changes the name of them to "empty". I honestly wouldn't do it.

freeformz:
Thank you, but in order to clean up older PRs and issues (this PR hasn't had any activity in ~ 1 year), I'm going to close this PR. Feel free to re-open with an updated branch if this is still of interest. PS: I'm sorry this sat for so long.

Issue #451, opened 2016-12-01
Case change breaks builds?


lonelycode:
Not sure what's started to happen, but overnight our builds have started to break:

`can't load package: package github.com/TykTechnologies/tyk: case-insensitive import collision: "github.com/Sirupsen/logrus" and "github.com/sirupsen/logrus"`

I wonder if it's the case change from `S` to `s` in the github username of the account. Even if we change the import path to use the lowercase version, dependencies that use the uppercase version that are not under our control seem to need updating too.

Vendoring the lib doesn;t seem to work either, as the log.WithFields seems to self referential, but maybe I was too hasty in dismissing vendoring:

`./api.go:124: cannot use "github.com/TykTechnologies/tyk/vendor/github.com/Sirupsen/logrus".Fields literal (type "github.com/TykTechnologies/tyk/vendor/github.com/Sirupsen/logrus".Fields) as type "github.com/Sirupsen/logrus".Fields in argument to log.WithFields`

Any tips appreciated.

lonelycode:
Disregard the vendoring bit, I think that's our centralised logger screwing up.

cstockton:
I can't get gobgpd to build anymore either, but I'm having trouble tracking down why.

```bash
$ go get -v -u github.com/osrg/gobgp/gobgpd
github.com/osrg/gobgp (download)
package github.com/osrg/gobgp/gobgpd: case-insensitive import collision: "github.com/Sirupsen/logrus" and "github.com/sirupsen/logrus"
```

The error doesn't say where the culprit file is, a grep through ALL my dependencies show that the only package that imports a lower case version in all my dependencies is `Sirupsen/logrus`

```bash
$ grep -r "github.com/sirupsen/logrus" ~/dev/local/go-workspaces/dependencies/
src/github.com/Sirupsen/logrus/alt_exit_test.go:	"github.com/sirupsen/logrus"
... snip ....
src/github.com/Sirupsen/logrus/README.md:  logrus_syslog "github.com/sirupsen/logrus/hooks/syslog"
src/github.com/Sirupsen/logrus/README.md:| [Syslog](https://github.com/sirupsen/logrus/blob/master/hooks/syslog/syslog.go) | Send errors to remote syslog server. Uses standard library `log/syslog` behind the scenes. |
src/github.com/Sirupsen/logrus/README.md:  log "github.com/sirupsen/logrus"
```

[...]

PR #459, opened 2016-12-19
Fix casing

iheitlager:
Following up on #451, PR in three commits to separate different levels of relevance.

Some code in the hooks directory was introduced which had `Sirupsen` in uppercase again. This causes problems for projects with vendoring like `glide` and `govendor`. This PR is putting everything to lowercase completing #384.

Links to godoc and travis are validated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* fixed all references from uppercase `Sirupsen` to lowercase (#451)
+
 # 0.10.0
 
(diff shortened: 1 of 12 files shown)


iheitlager:
Seems like decision was made to stick to Uppercase, bummer .....

mnzt:
@iheitlager bringing this back up... 

@sirupsen what if the airbrake hook was moved out to its own repo? Therefore logrus could be vendored in whatever case the user likes.

PR #462, opened 2017-01-03
Preserve Message, Time, Logger and Level fields for new entry when using WithFields

manisenkov:
When using `WithFields` method on Entry instance the result doesn't keep `Level`, `Logger`, `Message` and `Time` fields from original entry instance.

**NOTE**: not sure if it's bug or it was done intentionally 🤔

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -82,9 +82,13 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 	for k, v := range fields {
 		data[k] = v
 	}
-	newEntry := *entry
-	newEntry.Data = data
-	return &newEntry
+	return &Entry{
+		Logger:  entry.Logger,
+		Data:    data,
+		Time:    entry.Time,
+		Level:   entry.Level,
+		Message: entry.Message,
+	}
 }
 
 // This function is not declared with a pointer value because otherwise
--- a/entry.go
+++ b/entry.go
@@ -82,7 +82,9 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 	for k, v := range fields {
 		data[k] = v
 	}
-	return &Entry{Logger: entry.Logger, Data: data}
+	newEntry := *entry
+	newEntry.Data = data
+	return &newEntry
 }
 
 // This function is not declared with a pointer value because otherwise
--- a/entry_test.go
+++ b/entry_test.go
@@ -85,12 +85,14 @@ func TestEntryWithFields(t *testing.T) {
 		Level:   DebugLevel,
 		Data:    Fields{"foo": "bar"},
 		Logger:  logger,
+		Buffer:  bytes.NewBuffer([]byte{}),
 	}
 	newEntry := entry.WithFields(Fields{"baz": 42, "one": "more"})
 	assert.Equal(t, entry.Time, newEntry.Time)
 	assert.Equal(t, entry.Message, newEntry.Message)
 	assert.Equal(t, entry.Level, newEntry.Level)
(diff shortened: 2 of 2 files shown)


entry.go
```diff
@@ -82,7 +82,9 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 	for k, v := range fields {
 		data[k] = v
 	}
-	return &Entry{Logger: entry.Logger, Data: data}
+	newEntry := *entry
```
stevvooe: This shallow copy will inadvertently share the buffer. I'm not sure if that is intended.

manisenkov: @stevvooe fixed

stevvooe:
I forgot to ask: why do you need it to preserve the level? Typically, you'll call these with `log.Debug` and the level will get set at that time.

manisenkov:
type formatter struct {
	base logrus.Formatter
	tag  string
}

func (f *formatter) Format(entry *logrus.Entry) ([]byte, error) {
	// If tag specified, it should be added to log entries as a field
	if f.tag == "" {
		return f.base.Format(entry)
	}
	newEntry := entry.WithField("tag", f.tag)

	// Without changes from this PR I need to manually copy all fields into new entry
	/*
	newEntry.Time = entry.Time
	newEntry.Level = entry.Level
	newEntry.Message = entry.Message
	*/

	return f.base.Format(newEntry)
}

func setupLogger(tag string) *logrus.Logger {
	logger := logrus.New()
	logger.Formatter = &formatter{
		base: logger.Formatter,
		name: tag,
	}
	return logger
}

manisenkov:
I try to implement some kind of wrapper around formatter to implicitly add fields to log entries. Example:
```go
type formatter struct {
	base logrus.Formatter
	tag  string
}

func (f *formatter) Format(entry *logrus.Entry) ([]byte, error) {
	// If tag specified, it should be added to log entries as a field
	if f.tag == "" {
		return f.base.Format(entry)
	}
	newEntry := entry.WithField("tag", f.tag)

	// Without changes from this PR I need to manually copy all fields into new entry
	/*
	newEntry.Time = entry.Time
	newEntry.Level = entry.Level
	newEntry.Message = entry.Message
	*/

	return f.base.Format(newEntry)
}

func setupLogger(tag string) *logrus.Logger {
	logger := logrus.New()
	logger.Formatter = &formatter{
		base: logger.Formatter,
		name: tag,
	}
	return logger
}
```

stevvooe:
@manisenkov `WithEntry` is really meant to be used be callers of the log package. In this case, I'd just recommend copying `Data` from the entry and adding your value.

manisenkov:
@stevvooe yes, I can go this way. I've just been confused by this behaviour, because I expect that source message will be copied into result entry

stevvooe:
@manisenkov That doesn't make a whole lot of sense for the common use case of `WithField`. Typically, we have this:

```
logger := log.WithField("foo", "bar")
logger.Debug("some message") // logged with fields
```

In the above, `logger` is actually an `Entry` type, but we don't use like a regular log entry. The Time, Level and Message aren't actually bound. It effectively acts as a prototype, for the future emission of log messages.

In the case of the formatter, after the log has been dispatched. It might be safer to just to avoid the allocations that come with `WithField` and set the field on `entry.Data`.

manisenkov:
@stevvooe Yes, agree. Thank you for the explanation ;)
I'll close this PR then.

stevvooe:
@manisenkov Cheers.

If that doesn't work out, let me know and we can another look at this PR.

Issue #457, opened 2016-12-18
miniTS in TextFormatter is inaccurate when hook takes a long time


at15:
## Problem

When using elapsed time instead of full time, it is calculated using the miniTS function

````go
func miniTS() int {
	return int(time.Since(baseTimestamp) / time.Second)
}
````

and `time.Since` is a shortcut for `time.Now().Sub(t)`, if there is no hook executed, this function is correct, since `entry.Time` would be the same as `time.Now()`. However, the formatter is called after all the hooks are executed, thus if the hooks took a long time to finish, it would formatter would output a wrong elapsed time. 

## Replay

i.e. the following code use a dummy hook that do nothing but sleep for 1 second when it is called

````go
package main

import (
	"time"

	"fmt"

	log "github.com/Sirupsen/logrus"
)

type DelayHook struct {
}

func (h *DelayHook) Levels() []log.Level {
	return []log.Level{log.InfoLevel}
}

func (h *DelayHook) Fire(entry *log.Entry) error {
	time.Sleep(1 * time.Second)
	return nil
}

func init() {
	h := &DelayHook{}
	log.AddHook(h)
}

func main() {
	fmt.Println("start")
	log.Info("1")
	log.Info("2")
	fmt.Println("end")
}
````
the output is 

````
start
INFO[0001] 1                                            
INFO[0002] 2                                            
end
````

but it should be

````
start
INFO[0000] 1                                            
INFO[0001] 2                                            
end
````

## Solution

A possible solution is to replace `mintTS` with `int(entry.Time.Sub(baseTimestamp)/time.Second)`, change https://github.com/sirupsen/logrus/blob/master/text_formatter.go#L119

````go
fmt.Fprintf(b, "\x1b[%dm%s\x1b[0m[%04d] %-44s ", levelColor, levelText, miniTS(), entry.Message)
````

to

````go
fmt.Fprintf(b, "\x1b[%dm%s\x1b[0m[%04d] %-44s ", levelColor, levelText, int(entry.Time.Sub(baseTimeStamp)/time.Second), entry.Message)
````

Also as mentioned by https://github.com/sirupsen/logrus/issues/435, maybe enable non-blocking execution of hook is a better way, also adding `Flush` may improve performance by batching when the `Writer` is a remote service that takes long round trip time, but this may require a lot of change in the code

btw: the travis badge in README have broken link due to case sensitive problem

stevvooe:
@at15 Please submit a PR.

[...]

PR #465, opened 2017-01-11
Remove miniTS in TextFormatter

at15:
Fixes https://github.com/sirupsen/logrus/issues/457

Remove `miniTS` and use  `int(entry.Time.Sub(baseTimestamp)/time.Second)` in `printColored` for accurate timing when hooks took a long time.


stevvooe:
LGTM

PR #469, opened 2017-01-21
Add MaxLevel constant

sargun:
This adds the max level constant. If we add new log levels, we just need to make sure that lastLevel remains..last.

### DIFF
--- a/logrus.go
+++ b/logrus.go
@@ -82,7 +82,14 @@ const (
 	InfoLevel
 	// DebugLevel level. Usually only enabled when debugging. Very verbose logging.
 	DebugLevel
+	// Not used. It's used to derive the const maxLevel in case
+	// someone needs to enumerate the levels.
+	lastLevel
 )
+
+// This can be used as the highest level possible, if you want to
+// iterate through all log levels, iterate through [0, MaxLevel].
+const MaxLevel Level = lastLevel - 1
 
 // Won't compile if StdLogger can't be realized by a log.Logger
 var (


sirupsen:
I'm curious what your use-case is for this?

sargun:
Iterating through all of the levels / validate an incoming level is valid while translating non-logrus log levels (number based) to logrus ones. It's slightly nicer to do than call Levels() and iterating through to check for membership.

logrus.go
```diff
@@ -82,8 +82,15 @@ const (
...
 	// DebugLevel level. Usually only enabled when debugging. Very verbose logging.
 	DebugLevel
+	// Not used. It's used to derive the const maxLevel in case
+	// someone needs to enumerate the levels.
+	lastLevel
 )
 
+// This can be used as the highest level possible, if you want to
+// iterate through all log levels, iterate through [0, MaxLevel].
+const MaxLevel Level = lastLevel - 1
```
sirupsen: Why not just do `const MaxLevel Level = DebugLevel`?

sargun: In general this is how I've done it in C programs, so it's obvious, and you don't need to change two lines when adding a new level.

freeformz:
The doc on MaxLevel says it makes it easier to iterate through [0, MaxLevel], but that is already the case easy with `AllLevels`.

```go
for i, level := range logrus.AllLevels {
 ...
}
```

Given that and since this has sat for over a 1.5 years in an unresolved state I'm going to close this. If this is still something you feel strongly about, please resolve the conflicts and re-open the PR. Thanks!

Issue #466, opened 2017-01-11
TextFormatter for non tty terminals


undiabler:
- everything works fine in terminal (local/ssh)
- I get logfmt output when running my program via supervisord
- `log.SetFormatter(&log.TextFormatter{})` seems do nothing

how can I skip auto checks and get textFormatter for non tty terminals?

undiabler:
Solved myself. 
It seems confusing but with `log.SetFormatter(&log.TextFormatter{ForceColors: true})` everything works as expected. 

Now I have new question. Is it realy good to overrite check if we run program in tty with ForceColors flag?
What about situation when we want normal output (not logfmt) but without colorls. Now it is impossible. 

Better way is:
- leveled variable for color
-- auto check if can use colors
-- enable colors manual
-- disable colors manual
- leveled variable for output
-- auto check temrinal (logfmt style/simple output)
-- simple output manual
-- logfmt output manual

sirupsen:
https://github.com/sirupsen/logrus/pull/471 likely fixes this

Issue #414, opened 2016-09-20
no way to write log to a file!


HomeGamer19880330:
logger := logrus.New()
fp, err = os.OpenFile(logFilePath, os.O_CREATE|os.O_APPEND|os.O_RDWR, 0666)
logger.Out = bufio.NewWriterSize(fp, maxFileSize)

code like this,I run the program happy

but!
I can not write log to file!why it is?

how could it possilbe???

HomeGamer19880330:
any way.... is there a example to show how to write log to a local file?thanks!

chennqqi:
@HomeGamer19880330 
My code may help you 

	logrus.SetLevel(logrus.ErrorLevel)
	var f *os.File
	var err error
	if f, err = os.Create(APPNAME + ".log"); err != nil {
		logrus.Error("creat log failed", err)
		return
	}
	//close log
	logrus.SetOutput(f)
	defer func(f *os.File) {
		f.Sync() //look here
		f.Close()
	}(f)

proudlily:
@chennqqi as far as i know,if using the way that you offered,we cant use Field into the logFile,only a sentence.

stevvooe:
`Logger.Out` should do the trick but I haven't confirmed it. Did you call flush on the buffered writer?

Looks like this is a duplicate of #410.

sirupsen:
https://github.com/sirupsen/logrus/pull/471 fixes this with a README entry

Issue #379, opened 2016-07-05
[Windows 10]: TTY not detected.


LukasMa:
_OS_: Windows 10 Pro 64bit
_Go version_: 1.6

On windows logrus always prints the text in the _logfmt_ format to stdout. It seems that it does not detect the tty. I tried it with msysgit, cmd and powershell.

aarongreenlee:
Thanks for reporting this issue.

tommyshem:
See issue #172 and #306(comment) for some ideas to get it working with a walk around.
Also look at this github project https://github.com/mattn/go-colorable

sirupsen:
Does https://github.com/sirupsen/logrus/pull/471 fix this?

cup:
https://github.com/sirupsen/logrus/issues/306

https://github.com/sirupsen/logrus/issues/379
https://github.com/sirupsen/logrus/issues/763

Issue #306, opened 2015-12-12
TextFormatter does not treat msys git bash shell as tty


jerrodrurik:
https://github.com/Sirupsen/logrus/blob/master/text_formatter.go#L73

This seems to be source of the error.

Sirupsen:
PR welcome

aybabtme:
@jerrodrurik if you have a patch for this, I'll be happy to review it.

aarongreenlee:
@jerrodrurik, would you be open to submitting a pull-request?

tommyshem:
I have done some tests in the msys git bash on windows and the isTerminal is returning false.
The runtime.GOOS is returning ="windows" which is correct as it is windows.

Not sure how to test for bash on windows.

He could try the forceColors flag to force the colour output on so it does not check for tty, not sure if it would work being bash on windows and not sure how to turn the forceColors flag on.

code snippet from the same file line 37.
https://github.com/Sirupsen/logrus/blob/master/text_formatter.go#L36

tommyshem:
Got it working on the msys git bash terminal by putting the force flag on as a temp fix.
Code below.

` var log = logrus.New()`

` func main() { `

      var testf = new(logrus.TextFormatter)
       testf.ForceColors = true
       log.Formatter = testf

      `log.Errorf("runtime system = " + runtime.GOOS)`
`}`
`

tommyshem:
Found this on Issue #172 

// import this library 
 import "github.com/shiena/ansicolor" 

 // force colors on for TextFormatter 
 log.Formatter = &logrus.TextFormatter{ ForceColors: true, }
 
 // then wrap the log output with it 
 log.Out = ansicolor.NewAnsiColorWriter(os.Stdout)

chappjc:
@tommyshem Thanks for the tip.  Another option:

``` go
import "github.com/mattn/go-colorable"

log.Formatter = &logrus.TextFormatter{ForceColors: true}
log.Out = colorable.NewColorableStdout()
```

sirupsen:
Thanks for the workaround @mattn and @chappjc, I've added this to the README in feda234

cup:
https://github.com/sirupsen/logrus/issues/306
https://github.com/sirupsen/logrus/issues/379
https://github.com/sirupsen/logrus/issues/763

PR #471, opened 2017-02-05
text_formatter: detect tty based on fd

sirupsen:
@stevvooe we have a lot of problems with Logrus' TTY detection. I think the root of this is that `log.Out` when set to not `stderr` the TTY detection logic will _still_ use `os.Stderr` to check for TTY... this doesn't make sense as many people will want the text formatter to a file. This PR changes this for all platforms.

The mutex here is ugly... but I'm afraid of the race here and CI will complain if I don't do it. Ideas, @stevvooe? It's simpler to piggy-back off of the lock in `Entry`, but that'd prevent being able to serialize with the formatters in parallel which is potentially a large performance hit. Alternatively we could check for TTY on every logging statement, but the context switch I'd assume is even more expensive.

This should resolve issues such as https://github.com/sirupsen/logrus/issues/466 https://github.com/sirupsen/logrus/issues/414 https://github.com/sirupsen/logrus/issues/379 https://github.com/sirupsen/logrus/issues/306

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -3,7 +3,6 @@ package logrus
 import (
 	"bytes"
 	"fmt"
-	"runtime"
 	"sort"
 	"strings"
 	"time"
@@ -20,12 +19,10 @@ const (
 
 var (
 	baseTimestamp time.Time
-	isTerminal    bool
 )
 
 func init() {
 	baseTimestamp = time.Now()
-	isTerminal = IsTerminal()
 }
 
 type TextFormatter struct {
@@ -50,6 +47,10 @@ type TextFormatter struct {
 	// that log extremely frequently and don't use the JSON formatter this may not
 	// be desired.
 	DisableSorting bool
+
+	// Whether the logger's out is to a terminal
+	isTerminal         bool
+	terminalDetermined bool
 }
 
 func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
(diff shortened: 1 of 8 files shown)


text_formatter.go
```diff
@@ -74,8 +77,19 @@ func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
...
+	isTerminal := f.isTerminal
+	f.terminalLock.RUnlock()
+
+	if !terminalDetermined {
+		f.terminalLock.Lock()
+		f.isTerminal = IsTerminal(entry.Logger.Out)
+		f.terminalDetermined = true
+		f.terminalLock.Unlock()
+	}
+
```
mlaventure: `isTerminal` needs to be set here since `terminalDetermined` will be false the first time

text_formatter.go
```diff
@@ -74,8 +77,19 @@ func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
 
 	prefixFieldClashes(entry.Data)
 
-	isColorTerminal := isTerminal && (runtime.GOOS != "windows")
-	isColored := (f.ForceColors || isColorTerminal) && !f.DisableColors
+	f.terminalLock.RLock()
```
stevvooe: I think you can use a `sync.Once` here to make this a bit less complicated:

```
f.terminalOnce.Do(func() {
  f.isTerminal = IsTerminal(entry.Logger.Out)
})
```

Once `f.isTerminal` is determined and guarded by this statement, it should be safe to read. The cost will be reduced to the fast case for `sync.Once`.

stevvooe:
@sirupsen This one is extremely tricky. The challenge with `Formatter` is that there is no hook location to synchronize the call by keeping it a single place. I've place a few comments on this technique, but it would be best to avoid a lock on every `Format` call (although, there are cases where I am worried we cannot avoid it).

The other approach would be to put the detection on the logger and propagating it down to the entry when messages are emitted. In certain cases, this can be almost lock free, but we gamble on someone setting the output to something different.

Babbling aside, the `sync.Once` approach is probably the most correct one without getting fancy.

sirupsen:
Oh man, I really haven't done enough Go lately. I've changed to use `sync.Once` a lot simpler. Thanks @stevvooe 

LGTY?

stevvooe:
LGTM

PR #473, opened 2017-02-06
Make LevelHooks goroutine safe.

yu81:
There is a concurrent write problem because LevelHooks has been alias of map.
I think it would be better to make LevelHooks struct with map and Mutex, and lock in Adding hooks.

### DIFF
--- a/hooks.go
+++ b/hooks.go
@@ -1,5 +1,7 @@
 package logrus
 
+import "sync"
+
 // A hook to be fired when logging on the logging levels returned from
 // `Levels()` on your implementation of the interface. Note that this is not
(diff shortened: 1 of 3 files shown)


kaneshin:
👀

yu81:
🙏

freeformz:
What is the use case for this? Generally speaking adding hooks is done somewhere in main and then generally not modified once the main parts of the program are started up.

freeformz:
Thank you, but in order to clean up older PRs and issues (this PR is > 2 years old), I'm going to close this PR. Feel free to re-open with an updated branch if this is still of interest. PS: I'm sorry this sat for so long.

Issue #408, opened 2016-08-31
error in one Hook.Fire() prevents subsequent hooks from firing


tochiai:
The loop in [LevelHooks.Fire()](https://github.com/Sirupsen/logrus/blob/master/hooks.go#L29) is interrupted when there is an error, which is undesirable in our use case. This is different behavior from the doc comment:

> // Fire all the hooks for the passed level.

Our workaround is to `LevelHooks.Add()` the hooks in order of increasing likelihood of an error so that the loop will not terminate prematurely.

We think it is preferable to allow all of the hooks in `LevelHooks` to attempt to fire. This can be accomplished by using [go-multierror](https://github.com/hashicorp/go-multierror) to capture errors as they arise, and then return the collective error after the loop has fired all the hooks for a log level. I'm happy to submit a PR if this seems like a desirable change.

Thanks!

sirupsen:
Great issue. I don't think we need a full import of `multierror`. Something simpler should be OK.

dgsb:
up

PR #474, opened 2017-02-07
hooks: don't skip entire chain if a hook fails

sirupsen:
It's pretty problematic that if one hook fails none of the others in the chain after it will run. We can do better. Since `LevelHooks.Fire()` is unfortunately publicly exported with its `error` signature, we can't return an `[]error` type. Ideally, `Hook.Fire()` shouldn't even allow returning an error, but there's no reason to break the 100s of hooks out there for that.

Should fix https://github.com/sirupsen/logrus/issues/408

@stevvooe

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -95,7 +95,7 @@ func (entry Entry) log(level Level, msg string) {
 
 	if err := entry.Logger.Hooks.Fire(level, &entry); err != nil {
 		entry.Logger.mu.Lock()
-		fmt.Fprintf(os.Stderr, "Failed to fire hook: %v\n", err)
+		fmt.Fprintf(os.Stderr, "Failed to fire hook(s): %v\n", err)
 		entry.Logger.mu.Unlock()
 	}
 	buffer = bufferPool.Get().(*bytes.Buffer)
--- a/hooks.go
+++ b/hooks.go
@@ -1,5 +1,9 @@
 package logrus
 
+import (
+	"fmt"
+)
+
 // A hook to be fired when logging on the logging levels returned from
 // `Levels()` on your implementation of the interface. Note that this is not
 // fired in a goroutine or a channel with workers, you should handle such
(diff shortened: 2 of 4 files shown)


hooks.go
```diff
@@ -24,11 +28,13 @@ func (hooks LevelHooks) Add(hook Hook) {
 // Fire all the hooks for the passed level. Used by `entry.log` to fire
 // appropriate hooks for a log entry.
 func (hooks LevelHooks) Fire(level Level, entry *Entry) error {
+	var aggErr error
+
 	for _, hook := range hooks[level] {
 		if err := hook.Fire(entry); err != nil {
-			return err
+			aggErr = fmt.Errorf("%s | %s", aggErr, err)
```
sirupsen: This is pretty ghetto, I'm not sure if there is an idiom out there to do this better.

entry.go
```diff
@@ -95,7 +95,7 @@ func (entry Entry) log(level Level, msg string) {
 
 	if err := entry.Logger.Hooks.Fire(level, &entry); err != nil {
 		entry.Logger.mu.Lock()
```
stevvooe: Maybe unrelated: what are these locks here for?

hooks.go
stevvooe: There aren't great solutions for this without getting fancy. You could fold it into a lightweight interface that does the formatting and collection management.

Where is this called? It might be reasonable just to log these failures in place.

entry.go
sirupsen: I believe the reason is that since the default for `logrus` is to log to `stderr` this could mix into the logging otherwise and cause corrupted, strange output.

hooks.go
stevvooe: @sirupsen A circuit breaker sounds fine. I've also used https://godoc.org/golang.org/x/time/rate successfully, but I am not sure if this warrants the dependency. You might be able to crib from [here](https://github.com/docker/go-events/blob/master/retry.go#L126) (don't use that as a dependency, as we may change the interface).

Another possibility would be to increment and export a counter through expvar or prometheus, but that might be more involved.

Issue #475, opened 2017-02-07
Undefined errors on windows server 2012


majimboo:
```
C:\gocode\src\github.com\Sirupsen\logrus\terminal_windows.go:22: undefined: io in io.Writer
C:\gocode\src\github.com\Sirupsen\logrus\terminal_windows.go:23: use of .(type) outside type switch
C:\gocode\src\github.com\Sirupsen\logrus\terminal_windows.go:24: undefined: os in os.File

```

majimboo:
It looks like `terminal_windows.go` didn't import both `io` and `os`. Should use the auto-importer when coding.

powareverb:
Same issue in Win10, your fix seems to work here also.

sirupsen:
Thanks for the quick fix. Merged.

PR #476, opened 2017-02-07
Fixed missing imports for windows terminal

majimboo:
As reported here https://github.com/sirupsen/logrus/issues/475


sirupsen:
Ugh, thanks. I don't have a Windows testing environment.

PR #481, opened 2017-02-08
Fixed import

darrenmcc:


jprobinson:
:bump:

Any chance this can get merged? This bug causes a compile time error for any app engine user.

xrstf:
The same change needs to happen in ``terminal_solaris.go``.

zeekay:
@sirupsen anything in particular holding up this PR? Logrus is breaking our build (and while we can manually apply this patch, it's annoying to deal with from the perspective of continuous integration).

xrstf:
Our build includes running ``goimports``, which "accidentally" also fixes files in ``vendor/`` and hence these missing imports. Maybe this would help you as well?

darrenmcc:
We've pinned to a previous working version with glide, but a _real_ fix is preferred.

jprobinson:
Almost two weeks and no word from any maintainers on this extremely simple fix? This is pretty absurd especially given that these are compile time errors that should have never made It to master.

@sirupsen please pull this in. I don't want to remove logrus from gizmo, but this is pretty bad practice of OSS and I don't want my other engineers to deal with this kind of behavior on the future.

sirupsen:
> Almost two weeks and no word from any maintainers on this extremely simple fix?

Would you like to help maintain Logrus? I'd be happy to make you a collaborator.

sirupsen:
Fixed in v0.11.3

Issue #478, opened 2017-02-07
Breaking API Change in IsTerminal()


bbrks:
Hi,

Just raising an issue to let you know #471 seemed to break the API without incrementing the major version number. I'm not asking for a revert or change, just logging an issue to help with people trying to debug if they encounter the same thing.

`logrus.IsTerminal()` used to take no parameters, now it takes an `io.Writer` interface.

Compilation error for searchability:
`not enough arguments in call to logrus.IsTerminal`

sirupsen:
Yeah, I did consider this—the trade-off here is to fix a bunch of TTY-related bugs for a function that I wouldn't expect people to use publicly. A TTY-package should be used for this. Regardless, I think it's a bug to say `logrus.IsTerminal()`. It's completely opaque what that means. Is the main logger's out a TTY? Is stderr a TTY?

When doing this work it's weighed between how many people breaking backwards compatibility would affect (in this case, my hunch is few) between how many people the bug is affecting. It's possible to make this change with complete backwards compatibility, with the cost of nasty cruft, but I didn't think this was going to affect many.

Thanks for logging this issue @bbrks

bbrks:
No worries @sirupsen, we only ran into the issue as we have a custom formatter for our logs, based on Logrus' own :) I did think it's likely not a widely used function.

sirupsen:
Out of curiosity (I have considered removing formatters in favour of hooks for years), what are you using your formatter for?

bbrks:
We only have a few changes to suit our particular use. In particular:

- Added `DefaultFields` to the `TextFormatter` struct and append them in the Format method. This allows us to share a global logger which has a set of default fields (such as the name of the service, which environment it is running in, etc.)
- Minor changes to the way fields are quoted to suit our logging infrastructure (quote empty fields, quote with single quotes rather than double)

sirupsen:
> Added DefaultFields to the TextFormatter struct and append them in the Format method. This allows us to share a global logger which has a set of default fields (such as the name of the service, which environment it is running in, etc.)

Is there any reason you don't pass a `logrus.Entry` around? If you do `logger.WithFields(..)` you can pass that around for an environment-specific logger. I end up doing this a lot.

> Minor changes to the way fields are quoted to suit our logging infrastructure (quote empty fields, quote with single quotes rather than double)

These actually seem like useful additions to the upstream `TextFormatter`

bbrks:
[...]

PR #484, opened 2017-02-14
Text formatter quote configuration

bbrks:
Hi @sirupsen,

As discussed the other day in #478, this PR will allow you to quote empty fields, and also provide a rune to override the default quote character.

Please let me know what you think, or highlight anything that you would like changing.

Thanks!

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode/utf8"
 )
 
 const (
@@ -52,9 +53,13 @@ type TextFormatter struct {
 	// QuoteEmptyFields will wrap empty fields in quotes if true
 	QuoteEmptyFields bool
 
+	// QuoteRune can be set to override the default quote style
+	QuoteRune rune
+
 	// Whether the logger's out is to a terminal
-	isTerminal   bool
-	terminalOnce sync.Once
+	isTerminal bool
+
+	sync.Once
 }
 
 func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


bbrks:
I'll take a look at fixing this PR for Go 1.6 tomorrow.

text_formatter.go
```diff
@@ -49,9 +50,16 @@ type TextFormatter struct {
 	// be desired.
 	DisableSorting bool
 
+	// QuoteEmptyFields will wrap empty fields in quotes if true
+	QuoteEmptyFields bool
+
+	// QuoteRune can be set to override the default quote style
+	QuoteRune rune
```
sirupsen: I think it'd be completely clear what this means if the description was:

"QuoteRune can be set to the override the default quoting `"` with e.g. `'`."

Should we consider making this a string instead of a rune to not limit people to a single character? That would be hard to fix without breaking backwards compatibility later.

text_formatter.go
```diff
@@ -72,7 +80,10 @@ func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
 
 	prefixFieldClashes(entry.Data)
 
-	f.terminalOnce.Do(func() {
+	f.Do(func() {
+		if f.QuoteRune == 0 || !utf8.ValidRune(f.QuoteRune) {
+			f.QuoteRune = '"'
+		}
```
sirupsen: Good idea to use this pattern. Since the function is larger now, do you think we could put this in a named function?

sirupsen:
Thanks @bbrks

bbrks:
Cool, thanks @sirupsen

PR #491, opened 2017-02-17
Remove os.Exit from entry.go

sargun:
This removes os.Exit from entry.go, because we already exit at
logger.go:166. Internal usage of entries in other libraries
ends up causing unexpected failures.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -161,7 +161,6 @@ func (entry *Entry) Fatal(args ...interface{}) {
 	if entry.Logger.Level >= FatalLevel {
 		entry.log(FatalLevel, fmt.Sprint(args...))
 	}
-	Exit(1)
 }
 
 func (entry *Entry) Panic(args ...interface{}) {


sargun:
*nudge*

sargun:
Right now, if you have multiple logging hooks, it causes them not to fire because os.exit is called before all hooks are triggered. I believe this is a bug. @sirupsen

sirupsen:
This is intended to exit at some point, though. When do you suggest we exit instead?

stevvooe:
It sounds like there needs to be a way to fire the hooks then exit.

DmitriyMV:
If hooks do not do any async operations or they do and wait for them, current behavior should work fine. However, if they do not, with current design we have no way to verify that all hooks have done their job. But you can register you own exit handler with `RegisterExitHandler`. Maybe this should do the trick?

sargun:
We still have an exit function that gets called here:

https://github.com/sirupsen/logrus/blob/master/logger.go#L230

DmitriyMV:
@sargun `WithFields` returns an `Entry` not `Logger`. Calling `Fatal` on `Entry` will not call it on `Logger`.

dgsb:
@sargun we can not merge this. It is a behaviour change which will eventually hurt other users. Did you try the RegisterExitHandler suggestion in order to synchronize with the completion of asynchronous hooks ?

PR #492, opened 2017-02-20
Allow timestamp for included formatters to be in UTC

mzuneska:
Hi @sirupsen and @stevvooe. This PR is closely related to https://github.com/sirupsen/logrus/pull/294 and https://github.com/sirupsen/logrus/pull/402.
I'd still like to put forward the idea of allowing formatted timestamps to be in UTC but without going as far as allowing specific timezones. I'd guess most people just want UTC all the time. As to any questions about why would the time on the server not be set to UTC, well it should be as far as I know but things happen and I'd rather specify in my app as well. As far as I can tell this doesn't break any backwards compatibility.

### DIFF
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -29,6 +29,8 @@ type JSONFormatter struct {
 	// DisableTimestamp allows disabling automatic timestamps in output
 	DisableTimestamp bool
 
+	TimestampInUTC bool
+
 	// FieldMap allows users to customize the names of keys for various fields.
 	// As an example:
 	// formatter := &JSONFormatter{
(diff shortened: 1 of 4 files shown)


stevvooe:
@mzuneska There was nothing wrong with #401. It looks like the contributor got fed up with finding a correct design for a package that affects a lot of users. If you want to carry #401, that would be great!

I'm not sure I like this approach. Adding yet another boolean just creates extra space for unclear parameterization. For example, what would be the behavior if we merged this PR and #401?

mzuneska:
@stevvooe Totally understandable about the additional boolean parameters. I just don't need anything other than UTC and that's been the few other comments I've seen so it seemed extra complication to add the time/location. I can close this PR if it doesn't look like it will be merged.

PR #505, opened 2017-03-15
Format full values for fields

patrick-east:
Previously we would just use the equivelent of a
%s but this wouldn't show structs that were passed
in as key values very nicely. If we change to process
them as %+v the output is much more helpful, and
saves from having to implement stringers for every
single struct that might be logged.


lledey:
Would love to see this merged!

freeformz:
Thank you, but in order to clean up older PRs and issues (this PR is > 2 years old), I'm going to close this PR. Feel free to re-open with an updated branch if this is still of interest. PS: I'm sorry this sat for so long.

PR #512, opened 2017-03-23
Fix SetLevel data-race

DmitriyMV:
From #374.

The commit fixes data race using atomics Load\Save. We switch type of level from uint8 to uint32 but due memory alignment on most platforms it will not result in any additional memory. This can actually be a breaking change if you store Level somewhere in your code. The alternative would be creating a new type and using it inside `Logger` struct. On package boundaries we could convert it from existing `Level` silently.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -126,7 +126,7 @@ func (entry Entry) log(level Level, msg string) {
 }
 
 func (entry *Entry) Debug(args ...interface{}) {
-	if entry.Logger.Level >= DebugLevel {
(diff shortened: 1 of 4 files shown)


sirupsen:
Why is there a race here if there's a lock in `setLevel`? I'm assuming because every log outputting doesn't actually grab the lock?

DmitriyMV:
Exactly - if we try to set a different level and concurrent goroutine will try to write a log - we will get datarace.

sirupsen:
Thanks for the fix!

PR #513, opened 2017-03-23
Fix capitalization in README.md first example

maxthomas:
seems the example import should be lower-case (?)


maxthomas:
see #514 - more complete PR

PR #516, opened 2017-03-28
Re-incorporating the trace logging originally written by Sean Berry.

rogerclermont:
Sean Berry incorporated trace-level logging into the logrus library we're using.  Incorporating these changes into a fork so they're not lost.


rogerclermont:
Apologies - this wasn't meant for this repo.

PR #517, opened 2017-03-29
Update readme to point to same case org

zbintliff:
Your actual org and remote has a lowercase `s` readme references uppercase `S`.  Believe it or not this breaks dep tools like `glide` in edgecases. While fixing those tools are better I think we should import using the actual org and repo name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ The simplest way to use Logrus is simply the package-level exported logger:
 package main
 
 import (
-  log "github.com/Sirupsen/logrus"
+  log "github.com/sirupsen/logrus"
 )
 
 func main() {
@@ -80,7 +80,7 @@ package main
 
 import (
   "os"
-  log "github.com/Sirupsen/logrus"
+  log "github.com/sirupsen/logrus"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


harnash:
This is also problem for code in `github.com/sirupsen/logrus/hooks/syslog`

polds:
Any chance we could get this merged? It's still leading to some confusion.

harnash:
This was reviewed sometime before and reverte. For more info go to #384

zbintliff:
Right but i'm not sure what the issue is.  Capital `S` is wrong.  The actual pathing is lowercase on github.  Tools like `glide` that validate that the remote matches the import break on updates.

harnash:
It may be wrong but changing it now disrupts builds too especially when you are pulling some libraries you have no control over. See first lines in README:

> Seeing weird case-sensitive problems? See this issue. This change has been reverted. I apologize for causing this. I greatly underestimated the impact this would have. Logrus strives for stability and backwards compatibility and failed to provide that.

Well it seems I need to change it back in our codebase.

sirupsen:
Glide will work if you use lower-case, and this repo supports both upper-case (for backwards compatibility) and lower-case.

PR #526, opened 2017-04-13
Update logrus import path

gravis:
See gemnasium/logrus-graylog-hook/issues/23


gravis:
Looks like it's a duplicate of #384, which is merged, so I suspect the change to have been reverted at some point. /me investigating...

PR #530, opened 2017-04-23
add caller offset to fix caller's problem.

bobcyw:
After read the zap, I think add caller offset to entry is a good idea to solve #63.

Then we can use a hook like this 

```

import (
	"path/filepath"
	"runtime"
	"strconv"
	"strings"
	
	"github.com/bobcyw/logrus"
)

type CallerHook struct {
}

func (hook *CallerHook) Fire(entry *logrus.Entry) error {
	entry.Data["caller"] = hook.caller(entry)
	return nil
}

func (hook *CallerHook) Levels() []logrus.Level {
	return []logrus.Level{
		logrus.PanicLevel,
		logrus.FatalLevel,
		logrus.ErrorLevel,
		logrus.WarnLevel,
		logrus.InfoLevel,
		logrus.DebugLevel,
	}
}

func (hook *CallerHook) caller(entry *logrus.Entry) string {
	if _, file, line, ok := runtime.Caller(entry.CallerOffset); ok {
		return strings.Join([]string{filepath.Base(file), strconv.Itoa(line)}, ":")
	}
	// not sure what the convention should be here
	return ""
}
```

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -42,6 +42,10 @@ type Entry struct {
 
 	// When formatter is called in entry.log(), an Buffer may be set to entry
 	Buffer *bytes.Buffer
+	// Record the caller's depth
+	CallerOffset int
+	// Init caller's depth
+	InitCallerOffset int
 }
 
 func NewEntry(logger *Logger) *Entry {
@@ -49,6 +53,8 @@ func NewEntry(logger *Logger) *Entry {
 		Logger: logger,
 		// Default is three fields, give a little extra room
 		Data: make(Fields, 5),
+		CallerOffset: 6,
+		InitCallerOffset: 6,
 	}
 }
 
@@ -82,7 +88,7 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 	for k, v := range fields {
 		data[k] = v
 	}
-	return &Entry{Logger: entry.Logger, Data: data}
+	return &Entry{Logger: entry.Logger, Data: data, CallerOffset:4, InitCallerOffset:4}
 }
 
 // This function is not declared with a pointer value because otherwise
@@ -98,6 +104,7 @@ func (entry Entry) log(level Level, msg string) {
 		fmt.Fprintf(os.Stderr, "Failed to fire hook: %v\n", err)
 		entry.Logger.mu.Unlock()
 	}
+	
 	buffer = bufferPool.Get().(*bytes.Buffer)
 	buffer.Reset()
 	defer bufferPool.Put(buffer)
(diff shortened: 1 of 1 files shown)


sirupsen:
I like this idea a lot more than adding caller information to every entry, which is going to be too much data for those not using this feature. This is the only way to get the correct level in the hook? It cannot be hard-coded?

bobcyw:
I think there is some other trick to solve this problem. Such like keep all log function with same stack depth. Now when we invoke Println, the calling path is Println->Infoln->Info->log(stack depth is 3), but when we invoke Infoln, it's caller path is Infoln->Info->log(stack depth is 2). It always change. If we could keep their stack depth a fix number, we can solve this problem without CallerOffset.

sirupsen:
If you update this PR to just go directly to `Log` in all functions, perhaps we don't need to know anything about callers to implement a hook that adds the callee?

bobcyw:
It's right. I'll update it this weekend.

bobcyw:
I have complete it. But I find another problem some framework insert their own function object between the real caller and entry.log. Look at below stack info:
callernormal
This is a normal scenario.
callerwithcustomlayer
This is a special scenario. I found such like [socket-io](https://github.com/googollee/go-socket.io) will insert their own layer.

So the caller hook should be well defined for special scenario.

bobcyw:
I think I find out the reason. If we use ```Logger``` instance, the call stack is 
```Logger.Println -> Entry.Println -> Entry.log```. If we use ```Entry``` instance which is created by ```Logger.WithField```, the call stack is ```Entry.Println -> Entry.log```. So I update my PR to solve this problem. Use ```Logger.Println -> Entry.log instead``` of ```Logger.Println -> Entry.Println -> Entry.log```.

**Comment by bobcyw** [comment]
I write a hook like this. Magic number 5 work well both with Logger and Entry.
```go
func NewNormalCaller() logrus.Hook {
	return &callerHook{}
}

type callerHook struct {
}

func (hook *callerHook) Fire(entry *logrus.Entry) error {
	entry.Data["caller"] = hook.caller(entry)
	return nil
}

func (hook *callerHook) Levels() []logrus.Level {
	return []logrus.Level{
		logrus.PanicLevel,
		logrus.FatalLevel,
		logrus.ErrorLevel,
		logrus.WarnLevel,
		logrus.InfoLevel,
		logrus.DebugLevel,
	}
}

func (hook *callerHook) caller(entry *logrus.Entry) string {
	if _, file, line, ok := runtime.Caller(5); ok {
		return strings.Join([]string{filepath.Base(file), strconv.Itoa(line)}, ":")
	} else {
		return ""
	}
}

```

bobcyw:
Hi, @sirupsen , I have rebased this commit on the newest master branch. And solved the unstable caller problem. The solution is modifying not only entry's function but also public log's function.

PR #532, opened 2017-05-02
Option to report caller info

magic53:
Reference #63 Extends the functionality of logrus by providing the option to report caller info:
```
logr := &logrus.Logger{
	Out:       	os.Stdout,
	Formatter: 	new(logrus.TextFormatter),
	Hooks:     	make(logrus.LevelHooks),
	Level:		logrus.DebugLevel,
	ReportCallerOptions: &logrus.ReportCallerOptions{
		Field: "file", // default is "src"
		Flags: log.Lshortfile, // default is "log.Llongfile"
	},
}
logr.Println("Test log")
// time="2017-05-02T10:26:49-05:00" level=info msg="Test log" file="main.go:12" 
```

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -1,9 +1,13 @@
 package logrus
 
 import (
+	"fmt"
 	"io"
+	"log"
 	"os"
+	"path"
 	"sync"
+	"runtime"
 )
 
 type Logger struct {
@@ -30,6 +34,8 @@ type Logger struct {
 	mu MutexWrap
 	// Reusable empty entry
 	entryPool sync.Pool
+	// Report caller options
+	ReportCallerOptions *ReportCallerOptions
 }
 
 type MutexWrap struct {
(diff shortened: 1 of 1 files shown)


sirupsen:
To get this into core, it needs to be implemented as a hook. What will you need from core to implement this as a hook?

magic53:
In short, core needs to do the heavy lifting in determining what the caller is at the earliest possible point in the stack, in order to ensure stability of reporting caller info in hooks. This is because we have to determine the exact number of stack frames to skip `runtime.Caller(skipFrames)` [std logger actually does this](https://golang.org/src/log/log.go?s=10393:10435#L336) 

You can then leave it up to devs to determine how to decorate this info in the logger entries through hooks. In the PR you can see that skipping frames 2 is hardcoded, this is acceptable because core knows how many frames it needs to skip, whereas hooks do not (they have to guess). Std logger also hardcodes the number frames to skip.

*Proposal*
- Core determines caller info and safely passes this information to logger Entry
 - Hooks can safely access logger Entry without making assumptions about stack frames
- Core stores caller info in *Entry*
- Hooks shouldn't be allowed to determine the caller info
 - If core changes placement of one nested function call it would break those implementations
 - Other work arounds include dynamically traversing the stack to guess caller (maintenance nightmare and `runtime.Caller` is expensive)

magic53:
@sirupsen I will submit another PR that meets the requirements for the Proposal above, as well as your requirements that Hooks should be responsible for decorating caller info.

magic53:
Ok, see https://github.com/sirupsen/logrus/pull/544 let's continue conversation there

timbunce:
If #544 supersedes this one, can this one be closed? (And thanks for working on this @magic53.)

magic53:
@timbunce yes closing this in favor of keeping discussion at #544

Issue #529, opened 2017-04-20
Data race in log()


flimzy:
I discovered a data race in the [log()](https://github.com/sirupsen/logrus/blob/master/entry.go#L90) function.  Specifically, in [this code](https://github.com/sirupsen/logrus/blob/master/entry.go#L96-L106):

     96    if err := entry.Logger.Hooks.Fire(level, &entry); err != nil {
     97		entry.Logger.mu.Lock()
     98		fmt.Fprintf(os.Stderr, "Failed to fire hook: %v\n", err)
     99		entry.Logger.mu.Unlock()
    100	}
    101	buffer = bufferPool.Get().(*bytes.Buffer)
    102	buffer.Reset()
    103	defer bufferPool.Put(buffer)
    104	entry.Buffer = buffer
    105	serialized, err := entry.Logger.Formatter.Format(&entry)
    106	entry.Buffer = nil

Line 96 sends `entry` away to the logging backend via `Fire()`, but then proceeds to modify `entry` on lines 102 and 104.

This causes a data race if the backend attempts to read `entry`--and it most certainly will. :)

To resolve this without changing the API, I suggest sending a separate copy of `entry` to `Fire()`, so that the local copy can be modified at will without fear of racing.

amits1995:
Had panics in production because of this :(
I manually copied the entry in my Fire method to resolve that

DmitriyMV:
Hooks can modify the passed entry, so this will change the current behavior. Maybe we can change the current behavior design? Hooks shouldn't be able to change the passed data IMHO.

xor-gate:
I was also bitten by this one. I had a unit test with multiple goroutines and the `hooks/test` made a copy of the pointer. Then getting the last entry results in a data race because after fire the `entry.Buffer` is modified.

code

```go
package redis

import (
        "strings"
        "testing"
        "time"

        "github.com/mna/redisc/redistest"
        logrus_test "github.com/sirupsen/logrus/hooks/test"
        "github.com/stretchr/testify/assert"
        "github.com/stretchr/testify/require"
)

func TestRedisPubSubReconnect(t *testing.T) {
        lt := logrus_test.NewGlobal()

        rcClose, rcPorts := redistest.StartCluster(t, nil)
        defer rcClose()

        r, err := New([]string{":" + rcPorts[0]})
        require.Nil(t, err)
        defer r.Close()

        time.Sleep(10 * time.Second)
[...]

PR #536, opened 2017-05-05
Entry.WithFields(): copy the original entry

Julio-Guerra:
`Entry.WithFields()` currently creates a new `Entry` and partially copies it. Calling several times `WithFields()` on an already logged entry (e.g. in a formatter) thus leads to an incorrectly initialized `Entry` structure.

The patch makes the method copy the original entry and only modify the `Data` member with the new set of fields. So that calling several times `WithFields()` on an already logged entry
(e.g. in a formatter `Format()` method) avoids this problem.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -82,7 +82,10 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 	for k, v := range fields {
 		data[k] = v
 	}
-	return &Entry{Logger: entry.Logger, Data: data}
+	// shallow copy
+	newentry := *entry
+	newentry.Data = data
+	return &newentry
 }
 
 // This function is not declared with a pointer value because otherwise
--- a/entry_test.go
+++ b/entry_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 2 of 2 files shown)


entry.go
```diff
@@ -82,7 +82,10 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 	for k, v := range fields {
 		data[k] = v
 	}
-	return &Entry{Logger: entry.Logger, Data: data}
+	// shallow copy
+	newentry := *entry
+	newentry.Data = data
+	return &newentry
```
stevvooe: Is it safe to carry over the buffer? I think this was restricted to a particular set of fields to avoid that problem.

Julio-Guerra: It *seems* to be safe: so far the implementation of `Entry.log()` overwrites `Entry.Buffer` [every time](https://github.com/sirupsen/logrus/blob/master/entry.go#L104) and the buffer pool [is protected](https://github.com/sirupsen/logrus/blob/master/entry.go#L11) with `sync.Pool`.

stevvooe: @sirupsen Any history on this one?

sirupsen: Unsure, I'd be most concerned about performance with this?

DmitriyMV: The unused fields are zero initialized anyway, so this shouldn't result in any significant performance penalties. Benchmark it maybe?

DmitriyMV:
We should probably fix #529 first because merging this will cause race detector to go nuts.

entry.go
stevvooe: I don't think the extra fields are going to cause a large performance issue compared with the cost of the allocation. What I am worried about is the errant copying causing a future race. The shared buffer also has cause for concern. It might be better/safer to move to a buffer pool.

PR #537, opened 2017-05-08
[Fix bug] entry.Data would be inherited when entry.Data was modified by hook

git-hulk:
This issue can be reproduced easily:

```golang
type TestHook struct {
}

func (th *TestHook) Fire(entry *logrus.Entry) error {
    entry.Data["test"] = "test"
    return nil 
}

func (sh *TestHook) Levels() []logrus.Level {
    return []logrus.Level{
        logrus.WarnLevel,
    }   
}

func main() {
    logger := logrus.New()
    logger.Hooks.Add(new(TestHook))
    logger.Warn("log warn message")
    logger.Info("log info message")
}
```
and the result is:

```
WARN[0000] log warn message                           test=test
INFO[0000] log info message                               test=test
```

`test=test` shouldn't be printed while the hook was only added to `WarnLevel`


git-hulk:
@dmathieu Can you review this PR, thanks~

dmathieu:
Could you add a test case?

git-hulk:
@dmathieu Done, thanks for the review

git-hulk:
@dmathieu Can you merge this pr? And it may cause conflicts.

git-hulk:
@dmathieu  Do you know if there's anything that would stop this being merged?

PR #540, opened 2017-05-11
fix readme example go import path

JodeZer:
Hi, l found the example not work for me cause the import path is not right.
Fix this may help some new.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ The simplest way to use Logrus is simply the package-level exported logger:
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


AlexRudd:
There are also various import paths dotted around the project referencing `github.com/Sirupsen/logrus`
Should correcting these be part of this request or possibly a separate one?

sirupsen:
Yep, I think this is the right thing to do going forward—you can use both now, but I can't break backward compatibility inside the build. I've committed a `test_lowercase` for the hook that referenced the upper-case path.

Issue #496, opened 2017-02-27
Enable ANSI colors for Windows 10 and newer


hacdias:
Hey!

I've been reading a bit about Windows CMD and Virtual Processing. This code bellow is enough to make ANSI work on Windows 10 and newer:

```go
handle := syscall.Handle(os.Stdout.Fd())
kernel32DLL := syscall.NewLazyDLL("kernel32.dll")
setConsoleModeProc := kernel32DLL.NewProc("SetConsoleMode")
setConsoleModeProc.Call(uintptr(handle), 0x0001|0x0002|0x0004)
```

More information about it here: https://msdn.microsoft.com/en-us/library/windows/desktop/ms686033(v=vs.85).aspx

It would be interesting to add!

hacdias:
Now I got really confused. I changed from "0x0004" to "7" and it works now. I saw it on a Python example online and it works. But why?

ezk84:
Indeed it would be nice to be able to see coloured output rather than the following in Windows 10:

    ?[34mINFO?[0m[0000] A walrus appears                              ?[34manimal?[0m=walrus

hacdias:
The addition of that code would solve that. What do you think @sirupsen?

hacdias:
Done @ezk84 😄 Let's see what happens.

ezk84:
I've added a PR for your PR, where I make a few alterations to make it more likely to be accepted into @sirupsen's repo.

hacdias:
Done! Thank you 😄

dmathieu:
It looks like this issue has been resolved. Closing.
Please let me know if this isn't the case.

omar-h:
Continues to be an issue for me, this is what my console looks like:
```
?[36mINFO?[0m[0000] Loaded module: Help
?[37mDEBU?[0m[0002] hi
?[37mDEBU?[0m[0003] Error sending message.                        ?[37merror?[0m="HTTP 400 Bad Request, {\"embed\": [\"url\"]}"
```

Using Ubuntu on Windows (in cmd) it seems to work, just not in a standard cmd window.

hacdias:
Make sure this isn't checked:

![image](https://user-images.githubusercontent.com/5447088/30155037-b2a111ec-93b3-11e7-9424-1ba78bb2507c.png)

omar-h:
It isn't, still doesn't work.
![cmd Properties](https://i.imgur.com/OwQPMnZ.png)

hacdias:
[...]

PR #541, opened 2017-05-11
Enable ANSI colors for Windows 10 and newer

hacdias:
See issue #496.

### DIFF
--- a/terminal_windows.go
+++ b/terminal_windows.go
@@ -41,7 +41,8 @@ func getVersion() (float64, error) {
 	if err != nil {
 		return -1, err
 	}
-
+	
+	// The output should be like "Microsoft Windows [Version XX.X.XXXXXX]"
 	version := strings.Replace(stdout.String(), "\n", "", -1)
 	version = strings.Replace(version, "\r\n", "", -1)
 
@@ -57,7 +58,7 @@ func getVersion() (float64, error) {
 func init() {
 	ver, err := getVersion()
 	if err != nil {
-		panic(err)
+		return
 	}
 
 	// Activate Virtual Processing for Windows CMD
--- a/terminal_windows.go
+++ b/terminal_windows.go
@@ -8,8 +8,13 @@
 package logrus
 
 import (
+	"bytes"
+	"errors"
 	"io"
 	"os"
+	"os/exec"
+	"strconv"
+	"strings"
 	"syscall"
 	"unsafe"
 )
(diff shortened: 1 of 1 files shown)


sirupsen:
Have you tested this on all relevant versions of Windows? We don't have CI for this, so things go a bit havoc if it's merged and doesn't work across platforms.

hacdias:
It only works for Windows 10 and newer because this functionality was announce for Win10.

hacdias:
Just had a lot of troubles squashing. It was my first time 😆 and I made some mistakes. Anyway, I added a version check to ensure that the feature is only enabled on Windows 10 and newer. Although I will try to check if running the code on older versions harms or not.

sirupsen:
This is really the simplest way to check Windows version? 😬

sirupsen:
@rnapier you wrote this, can you review? I have zero experience with Windows 😅

hacdias:
@sirupsen I wish I could have found a better way. There is a syscall for that but it was deprecated on Windows 8.1 and now, unless you direct your app exclusively for Windows 10 on the manifest data, it will always say you are on Windows 8. I tried that. It would be like... 5 lines?

sirupsen:
Ok, let's wait for @rnapier to give this a spin.

Have you tested this on a Windows machine on versions less than 10?

terminal_windows.go
```diff
@@ -18,8 +23,51 @@ var kernel32 = syscall.NewLazyDLL("kernel32.dll")
...
+	cmd.Stderr = stderr
+	err := cmd.Run()
+	if err != nil {
+		return -1, err
+	}
+
+	version := strings.Replace(stdout.String(), "\n", "", -1)
+	version = strings.Replace(version, "\r\n", "", -1)
+
+	x1 := strings.Index(version, "[Version")
```
sirupsen: Can you add a comment with the output?

terminal_windows.go
```diff
@@ -18,8 +23,51 @@ var kernel32 = syscall.NewLazyDLL("kernel32.dll")
...
+		return -1, errors.New("Can't determine Windows version")
+	}
+
+	return strconv.ParseFloat(version[x1+9:x1+13], 64)
+}
+
+func init() {
+	ver, err := getVersion()
+	if err != nil {
+		panic(err)
```
sirupsen: I don't think we need to panic if we can't find the version, let's just below not do anything.

terminal_windows.go
hacdias: Done 😄

terminal_windows.go
hacdias: Sure! 😄

hacdias:
> Have you tested this on a Windows machine on versions less than 10?

I haven't, but since the code doesn't run if you're not on Windows 10 or newer, I don't think there is any problem.

Kushcabbage:
I know this is a year old but, is there a way to enable ANSI color in a .bat script for a java console program?

PR #546, opened 2017-05-19
Sets first letter of all sirupsen imports lower-case

stweiz:
This pull request sets all import paths and other references of "sirupsen" to lower case letters, so everything should be unified.


sirupsen:
This is going to break Logrus from too many people yet. We'll have to create a new syslog package, preferably out of repo, that imports lower-case.

schultz9999:
We have a patch to go thru sources and do the manual patch once we import to vendor dir :-/

schultz9999:
If it's a breaking change, why not to release a major version?

sirupsen:
This was done in master with a bump of a major version.

PR #547, opened 2017-05-19
Update due to rename of github account

bobtfish:
This should fix the syslog hook which is currently broken


claytononeill:
I have the same problem and was about to submit a PR for the same fix.  The syslog hook appears to be broken for me because of this.

PR #550, opened 2017-05-30
switch terminal_linux to x/sys/unix from syscall

clnperez:
Signed-off-by: Christy Perez <christy@linux.vnet.ibm.com>

Closes sirupsen/logrus/issues/515

### DIFF
--- a/terminal_linux.go
+++ b/terminal_linux.go
@@ -7,8 +7,8 @@
 
 package logrus
 
-import "syscall"
+import "golang.org/x/sys/unix"
 
-const ioctlReadTermios = syscall.TCGETS
+const ioctlReadTermios = unix.TCGETS
 
 type Termios syscall.Termios


clnperez:
Looking at the vendoring issues.

clnperez:
Oh ha! There is no vendoring for this project (unless I'm missing it). Would that be something you'd be open to adding?

dmathieu:
Thank you @clnperez. This looks awesome.
I've opened #581 and will merge it tomorrow unless there are concerns.

Once this is done, if you rebase this PR and the tests pass, I'll be able to merge it.

clnperez:
Well, looks like that was shot down @dmathieu. 😆 Maybe I can just hardcode the right value in instead? Any other ideas?

dmathieu:
Yes, sorry about that. You will need to add a `go get` in `.travis.yml`, right after getting testify.
Thanks

clnperez:
@dmathieu that would only be helpful for this travis env though wouldn't it? How would projects using logrus be able to make sure to have the right version of x/sys?

dmathieu:
Ping @seiffert

dmathieu:
I think we don't have much of a choice. We need to rely on the latest version of x/sys/unix and update if their API introduces breaking changes.

seiffert:
Yes, I agree. That's a bit painful for libraries, but for all the applications that include logrus, it's much easier if logrus is compatible with the latest version of its dependencies! 👍

PR #563, opened 2017-06-18
Allow more chars in unquoted text formatter output

meatballhat:
I was surprised recently to find that `_` is considered a quotable character, which differs from other l2met implementations I tend to use, which has sometimes meant that I need to perform searches that look like:

```
foo=bar_baz OR foo="bar_baz"
```
It's a minor inconvenience, so I'm happy to accept a _wontmerge_ response 😃 ❤️


dmathieu:
I was actually bitten by this a couple weeks ago too, and we ended up changing the underscore to a dash.

PR #565, opened 2017-06-30
Add color setting support to text formatter

msiebeneicher:
This PR add a simple color change feature to the text formatter. All current default behaviours stays as they are. How to set up a different color map added to the `examples/basic/basic.go`.

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -9,23 +9,44 @@ import (
 	"time"
 )
 
+// NoColor: reset; clears all colors and styles (to white on black)
+// Red: set foreground color to red
+// Green: set foreground color to green
+// Yellow: set foreground color to yellow
+// Blue: set foreground color to blue
+// Magenta: set foreground color to magenta
+// Cyan: set foreground color to cyan
+// Gray: set foreground color to gray
+// Grey: set foreground color to gray (UK)
 const (
-	nocolor = 0
-	red     = 31
-	green   = 32
-	yellow  = 33
-	blue    = 34
-	gray    = 37
+	NoColor = uint8(0)
+	Red     = uint8(31)
+	Green   = uint8(32)
+	Yellow  = uint8(33)
+	Blue    = uint8(34)
+	Magenta = uint8(35)
+	Cyan    = uint8(36)
+	Gray    = uint8(37)
+	Grey    = Gray
 )
 
 var (
-	baseTimestamp time.Time
+	baseTimestamp   time.Time
+	defaultColorMap = ColorMap{
+		DebugLevel: Gray,
+		InfoLevel:  Blue,
+		WarnLevel:  Yellow,
+		ErrorLevel: Red,
+		FatalLevel: Red,
+		PanicLevel: Red,
+	}
 )
 
 func init() {
 	baseTimestamp = time.Now()
 }
 
+// TextFormatter for text and terminal output
 type TextFormatter struct {
 	// Set to true to bypass checking for a TTY before outputting colors.
 	ForceColors bool
@@ -56,21 +77,31 @@ type TextFormatter struct {
 	// with something else. For example: ', or `.
 	QuoteCharacter string
 
+	// ColorMap can be set to override the default color map
+	ColorMap ColorMap
+
 	// Whether the logger's out is to a terminal
 	isTerminal bool
 
 	sync.Once
 }
 
+// ColorMap maps levels to colors
+type ColorMap map[Level]uint8
+
 func (f *TextFormatter) init(entry *Entry) {
 	if len(f.QuoteCharacter) == 0 {
 		f.QuoteCharacter = "\""
+	}
+	if f.ColorMap == nil {
+		f.ColorMap = defaultColorMap
 	}
 	if entry.Logger != nil {
 		f.isTerminal = IsTerminal(entry.Logger.Out)
 	}
 }
 
+// Format execute the formatter
 func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
 	var b *bytes.Buffer
 	keys := make([]string, 0, len(entry.Data))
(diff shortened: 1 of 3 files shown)


text_formatter.go
```diff
@@ -9,23 +9,44 @@ import (
...
 
+// NoColor: reset; clears all colors and styles (to white on black)
+// Red: set foreground color to red
+// Green: set foreground color to green
+// Yellow: set foreground color to yellow
+// Blue: set foreground color to blue
+// Magenta: set foreground color to magenta
+// Cyan: set foreground color to cyan
+// Gray: set foreground color to gray
+// Grey: set foreground color to gray (UK)
```
dmathieu: Those comments are really superfluous. They basically say "green: this is green"

msiebeneicher:
> I like the idea, but I think the implementation could be finetuned. I'm a bit bothered that you're making the colors public.

The color constants should be public because you are using them from your own main application.

> How about keeping them private and having a SetLevelColor method:

> Having to worry about a full color map is weird too. If I were to change the level colors, I would probably be changing a single one, where you require changing them all (possibly to the same) or nothing.

The SetLevelColor() method is a good idea - cleaner and easier for the users. Still i would keep the colors public to be able to use them.

text_formatter.go
msiebeneicher: true but otherwise the linter will complain if public constants have no description

dmathieu: Golint accepts a single comment for this kind of constants.

dmathieu:
Right now, we're only holding the list of colors we need as you can't change them.
But as we start implementing that, there might be requests of handling 256 colors range. Then, the public constants become a huge mess.
Furthermore, logrus is a logging library, not a terminal configuration one. Holding a full list of all bash color codes really seems superfluous.

I'd say that if you're trying to change those colors, you need to know about bash colors codes (we can put a link in the documentation about them) and can handle integers.

msiebeneicher:
I reintegrated the master and added a `func (f *TextFormatter) SetLevelColor(l Level, c int)` method with example.

I still let the color constants public because in my opinion it is more conspicuous to have a small set of defaults for  inexperienced users in place. The color constants are only the basic ones and I see no reason why the should grow in this library. 

But if this should be a "no go" i can change them back to private.

dmathieu:
I'm sorry, but the more I think about it, the more I really don't think the colors should be public.
It's very easy to do so, but making them private later would be quite harder (as well as renaming them, if we change the default color for example).

Also, why do we still need to be able to set a full ColorMap?
The `SetLevelColor` method should be the only needed API.

Small changes are much better and more likely to be merged, as they are seen as less risky and much easier to maintain in the future.

msiebeneicher:
Any further comments for this PR?

dmathieu:
Sorry for the delay.
I'm trying to clear issues and PRs, and I'm still on the straightforward ones. So anything which isn't has less priority.

text_formatter.go
```diff
@@ -52,18 +66,35 @@ type TextFormatter struct {
 	// QuoteEmptyFields will wrap empty fields in quotes if true
 	QuoteEmptyFields bool
 
+	// QuoteCharacter can be set to the override the default quoting character "
+	// with something else. For example: ', or `.
+	QuoteCharacter string
```
dmathieu: This is a rebase issue. That option has been removed.

text_formatter.go
```diff
@@ -52,18 +66,35 @@ type TextFormatter struct {
...
 
 	sync.Once
 }
 
+// ColorMap maps levels to colors
+type colorMap map[Level]uint8
+
 func (f *TextFormatter) init(entry *Entry) {
+	if len(f.QuoteCharacter) == 0 {
+		f.QuoteCharacter = "\""
```
dmathieu: This has been removed and shouldn't be readded in this PR.

text_formatter.go
```diff
@@ -109,19 +140,16 @@ func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
...
-		levelColor = gray
-	case WarnLevel:
-		levelColor = yellow
-	case ErrorLevel, FatalLevel, PanicLevel:
-		levelColor = red
-	default:
-		levelColor = blue
+// SetLevelColor sets a color to a log level in the ColorMap
+func (f *TextFormatter) SetLevelColor(l Level, c uint8) {
+	if f.colorMap == nil {
```
dmathieu: When will colorMap be nil here? It should always be set during `init`. So this condition should be removable.

text_formatter.go
```diff
@@ -9,23 +9,37 @@ import (
...
+	blue    = uint8(34)
+	magenta = uint8(35)
+	cyan    = uint8(36)
+	gray    = uint8(37)
 )
 
 var (
-	baseTimestamp time.Time
+	baseTimestamp   time.Time
+	defaultColorMap = colorMap{
```
dmathieu: I like this.

text_formatter.go
```diff
@@ -138,6 +166,13 @@ func (f *TextFormatter) printColored(b *bytes.Buffer, entry *Entry, keys []strin
 	}
 }
 
+func (f *TextFormatter) getLevelColor(l Level) uint8 {
+	if val, ok := f.colorMap[l]; ok {
+		return val
+	}
+	return gray
```
dmathieu: This remains unconfigurable. How about having a `DefaultColor` in `colorMap`?

text_formatter.go
msiebeneicher: The `init(entry *Entry)` call is currently implemented in the `Format(entry *Entry)` method. If you want to set a new color **before** a log entry call you need to set the color map. 
You can check [here](https://github.com/msiebeneicher/logrus/blob/feature/add_color_setting_support_in_text_formatter/example_basic_test.go) the updated example.

Alternative you can provide a func NewTextFormatter() function.

text_formatter.go
msiebeneicher: "DefaultColor" as part of the color map is because of the level to color mapping (type colorMap map[Level]uint8) currently not good in my opinion. I would let the default color setting like it is for now and only implement a "DefaultColor handling" later if it is really necessary.

PR #572, opened 2017-07-07
add shortcut for caller info

sidbusy:


dmathieu:
Thank you for this PR. Could you add a test?

sidbusy:
@dmathieu Test added :-)

entry.go
```diff
@@ -63,6 +67,21 @@ func (entry *Entry) String() (string, error) {
...
 }
 
+// Add caller info as single field (using the key defined in CallerKey) to the Entry.
+func (entry *Entry) WithCaller(skip int) *Entry {
+	var (
+		file string
+		line int
+		ok   bool
+	)
+	_, file, line, ok = runtime.Caller(skip)
```
dmathieu: No need to explicitely set those vars here, as they get a value right below.

```
_, file, line, ok := runtime.Caller(skip)
```

entry_test.go
```diff
@@ -3,11 +3,43 @@ package logrus
...
+	logger.Out = &bytes.Buffer{}
+	entry := NewEntry(logger)
+
+	var (
+		file string
+		line int
+		ok   bool
+	)
+	calldepth := 1
+	_, file, line, ok = runtime.Caller(calldepth)
```
dmathieu: Same here for setting the vars.

entry_test.go
```diff
@@ -3,11 +3,43 @@ package logrus
...
+	if !ok {
```
dmathieu: When should this not be ok? Shouldn't it make the test fail if it happened?

dmathieu:
@sirupsen what do you think of this?

PR #573, opened 2017-07-07
Safely format data when printing

### LINKED ISSUE #531: Unsafe escaping of log data

```
	case string:
		if !f.needsQuoting(value) {
			b.WriteString(value)
		} else {
		}
```

```
			fmt.Fprintf(b, "%q", value)
```

kpcyrd:
Fixes #531

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -174,14 +174,14 @@ func (f *TextFormatter) appendValue(b *bytes.Buffer, value interface{}) {
 		if !f.needsQuoting(value) {
 			b.WriteString(value)
 		} else {
-			fmt.Fprintf(b, "%s%v%s", f.QuoteCharacter, value, f.QuoteCharacter)
+			fmt.Fprintf(b, "%q", value)
 		}
 	case error:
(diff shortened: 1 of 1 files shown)


dmathieu:
To follow up on your comment in the other issue, it doesn't seem to me like that option has any actual use.
But it should then be entirely removed, not be kept around broken ☺️

seiffert:
I agree with @dmathieu. The idea of having such an option is nice, however I also think that the use cases are limited. 
The option should either be removed from the library or be used. If you agree to use the option, `strconv` has a function `quoteWith` that basically does the same thing.

kpcyrd:
I'd prefer if somebody else could take over this PR, I'm not actually a golang developer, I've just noticed logrus logs insecurely.

seiffert:
Hey @kpcyrd, I'm happy to take this one over from you. I've copied your commit into #575 and will implement quoting with `QuoteCharacter` in there. (you can close this PR if you want)

kpcyrd:
@seiffert thanks!

Issue #503, opened 2017-03-14
Logger.SetLevel


ChimeraCoder:
It appears that we can't (safely) configure the loglevels of loggers *except* for the standard logger. 

For the standard logger, we can use the `logrus.SetLevel` method, which protects the level with a mutex, but this mutex is unexported. So if we set `var log = logrus.New()` in our application (which is [recommended in the docs](https://godoc.org/github.com/sirupsen/logrus#New)), we actually can't configure the log level.

The only other thing we can do is to call `SetLevel` on the standard logger and *then* assigning it:

```go

var log *logrus.Logger

func init(){
    logrus.SetLevel(logrus.DebugLevel)
    log = logrus.StandardLogger()
}
```

but that's a bit clumsier, and it also means we're restricted to just using the standard logger.

Am I missing an alternative?

flashmob:
:+1:  Had a similar situation.

Perhaps instead `logrus.New()` try instantiating logger on your own:

```go
&log.Logger{
    Out: os.Stderr,
    Formatter: new(log.TextFormatter),
    Hooks: make(log.LevelHooks),
    Level: log.DebugLevel,
}
```

ChimeraCoder:
Unfortunately that doesn't work for us, because we need to be able to set the log level based on the configuration file that's passed. The only way to do that and also preserve the `var log *logrus.Logger` declaration as a drop-in replacement for the standard library `log` is to have `SetLevel`. Otherwise, it still introduces a race condition.


@sirupsen I'm happy to put together a PR for this if it's something you're interested in adding.

flashmob:
Yes, that's still a limitation - can't change the log level after it has been set,.

Would be interested in testing out your PR too.

sirupsen:
This makes sense, feel free to propose a PR upstream.

DmitriyMV:
@ChimeraCoder @sirupsen 
because #512  had been merged - maybe it will make sense to change from `func (logger *Logger) setLevel(level Level)` to `func (logger *Logger) SetLevel(level Level) {` to make this method publicly available? Any downsides?

dmathieu:
This has shipped 🎉

sirupsen:
🎉

PR #574, opened 2017-07-07
Export logger's set level function to address #503.

stevejarvis:
There's no way to set the log level on an `New()` logger. This just exports the existing `setLevel` function, as suggested in #503.

### DIFF
--- a/exported.go
+++ b/exported.go
@@ -31,7 +31,7 @@ func SetFormatter(formatter Formatter) {
 func SetLevel(level Level) {
 	std.mu.Lock()
 	defer std.mu.Unlock()
-	std.setLevel(level)
+	std.SetLevel(level)
 }
 
 // GetLevel returns the standard logger level.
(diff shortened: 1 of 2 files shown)


stevejarvis:
@ChimeraCoder @sirupsen @DmitriyMV does it seem like this would resolve the issue now?

dmathieu:
This is looking good to me. Could you add a changelog entry (in 1.0.2)?

DmitriyMV:
LGTM - I was actually wondering what happened to the original proposal. We should probably also document that all Entry's struct objects have the Level of Logger which created them, and those can be changed dynamically.

stevejarvis:
I see the changelog entry was added in a3f95b5c423. Thanks!

DmitriyMV:
We should probably close #503 now. @dmathieu ?

dmathieu:
Thank you @DmitriyMV 😄

PR #575, opened 2017-07-10
Safely format data when printing

### LINKED ISSUE #531: Unsafe escaping of log data

seiffert:
Fixes #531

I'm taking this over from @kpcyrd who started in https://github.com/sirupsen/logrus/pull/573 and will take care of the changes discussed in the initial PR:

> Either we should remove the `QuoteCharacter` option completely or make sure that it is used in a safe way.

I decided that removing the option is not a good solution because probably people depend on it. Therefore, we need to implement proper escaping.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.0.1
+
+* bug: fix escaping in text formatter (#575)
+
 # 1.0.0
 
 * Officially changed name to lower-case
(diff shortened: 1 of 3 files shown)


seiffert:
PTAL @sirupsen @kpcyrd 
FYI @bonko @denderello

sirupsen:
Thoughts @dmathieu?

sirupsen:
Please add an entry to the changelog for the next patch version, and then this is good to merge. Thanks for your patch!

seiffert:
Would you mind tagging the v1.0.1 now already? Or are you planning to release sometimes in the near future?
Thanks!

sirupsen:
Will tag it now :)

sirupsen:
Done! `v1.0.1` is out.

PR #578, opened 2017-07-12
Fields object support directly print log information, add the hook stack of print source line

penhauer-xiao:
1. Fields object support direct print log information, the default print to the standard output，also can be set; And completely compatible the old way, this field in the field of direct print log application scene is very simple, and a service will generally only define a log object.
```
// If old must use WithFields to print the log
log.WithFields(logrus.Fields{
	"animal": "walrus",
	"size":   10,
}).Info("A group of walrus emerges from the ocean")

// The new print log is simpler; it prints to standard output by default.
logrus.Fields{
	"animal": "walrus",
	"size":   10,
}.Info("A group of walrus emerges from the ocean")

// Of course you can also be associated with the specified log object, 
// for example, where the log is specified
logrus.SetFieldsLogger(log)
logrus.Fields{
	"animal": "walrus",
	"size":   10,
}.Info("A group of walrus emerges from the ocean")
```

2. Add the stack of print source lines to the line; does not affect the old, and fully compatible with the old way; and does not affect the performance, because only when you add the hook CodeLineHook, it will be executed; Examples of specific examples: example / hook / stack.go
```
package main

import (
	"github.com/sirupsen/logrus"
	"github.com/sirupsen/logrus/hooks/source_file"
)

var log = logrus.New()

func init() {
	log.Formatter = new(logrus.TextFormatter) // default
	log.Hooks.Add(&source_file.CodeLineHook{LogLevel: logrus.DebugLevel})
}

func main() {
	log.WithFields(logrus.Fields{
		"animal": "walrus",
		"size":   10,
	}).Info("A group of walrus emerges from the ocean")

	log.WithFields(logrus.Fields{
		"omg":    true,
		"number": 122,
	}).Warn("The group's number increased tremendously!")

	// If you set FieldsLogger, you can print the log directly to the object Fields
	logrus.SetFieldsLogger(log)
	logrus.Fields{
		"animal": "walrus",
		"size":   10,
	}.Info("A group of walrus emerges from the ocean")

	log.WithFields(logrus.Fields{
		"omg":    true,
		"number": 100,
	}).Fatal("The ice breaks!")
}
```

3. Improve the potential consumption performance in high concurrent scenes at the entry level, such as:
```
func (entry *Entry) Info(args ...interface{}) {
	if entry.Logger.level() >= InfoLevel {
		entry.WithSkip(skip_4).log(InfoLevel, fmt.Sprint(args...))
	}
}
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,5 +8,8 @@ env:
   - GOMAXPROCS=4 GORACE=halt_on_error=1
 install:
   - go get github.com/stretchr/testify/assert
+  - go get github.com/kr/pretty
+  - go get github.com/kr/text
+  - go get github.com/kr/pty
 script:
   - go test -race -v .
--- a/entry.go
+++ b/entry.go
@@ -98,13 +98,13 @@ func (entry *Entry) WithSkip(skip int) *Entry {
 
 // This function is not declared with a pointer value because otherwise
 // race conditions will occur when using multiple goroutines
-func (entry *Entry) log(level Level, msg string) {
+func (entry Entry) log(level Level, msg string) {
 	var buffer *bytes.Buffer
 	entry.Time = time.Now()
 	entry.Level = level
 	entry.Message = msg
 
-	if err := entry.Logger.Hooks.Fire(level, entry); err != nil {
+	if err := entry.Logger.Hooks.Fire(level, &entry); err != nil {
 		entry.Logger.mu.Lock()
 		fmt.Fprintf(os.Stderr, "Failed to fire hook: %v\n", err)
 		entry.Logger.mu.Unlock()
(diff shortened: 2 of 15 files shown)


dmathieu:
Hey,

A general advice to PRs in open-source projects: if you want them assessed/merged quickly, it's much better to have smaller, more focused ones which do a single thing.

While your changes here could be interesting, there's a lot of things to them, some of these things being breaking changes.
Indeed, improving the API would be nice. But it's something which needs to be collectively decided, not just thought by one and merged because it was implemented.

We recently created a #logrus channel on the [gophers slack](https://invite.slack.golangbridge.org/) to discuss the evolutions of this project. How about joining there so we can better discuss the ideas you have for the API?

Regarding the performance improvement, do you have benchmarks showing the actual improvement?
All of software is a fine balance between code readability and speed. Your changes might make the code faster, but they also make it harder to read, rendering slower to evolve. We trade speed somewhere for slowness somewhere else 😄

penhauer-xiao:
@dmathieu  Thank you, I understand that there is time I will be actively involved. The current late with the new features have been six months, very good

dmathieu:
@penhauer-xiao I'm going to close this PR then. Feel free to keep hacking in your fork though.

PR #580, opened 2017-07-12
Remove QuoteCharacter option

### LINKED ISSUE #531: Unsafe escaping of log data

dmathieu:
This seems to be one of the most reported issues, as it makes it a lot rarder to safely escape strings.

This option is very much an edge case, and it's causing too much issues compared to what it provide.

Closes #531

cc @sirupsen @bbrks @seiffert @kpcyrd

### DIFF
--- a/text_formatter_test.go
+++ b/text_formatter_test.go
@@ -3,10 +3,10 @@ package logrus
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
-	"fmt"
 )
 
 func TestQuoting(t *testing.T) {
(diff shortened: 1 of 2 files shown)


text_formatter_test.go
```diff
@@ -72,36 +60,13 @@ func TestEscaping_DefaultQuoteCharacter(t *testing.T) {
...
-}
-
 func TestTimestampFormat(t *testing.T) {
 	checkTimeStr := func(format string) {
 		customFormatter := &TextFormatter{DisableColors: true, TimestampFormat: format}
 		customStr, _ := customFormatter.Format(WithField("test", "test"))
 		timeStart := bytes.Index(customStr, ([]byte)("time="))
 		timeEnd := bytes.Index(customStr, ([]byte)("level="))
-		timeStr := customStr[timeStart+5+len(customFormatter.QuoteCharacter) : timeEnd-1-len(customFormatter.QuoteCharacter)]
+		timeStr := customStr[timeStart+5+len("\"") : timeEnd-1-len("\"")]
```
PurpureGecko: The `len("\"")` always 1. Any reason to not use constant?

dmathieu: No particular reason, no. This is in tests though, there is no huge speed issues.

PR #581, opened 2017-07-12
Vendorize dependencies

dmathieu:
So we can better track what is being used, and with which version.
Also, so we don't need to manually install dependencies in `.travis.yml`.


dmathieu:
@PurpureGecko could you comment on your 👎 ?
Vendoring dependencies is the recommended thing to do in Go apps these days. What's your thinking on being against it?

PurpureGecko:
@dmathieu it is ok for apps, but not for libraries. Size of logrus will increase greatly. 
How it will works if user of logrus use same library as logrus? Is this library will be vendored several times? I think it is ok to add Gopkg.toml. But not all libraries.

seiffert:
Vendoring dependencies is not recommended in Go libraries because it can 

* create issues depending on the vendor management tool
* be a cause for diamond problems

See these discussions for reference: https://groups.google.com/forum/#!topic/golang-dev/4FfTBfN2YaI and https://groups.google.com/forum/#!topic/golang-dev/nMWoEAG55v8%5B26-50%5D

dmathieu:
All right, let's keep it as is for now.

Issue #582, opened 2017-07-12
Text-Formatter not Quoting non-string Values


seiffert:
Hi,
I've noticed that all values that aren't `string`s or `error`s are written to the output buffer without any quoting. This causes issues when a value is e.g. a `time.Time` struct:

```
msg="..." updated_at=2017-07-12 14:24:59 +0000 UTC foo=bar
``` 

From such a log line, logfmt parsers extract fields like `"14:24:59": true` and `"+0000": true` and so on.

Currently, these values are written [like this](https://github.com/sirupsen/logrus/blob/10e5e38b53df0f5f80e72a929c01af6c84d7f847/text_formatter.go#L187): 

```go
fmt.Fprint(b, value)
```

My suggestion is to first string-encode the value using `fmt.Sprint` into a variable and then apply the regular and optional quoting. Will send a PR soon.

dmathieu:
Good idea! It would also probably be a good opportunity to use `b.WriteString` and be consistent in that method.

PR #583, opened 2017-07-12
Quote non-string Values

seiffert:
fixes https://github.com/sirupsen/logrus/issues/582

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -169,27 +169,21 @@ func (f *TextFormatter) appendKeyValue(b *bytes.Buffer, key string, value interf
 }
 
 func (f *TextFormatter) appendValue(b *bytes.Buffer, value interface{}) {
+	var stringVal string
+
 	switch value := value.(type) {
 	case string:
-		if !f.needsQuoting(value) {
-			b.WriteString(value)
-		} else {
-			b.WriteString(f.quoteString(value))
-		}
+		stringVal = value
 	case error:
-		errmsg := value.Error()
-		if !f.needsQuoting(errmsg) {
-			b.WriteString(errmsg)
-		} else {
-			b.WriteString(f.quoteString(errmsg))
-		}
+		stringVal = value.Error()
 	default:
-		s := fmt.Sprint(value)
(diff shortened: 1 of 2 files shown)


text_formatter_test.go
```diff
@@ -83,6 +84,16 @@ func TestEscaping_DefaultQuoteCharacter(t *testing.T) {
 	}
 }
 
+func TestEscaping_Time(t *testing.T) {
```
dmathieu: Should this be specific to `Time`? How about naming it `TestEscaping_Interface` and trying a couple more types?

text_formatter.go
```diff
@@ -169,22 +169,21 @@ func (f *TextFormatter) appendKeyValue(b *bytes.Buffer, key string, value interf
...
-		} else {
-			b.WriteString(f.quoteString(errmsg))
-		}
+		stringVal = value.Error()
 	default:
-		fmt.Fprint(b, value)
+		stringVal = fmt.Sprint(value)
+	}
+
+	if !f.needsQuoting(stringVal) {
```
dmathieu: Nice dry-up!

text_formatter_test.go
seiffert: Good point! 
While doing that, I discovered that `error` is handled by `fmt.Sprint` exactly like we did manually (because of https://golang.org/src/fmt/print.go?s=6273:6320#L590). Therefore I removed the special case for `error`. 

The only thing left now is a check whether the value already is a string. We could remove that one too because `fmt.Sprint` would return the string as is. However, I don't know whether `fmt.Sprintf` or the type assertion is faster. There could be a small performance trade-off... What do you think?

dmathieu:
Looks like tip doesn't like your tests, and adds `m=+0.424702556` to the time string.

seiffert:
Yeah, that must be due to the changes [here](https://github.com/golang/go/commit/0e3355903d2ebcf5ee9e76096f51ac9a116a9dbb). I've changed the test to directly compare what we are expecting without depending on an explicit format.

PR #591, opened 2017-07-21
Fix trailing space

### LINKED ISSUE #99: text log should remove trailing space

dmathieu:
Closes #99

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -154,11 +154,12 @@ func (f *TextFormatter) needsQuoting(text string) bool {
 }
 
 func (f *TextFormatter) appendKeyValue(b *bytes.Buffer, key string, value interface{}) {
-
+	if b.String() != "" {
+		b.WriteByte(' ')
+	}
 	b.WriteString(key)
 	b.WriteByte('=')
 	f.appendValue(b, value)
-	b.WriteByte(' ')
 }
 
 func (f *TextFormatter) appendValue(b *bytes.Buffer, value interface{}) {
--- a/text_formatter_test.go
+++ b/text_formatter_test.go
@@ -9,6 +9,25 @@ import (
 	"time"
 )
 
+func TestFormatting(t *testing.T) {
+	tf := &TextFormatter{DisableColors: true}
+
+	testCases := []struct {
+		value    string
+		expected string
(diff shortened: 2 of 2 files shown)


text_formatter.go
```diff
@@ -154,11 +154,12 @@ func (f *TextFormatter) needsQuoting(text string) bool {
 }
 
 func (f *TextFormatter) appendKeyValue(b *bytes.Buffer, key string, value interface{}) {
-
```
seiffert: ❤️

text_formatter.go
```diff
@@ -154,11 +154,12 @@ func (f *TextFormatter) needsQuoting(text string) bool {
...
+	if b.String() != "" {
+		b.WriteByte(' ')
```
seiffert: I wonder whether the buffer could be a string slice (then we could just use `strings.Join()`). However this refactoring is out of scope of this PR an your change definitely does the job!

dmathieu: We apparently allow it to be any buffer (the client can specify it). So it doesn't seem we could do that without breaking the api.

text_formatter.go
```diff
@@ -154,11 +154,12 @@ func (f *TextFormatter) needsQuoting(text string) bool {
 }
 
 func (f *TextFormatter) appendKeyValue(b *bytes.Buffer, key string, value interface{}) {
-
+	if b.String() != "" {
```
stevvooe: This is going to cost allocation of the whole buffer. If this is used in a loop, a string will be created when appending each value, generating a lot of garbage for every log message. It would be better to test the length of the buffer.

dmathieu: Good idea. Do you want to open a PR?

stevvooe: I'm not sure if I have time today, but it would be good to get this fixed.

PR #593, opened 2017-07-25
Use buffer length to avoid generating strings every time

dmathieu:


stevvooe:
LGTM

Thanks, @dmathieu

PR #594, opened 2017-07-25
Use x/sys/unix instead of syscall

### LINKED ISSUE #500: golang.org/x/sys/unix not work now
golang.org/x/sys/unix  this package seem no longer there, redirect to github.com/golang/sys now ,
would this be  changed ?

### LINKED ISSUE #515: Switch syscall to x/sys
See https://github.com/opencontainers/runc/issues/1364 and  https://github.com/docker/docker/issues/31925

ppc64le is skipping some docker integration tests b/c this is broken. They weren't a big deal at the time but now that we know the root cause it would be great to address it. There may be other things wrong for other architectures as well, since syscall is being replaced by x/sys.

I'm working on this for runc, and then will try to get to docker/docker, so I figured I would open this as well since we know it's affecting ppc64le there as well.

dmathieu:
Closes #515 
Closes #500 
Closes #550

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,6 @@ install:
   - go get github.com/stretchr/testify/assert
   - go get gopkg.in/gemnasium/logrus-airbrake-hook.v2
   - go get golang.org/x/sys/unix
+  - go get golang.org/x/sys/windows
 script:
   - go test -race -v ./...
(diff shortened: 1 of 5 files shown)


clnperez:
Hey thanks @dmathieu! Any idea when the next version of logrus will be released? Whenever that is, I can submit a PR to moby/moby to re-vendor and remove the remaining NotPpc64le requirements from  the integration tests. Not urgent by any means.

dmathieu:
Unless moby is using a specific tag, you should be able to re-vendor right away.

PR #597, opened 2017-07-25
IsTerminal should work in nacl too

DAddYE:
I'm running logrus on a nacl environment and I get:

```
github.com/sirupsen/logrus/text_formatter.go:70: undefined: IsTerminal
```


dmathieu:
LGTM, thank you.

PR #600, opened 2017-07-27
Add appveyor.yml

AndreKR:


dmathieu:
@sirupsen it looks like only you can enable the main project's appveyor builds.
See https://ci.appveyor.com/project/sirupsen/logrus

sirupsen:
Done, do we prefer this instead of Travis? 

https://ci.appveyor.com/project/sirupsen/logrus

Issue #599, opened 2017-07-27
undefined: windows.Syscall


AndreKR:
I'm on

> ```
> go version go1.8 windows/amd64
> ```
> 
and even just downloading logrus causes:
> ```
> src\github.com\sirupsen\logrus\terminal_windows.go:78: undefined: windows.Syscall
> ```

I'm pretty sure my Go installation is not modified and according to https://godoc.org/golang.org/x/sys/windows there is in fact no function `Syscall()` in `golang.org/x/sys/windows`. I don't understand where it's supposed to come from.

AndreKR:
Seriously, I have no idea how this could ever have worked.

The same error happens on AppVeyor:
https://ci.appveyor.com/project/AndreKR/logrus/build/4

@sirupsen I prepared a PR with an `appveyor.yml`: https://github.com/sirupsen/logrus/pull/600. If you would sign in with your GitHub account on https://www.appveyor.com and simply add sirupsen/logrus as a project (no further settings necessary) we would have automated tests on Windows like we already have with Travis for Linux.

AnotherGoogleFan:
Yes, I got the same problem as this issue.

dmathieu:
Hum, yeah we removed the usage of `syscall` yesterday, and it seems I broke windows at the same time.
Thanks for the appveyor setup, it looks like a good idea.

dmathieu:
Looks like the windows API is different from the unix one, when it was the same with syscall.
I don't have easy access to a windows machine. Do one of you want to try fixing this? It should just be updating `terminal_windows.go` to use the right method.

michaelvlaar:
I've found a solution to fix this problem.

The package `golang.org/x/crypto/ssh/terminal` can be used to detect if the StdErr is a terminal. I'll send a PR to fix this problem in just a minute.

levrik:
How to fix that temporarily? I'm not able to build a single Go project...

AndreKR:
It's already fixed, update your dependencies.

levrik:
@AndreKR Mhhh already did a `go get -u`. Will try again.

PR #603, opened 2017-07-27
Fixed issue #599

michaelvlaar:
Fixed issue #599

### DIFF
--- a/terminal_windows.go
+++ b/terminal_windows.go
@@ -15,8 +15,8 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
-	"unsafe"
 
+	"golang.org/x/crypto/ssh/terminal"
 	"golang.org/x/sys/windows"
 )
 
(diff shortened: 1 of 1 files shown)


dmathieu:
If this does work, why do we keep the multi-os terminal files? Couldn't we now have a single one?

michaelvlaar:
I've only tested this on windows. For windows this works, it looks like it may also work on the other platforms. I'm not able to test on linux for now.. I would like a single function as well.

dmathieu:
Provided it works, we could actually even entirely retire that method.

samuelakerman:
Hello. Was this issue fixed? I am still getting:
terminal_windows.go:78: undefined: windows.Syscall

dmathieu:
This isn't merged. Also, there's another PR resolving it on #604.
Feel free to try it and confirm it works on your OS.

AnotherGoogleFan:
I tried go get -u -v just now and find it still cannot be build. 
```terminal_windows.go:78: undefined: windows.Syscall```
still this error.

PR #602, opened 2017-07-27
fix windows syscall bug

AnotherGoogleFan:
fix a bug issue #599

PR #604, opened 2017-07-27
Remove os-specific IsTerminal methods

### LINKED ISSUE #599: undefined: windows.Syscall

dmathieu:
Use x/crypto/ssh/terminal instead

Closes #599 
Closes #603 
Closes #602 

Ping @sirupsen, there might be specificities I'm missing preventing us from doing this simplification.


sirupsen:
@dmathieu interesting, when was this put into the standard library?

dmathieu:
Looks like it's been around since [2012](https://github.com/golang/crypto/commit/521edf84a06b2769f7a673319dcaee4024d4ecff).

dmathieu:
I'm going to merge this. The bug is causing too many issues anyway.

text_formatter.go
```diff
@@ -61,7 +65,16 @@ type TextFormatter struct {
...
 func (f *TextFormatter) init(entry *Entry) {
 	if entry.Logger != nil {
-		f.isTerminal = IsTerminal(entry.Logger.Out)
+		f.isTerminal = f.checkIfTerminal(entry.Logger.Out)
+	}
+}
+
+func (f *TextFormatter) checkIfTerminal(w io.Writer) bool {
+	switch v := w.(type) {
+	case *os.File:
```
sirupsen: Sorry for the delay. Is `Stderr` considered an `os.File` by Go?

dmathieu: [it should be, yes](https://golang.org/pkg/os/#pkg-variables). But this hasn't changed. We're still checking the type the exact same way we were before.

sirupsen: Ah yes, I should've seen that. :-)

One note is that this actually breaks backwards compatibility. `IsTerminal` is unfortunately exported, but we recently changed the signature (didn't realize people relied on it) and it broke things for a few people. I think it's reasonable to _not_ provide this going forwards—just a heads up.

dmathieu: This is on master. If we do a new 1-x release, we shouldn't cherry-pick it.

clnperez:
Soooo it looks like this re-broke ppc64le, since the crypto package is *still* using the now-deprecated syscall package. So close! ;)

tsuna:
Sorry for the rant but pulling in over 1000 lines of code from `ssh/terminal` just for `isatty()` doesn't strike me as a great move... I understand it's annoying that this code varies so much across platforms and that there is nothing in the standard library or in `golang.org/x/sys/unix`, but come on...

PR #605, opened 2017-07-27
Fix creating temp files on non-unix platforms

dmathieu:
Now that we run the tests on windows, we need to handle non-unix paths.


dmathieu:
Merci Joris 😄

PR #606, opened 2017-07-27
Ability to change the message formatting of the TextFormatter as well as the ability to turn off empty messages.

markbates:
Previous if you were to call the logger with an empty message you would get something like this:

```
INFO                                              db=0s duration="554.299µs" human_size="9.0 kB" method=GET params="{}" path=/ render="227.675µs" request_id=010483c6e8-472c3590e0 size=9016 status=200
```

The big space at the beginning was rather weird and ugly.

This PR adds two things. The first is the ability to format the space where the message goes. Previously it was hard coded to `%-44s`.

The second is the ability to disable printing empty messages.

This PR is completely backward compatible, there is no breaking changes, and people's logs will continue to behave just as they do now.

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -56,12 +56,22 @@ type TextFormatter struct {
 	// Whether the logger's out is to a terminal
 	isTerminal bool
 
+	// Format the message a particular way. The default is ' %-44s '.
+	MessageFormat string
+
+	// Disable logging empty messages. If set to true empty messages will not be
+	// printed in the log statement
+	DisableEmptyMessages bool
+
 	sync.Once
 }
 
 func (f *TextFormatter) init(entry *Entry) {
 	if entry.Logger != nil {
 		f.isTerminal = IsTerminal(entry.Logger.Out)
+	}
+	if f.MessageFormat == "" {
+		f.MessageFormat = " %-44s "
 	}
 }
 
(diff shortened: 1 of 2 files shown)


text_formatter_test.go
```diff
@@ -28,6 +28,51 @@ func TestFormatting(t *testing.T) {
 	}
 }
 
+func TestColoredStringFormatting(t *testing.T) {
```
dmathieu: This test is very similar to the previous one. Couldn't we merge them?

text_formatter_test.go
```diff
@@ -28,6 +28,51 @@ func TestFormatting(t *testing.T) {
...
+		e := &Entry{Message: tc.message}
+		b, _ := tf.Format(e)
+
+		if string(b) != tc.expected {
+			t.Errorf("formatting expected for %q (result was %q instead of %q)", tc.message, string(b), tc.expected)
+		}
+	}
+}
+
+func TestEmptyStringFormatting(t *testing.T) {
```
dmathieu: Same here. Do we really need three different tests actually doing the same thing code-wide?

text_formatter.go
```diff
@@ -56,13 +56,23 @@ type TextFormatter struct {
 	// Whether the logger's out is to a terminal
 	isTerminal bool
 
+	// Format the message a particular way. The default is ' %-44s '.
+	MessageFormat string
+
+	// Disable logging empty messages. If set to true empty messages will not be
+	// printed in the log statement
+	DisableEmptyMessages bool
```
dmathieu: Urgh not yet more options 😞

markbates:
I'm open for suggestions now to make this more extensible. When trying to create my own formatter, I know I ran into lots of non-exported functions, which meant I had to basically copy everything and tweak the couple of lines I needed. I couldn't just embed this type and change one function I needed.

As for the tests, I can merge them if you'd like, I kept them separate since, while they're similar they're testing different things.

text_formatter.go
```diff
@@ -127,12 +137,17 @@ func (f *TextFormatter) printColored(b *bytes.Buffer, entry *Entry, keys []strin
...
+		fmt.Fprintf(b, "\x1b[%dm%s\x1b[0m", levelColor, levelText)
 	} else if !f.FullTimestamp {
-		fmt.Fprintf(b, "\x1b[%dm%s\x1b[0m[%04d] %-44s ", levelColor, levelText, int(entry.Time.Sub(baseTimestamp)/time.Second), entry.Message)
+		fmt.Fprintf(b, "\x1b[%dm%s\x1b[0m[%04d]", levelColor, levelText, int(entry.Time.Sub(baseTimestamp)/time.Second))
 	} else {
-		fmt.Fprintf(b, "\x1b[%dm%s\x1b[0m[%s] %-44s ", levelColor, levelText, entry.Time.Format(timestampFormat), entry.Message)
+		fmt.Fprintf(b, "\x1b[%dm%s\x1b[0m[%s]", levelColor, levelText, entry.Time.Format(timestampFormat))
 	}
+
+	if !f.DisableEmptyMessages || entry.Message != "" {
```
dmathieu: Hum so if the message is empty, you display the timestamp, but not the message itself? Why is so?

text_formatter.go
```diff
@@ -56,13 +56,23 @@ type TextFormatter struct {
...
+	DisableEmptyMessages bool
+
 	sync.Once
 }
 
 func (f *TextFormatter) init(entry *Entry) {
 	if entry.Logger != nil {
 		f.isTerminal = IsTerminal(entry.Logger.Out)
 	}
+	if f.MessageFormat == "" {
```
dmathieu: The thing here is that this doesn't really change the message format. It only changes the length of the space between the timestamp and the actual message. The naming is really not good.

Issue #611, opened 2017-08-03
Data race when adding hooks to non-default logger


ChimeraCoder:
(Similar to #503)

`logrus.AddHook` uses a mutex to guard the standard logger, but when a new Logger is created with `logrus.New()`, there's no corresponding helper function or mutex to guard `Logger.Hooks`.

As a result, we're ending up with a data race in our tests, caused by one goroutine calling `log.Info` and another adding a new hook.

The simplest fix would be to add `func (*Logger) AddHook(hook Hook)` as a safe helper function, which would use an internal mutex to guard the Hooks slice. That doesn't prevent people from accessing `Logger.Hooks` unsafely, but it at least provides a way to write a provably race-free codepath for people who want to.

ChimeraCoder:
I don't mind throwing something together in that shape, but please let me know first if that's the approach you'd like us to take here.

dmathieu:
This will be close by #612. Thank you for the PR.

PR #612, opened 2017-08-04
Fix data race for hooks

ChimeraCoder:
This adds a test for the bug identified in https://github.com/sirupsen/logrus/issues/611, and a corresponding fix.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -315,3 +315,9 @@ func (logger *Logger) level() Level {
 func (logger *Logger) SetLevel(level Level) {
 	atomic.StoreUint32((*uint32)(&logger.Level), uint32(level))
 }
+
+func (logger *Logger) AddHook(hook Hook) {
+	logger.mu.Lock()
+	defer logger.mu.Unlock()
+	logger.Hooks.Add(hook)
+}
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1 +1,4 @@
 logrus
+*.swp
+*.swo
+*.swn
```
dmathieu: Could you add that to your own global gitignore file and not here?

logger.go
```diff
@@ -315,3 +315,9 @@ func (logger *Logger) level() Level {
 func (logger *Logger) SetLevel(level Level) {
 	atomic.StoreUint32((*uint32)(&logger.Level), uint32(level))
 }
+
+func (logger *Logger) AddHook(hook Hook) {
```
dmathieu: This is only used by the test. It shouldn't be a public method. Does it even need to exist at all?

.gitignore
ChimeraCoder: (removed)

ChimeraCoder:
Removed the `.gitignore` update from this PR; this should be good to go now!

ChimeraCoder:
Thanks!

PR #616, opened 2017-08-10
Replace example files with testable examples

rafecolton:
Closes #596


seiffert:
I've released v1.0.3 so you can depend on a version with this change.

rafecolton:
Thank you sir!

PR #631, opened 2017-08-28
New systemd's Journal hook supports log messages of any size type, st…

ssgreg:
The new systemd's Journal hook:

- supports log messages of any size and type

- supports string and binary arrays as is (natively for systemd)

- allows to close journal (registers exit handler) in order not to lose log entries

- real systemd-journald tests


allenluce:
@ssgreg Should you not have a pull request with your changes from `ssgreg/journalhook` to `wercker/journalhook`?

Issue #634, opened 2017-09-06
App Engine broken in v1 (regression)


ernestoalejo:
Even if this library is not used directly inside App Engine it is in fact imported multiple times in dependencies we use. It was working previously since #310 added support for it until this commit: e66f22976feaed4e0a6ab43bc2b9226be8b0b979

The only thing standing in the way is `golang.org/x/crypto/ssh/terminal`, which imports `golang.org/x/sys/unix/bluetooth_linux.go` and fails. I refactored the method `checkIfTerminal` to a package private function to add build tags and it works now and tests pass successfully.

josegonzalez:
This can be closed :)

PR #635, opened 2017-09-06
Split terminal check to add build tags to support App Engine.

ernestoalejo:
Related issue: #634


jprobinson:
Howdy @sirupsen, can we please merge this?

This is forcing us to pin this dependency on _many_ services.

PR #639, opened 2017-09-21
Correct "end of line" symbol for Windows OS

### LINKED ISSUE #637: Incorrect Line Separator for Windows in text_formatter
To separate strings, one character is used without checking the current OS

https://github.com/sirupsen/logrus/blob/v1.0.3/text_formatter.go#L123
```go
func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
...
b.WriteByte('\n') //but for Windows OS need "\r\n"
return b.Bytes(), nil
```

ZurgInq:
- closes #637

### DIFF
--- a/formatter.go
+++ b/formatter.go
@@ -1,8 +1,13 @@
 package logrus
 
-import "time"
+import (
+	"runtime"
+	"time"
+)
 
 const defaultTimestampFormat = time.RFC3339
+
+var EndOfLine string
 
 // The Formatter interface is used to implement a custom Formatter. It takes an
 // `Entry`. It exposes all the fields, including the default ones:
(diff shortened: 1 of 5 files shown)


formatter.go
```diff
@@ -1,9 +1,14 @@
...
 
-import "time"
+import (
+	"runtime"
+	"time"
+)
 
 const defaultTimestampFormat = time.RFC3339
 
+var EndOfLine string
```
sirupsen: Do we have to make this public?

ZurgInq: Sorry, I will use google translate.
This is necessary to override the variable in `example_basic_test.go`. I got a build error for "go version go1.9 windows/amd64" https://ci.appveyor.com/project/sirupsen/logrus/build/46#L113 (line 113)
The magic comment "Output:" (https://golang.org/pkg/testing/#hdr-Examples) does not work with the separator "\r\n". I do not know how to fix this.

thaJeztah:
Closing for now as this PR is no longer in a mergeable state, and the project is mostly in maintenance mode.

as to the change itself; while Windows uses `\r\n`, it's common to use `\n` line endings; also as described in https://jsonlines.org, which does describe `\n` as the canonical newline ... but does _allow_ for `\r\n`;

> 3. Line Terminator is '\n'
>
> This means '\r\n' is also supported because surrounding white space is implicitly ignored when parsing JSON values.

I expect users of logrus to already have adopted the current format, and changing this to use Windows line-endings probably has little benefits from that perspective.

PR #643, opened 2017-10-02
Add Trace level logging

### LINKED ISSUE #629: Add Trace level logging
It would be very convenient if logrus had a TRACE level. There are times when I want to differentiate between log messages indicating where I am in the program vs. the on the wire messages (which can be very large).

gonix:
Closes #629.


dgsb:
duplicate of #844

PR #648, opened 2017-10-11
Fix the packages have been removed bug

liyuliang2013:
Fix the packages golang.org/x/crypto and golang.org/x/sys have been removed bug


dgsb:
hello @liyuliang2013 what are you trying to achieve ? Do you have any issue to build logrus we can help with ?

PR #649, opened 2017-10-17
Adding a flag to TextFormatter to omit level=L

jmervine:
In certain cases, consumers may not want to display the log level.
They should have the option to omit it.


dgsb:
No one has expressed this needs in any issue. Closing.

PR #652, opened 2017-10-20
Add TRACE level logging.

xandrewrampulla:


Random-Liu:
Can we just do this? @sirupsen 

This is really simple change and very useful.

stevvooe:
LGTM

I would say this needs a test or two.

I do warn that this could be backward incompatible, depending on how people employ the `FieldLogger` interface, but it should be trivially addressable. I'll leave it to another maintainer to make that call.

stevvooe:
@xandrewrampulla I think it might be good to pull in a unit test, as well.

stevvooe:
Ping!

@xandrewrampulla Could you pull in a test from #643? Or should we just pull in #643?

BenTheElder:
Looks like this PR needs a rebase, is there any chance of reviving it? @xandrewrampulla @stevvooe

loren-osborn:
Rebased as requested here: https://github.com/xandrewrampulla/logrus/pull/1

dgsb:
duplicate of #844

PR #653, opened 2017-10-20
Add ability to register default fields to be logged with every message

### LINKED ISSUE #444: Add possibility to create default Fields.
There are many things, what must be in every entry. For example appname or hostname.
This can be added to Logger or to Formatters. I think modify Formatters is better decision.

Alex-PK:
Hi,

this is a proposal to fix #444 , as this is a feature we needed in our projects, and passing around the logger is not always feasible for us.

If the solution looks good to you, I can add more comments and documentation to complete it, and/or more tests if required.

If it's not, I am open to try a different way. :)

Thanks!

### DIFF
--- a/exported.go
+++ b/exported.go
@@ -48,6 +48,20 @@ func AddHook(hook Hook) {
 	std.Hooks.Add(hook)
 }
 
+// AddDefault adds a default field to the standard logger
+func AddDefault(key string, value interface{}) {
+	std.mu.Lock()
+	defer std.mu.Unlock()
+	std.AddDefault(key, value)
+}
+
+// AddDefaults adds a default field to the standard logger
+func AddDefaults(fields Fields) {
+	std.mu.Lock()
+	defer std.mu.Unlock()
+	std.AddDefaults(fields)
+}
+
 // WithError creates an entry from the standard logger and adds an error to it, using the value defined in ErrorKey as key.
 func WithError(err error) *Entry {
 	return std.WithField(ErrorKey, err)
--- a/logger.go
+++ b/logger.go
@@ -31,6 +31,8 @@ type Logger struct {
 	mu MutexWrap
 	// Reusable empty entry
 	entryPool sync.Pool
+	// Default fields to be added to all entries
+	defaultFields Fields
 }
 
 type MutexWrap struct {
@@ -63,15 +65,17 @@ func (mw *MutexWrap) Disable() {
 //      Formatter: new(JSONFormatter),
 //      Hooks: make(LevelHooks),
 //      Level: logrus.DebugLevel,
+//      defaultFields: make(Fields),
 //    }
 //
 // It's recommended to make this a global instance called `log`.
 func New() *Logger {
 	return &Logger{
-		Out:       os.Stderr,
-		Formatter: new(TextFormatter),
-		Hooks:     make(LevelHooks),
-		Level:     InfoLevel,
+		Out:           os.Stderr,
+		Formatter:     new(TextFormatter),
+		Hooks:         make(LevelHooks),
+		Level:         InfoLevel,
+		defaultFields: make(Fields),
 	}
 }
 
@@ -80,7 +84,7 @@ func (logger *Logger) newEntry() *Entry {
 	if ok {
 		return entry
 	}
-	return NewEntry(logger)
+	return NewEntry(logger).WithFields(logger.defaultFields)
 }
 
 func (logger *Logger) releaseEntry(entry *Entry) {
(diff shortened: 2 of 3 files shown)


xentek:
You can do that today, with something like: 

 ```golang
var logger *logrus.Entry
func init() {
    log := logrus.New()
    logger = log.WithFields(logrus.Fields{"default":"value"})
}

func main() {
    logger.Debug("Logger with Default Field(s)")
}
```

**Comment by Alex-PK** [comment]
Yes, I know. The problem with that solution is that you need to pass the logger around everywhere, adding a dependency to every function or manually running an init for every module in your application to either share the logger or use a ServiceLocator or a DIC.

For very small applications it is feasible. For medium-sized applications, with separate (sub) modules it's easier to rely on the default logger but being able to add global fields to it.

For example, in our case, we need to log a unique ID assigned to each of our microservices on every message. Passing the logger around or (worse) reading the env var on every logger call is very painful.

For now I solved it by writing a custom logger wrapping logrus, but it's hundreds of lines of code for a feature that can be implemented easily in the core.

**Comment by miton18** [comment]
I need this feature too.... Someone can merge it ?

**Comment by M2shad0w** [comment]
i need this feature

**Comment by mkadin** [comment]
Can't this be accomplished with a custom formatter?

**Comment by miton18** [comment]
nobody to test and merge ?

**Comment by JulienBreux** [comment]
Up? Please 🍺?

**Comment by stevvooe** [comment]
On every project where we do this, we just centralize the log entry into a package. It works great and remove the necessity for this upstream. https://github.com/containerd/containerd/blob/master/log/context.go shows how we do this with context, but you could also just have a global.

**Comment by rzajac** [comment]
Is there a plan to merge it?

**Comment by beck917** [comment]
great job!

**Comment by jalan** [comment]
> Can't this be accomplished with a custom formatter?

This is what we came up with as well.

> For now I solved it by writing a custom logger wrapping logrus, but it's hundreds of lines of code for a feature that can be implemented easily in the core.

It's about ten lines for what we needed, not hundreds. For example:

```go
package main

import "github.com/sirupsen/logrus"

// formatter adds default fields to each log entry.
type formatter struct {
        fields logrus.Fields
        lf     logrus.Formatter
}
  
// Format satisfies the logrus.Formatter interface.
func (f *formatter) Format(e *logrus.Entry) ([]byte, error) {
        for k, v := range f.fields {
                e.Data[k] = v
        }
        return f.lf.Format(e)
}

func main() {
        logrus.SetFormatter(&formatter{
                fields: logrus.Fields{
                        // Put your default fields here
                        "appID": "0x897ABEE2",
                        "hello": "world",
                },
                lf: &logrus.TextFormatter{
                        // Put your usual options here
                        DisableColors: true,
                        FullTimestamp: true,
                },
        })

        logrus.Info("I get a default field")
        logrus.WithFields(logrus.Fields{
                "more":    "stuff",
                "another": 12345,
        }).Error("I get a default field too")
}
```

```
time="2019-01-15T09:52:07-07:00" level=info msg="I get a default field" appID=0x897ABEE2 hello=world
time="2019-01-15T09:52:07-07:00" level=error msg="I get a default field too" another=12345 appID=0x897ABEE2 hello=world more=stuff
```

richpoirier:
Closing this as I agree that you should wrap logrus in your own interface rather than depending on it directly all throughout your application. This makes it very easy to switch loggers later, and add defaults like this.

Alex-PK:
> > Can't this be accomplished with a custom formatter?
> 
> This is what we came up with as well.

No, it can't.
The "global" fields in this case are "local" to the formatter so, for example, hooks don't get the global fields. For example, we are using rollbar, which is implemented as a hook, and the global fields get lost.

I already explained why it's not entirely feasible to have a wrapper to pass around everywhere.

Anyway... your project, your decision.

dgsb:
@Alex-PK we try to be careful when  we add new api in the Logger, because we will have to maintain it until the end of time.
Anyway we may consider integrating your change if there is no other solution, to your problem.
I suspect the global fields are lost in the hook you are using because hooks are called before formatting.
Have you tried to solve your issue, by using a hook which would add the wanted fields ?

jalan:
@Alex-PK Thanks for correcting me regarding the interaction between hooks and the formatter. I don't have any hooks, so I didn't encounter that.

Looks like using a hook to add the wanted fields works too, but then there may be a concern about the order in which hooks fire. For example, if you use the syslog hook, you need to be sure the default fields get added before the entry is sent to syslog. Is there any guarantee that hooks fire in the order in which they were added?

dgsb:
@jalan and @Alex-PK, the hooks fire in the order they are added to the Logger.
This is not clearly documented, but this is what the implementation does and I think this is what a user should expect.
We may improve the documentation to clearly state this contract and add unit test, if they do not exist, to prevent unnoticed changes on this behaviour.

dgsb:
@Alex-PK I've just added a new example https://github.com/sirupsen/logrus/blob/master/example_default_field_value_test.go which adds default field.
Let me know if this example helps you to solve your issue.

dgsb:
A code sample has been added in the repo.

bigunyak:
It is indeed very inconvenient to not being able to configure default logger with some default fields. :-/

PR #664, opened 2017-11-05
[BUGFIX] keep level and message by adding Field

genofire:


genofire:
??

genofire:
After 15 month no reaction ? -> i would not resolve the conflicts ;(

PR #666, opened 2017-11-07
enabled the jsonformatter to serialise previously unserialiseable fie…

myles-mcdonnell:
…lds by skipping them


myles-mcdonnell:
What this enables you to do is throw any struct into a log event which IMO is huge bonus, particularly for debug events....it has just occurred to me now I'm writing this that I haven't made the equivalent change for the TextFormatter though..not sure if this is an issue, guess we will find out!

myles-mcdonnell:
doh, the Formatter can be supplied by the consumer..pointless change.

PR #674, opened 2017-11-17
Allow disabling new line appending for json_formatter

### LINKED ISSUE #949: Allow disabling newline for the default text formatter
Add a new field to the struct to disable trailing new line entered after log messages, similar to #674

almonteb:
- closes https://github.com/sirupsen/logrus/issues/949

This PR allows disabling of the trailing new line with json_formatter. It retains the default behavior and adds a param to the struct to allow disabling.

### DIFF
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -33,6 +33,9 @@ type JSONFormatter struct {
 	// DisableTimestamp allows disabling automatic timestamps in output
 	DisableTimestamp bool
 
+	// DisableNewline allows disabling of the trailing newline in output
+	DisableNewline bool
+
 	// FieldMap allows users to customize the names of keys for default fields.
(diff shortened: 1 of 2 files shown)


universam1:
missing this feature as well

thaJeztah:
Closing for now as this PR is no longer in a mergeable state, and the project is mostly in maintenance mode.

as to the change itself; it's common to use `\n` line endings to separate entries; also as described in https://jsonlines.org.

I expect users of logrus to already have adopted the current format, and changing this to probably has little benefits from that perspective. If this is a hard requirement for users, it should be possible to create a [custom Formatter](https://pkg.go.dev/github.com/sirupsen/logrus#Formatter) and use that instead of the default.

PR #677, opened 2017-11-22
Add FieldMap support to TestFormatter

neilisaac:
Use the same FieldMap class used by JSONFormatter for TextFormatter.

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -60,6 +60,15 @@ type TextFormatter struct {
 	// Whether the logger's out is to a terminal
 	isTerminal bool
 
+	// FieldMap allows users to customize the names of keys for default fields.
+	// As an example:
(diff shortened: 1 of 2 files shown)


neilisaac:
@sirupsen any thoughts on this?  I hacked this up ages ago, but am happy to revise or rebase if changes are needed, as long as someone is willing to accept it.

dgsb:
@neilisaac it mostly looks good to me, but can you merge from master in order to retrieve the new prefixFieldClashes signature and gives as parameter the configured FieldMap. It would alos be nice to have an associated unit test to check potentially clashing field are appropriatly prefixed.

neilisaac:
@dgsb it looks like it wasn't quite right to begin with, so this PR now includes a bug fix.

PR #685, opened 2017-12-29
TextFormatter behaviour aligned with stdlib log (fixes #167)

### LINKED ISSUE #167: log.Printf() and log.Infof append a newline to the output
I just started using logrus by replacing the import for log. Previously I had used log.Printf("bla\n") to log messages, but logrus appends an extra \n so that there is an empty line after each log message. What am I missing? If this is supposed to be API-compatible, this should behave differently.

HectorMalot:
Hoping to help logrus with a small fix for a long-lived issues (#167). I've made the newline behavior consistent with the standard library log package. The fix to `TextFormatter` removes a single new-line (if present) at the end of the provided `Entry.Message`. e.g. `"testmessage\n\n"` will become `"testmessage\n"`. This is the same behavior as `log.Printf`

Added a test and confirmed existing tests were passing. The benchmark test shows minor impact (up to 1% increase in ns/op), but happy to discuss more efficient options

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -126,6 +126,10 @@ func (f *TextFormatter) printColored(b *bytes.Buffer, entry *Entry, keys []strin
 
 	levelText := strings.ToUpper(entry.Level.String())[0:4]
 
+	// Remove a single newline if it already exists in the message to keep
+	// the behavior of logrus text_formatter the same as the stdlib log package
+	entry.Message = strings.TrimSuffix(entry.Message, "\n")
+
 	if f.DisableTimestamp {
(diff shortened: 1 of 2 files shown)


HectorMalot:
@dgsb , I've updated the pull request to work with the latest master. Implicit newlines should now work the same as the stdlib. e.g. this prevents the negroni logger middleware from spitting out empty lines when passing logrus instead of the default log.

Would you be open to merge this or alternatively have a discussion on how the PR can be improved?

dgsb:
@HectorMalot thanks for your contribution

PR #687, opened 2018-01-04
optimise: break when logger.out is dev/null

plusplus1:


stevvooe:
While this is a good optimization for `ioutil.Discard` or a `nil` output, it won't handle cases where `/dev/null` is opened separately.

If that is sufficient, LGTM.

alexlry:
Test for Logger.Out could be made on line 100 instead of line 127. This would also prevent using bufferPool when Logger.Out is disabled.
https://github.com/keymantics/logrus/commit/5e2f1ff435f036aa7e1c5b088d1b0938bc6fd279

dgsb:
hello @plusplus1, thanks for your contribution. But I'm not sure the added complexity worth it. The use case seems narrow to me.

dgsb:
No feedback from initial commiter. The use case is rather narrow and the added value vs the added complexity does not seem to being worth it.

PR #695, opened 2018-01-22
Fix deadlock on panics at Entry.log

mauricio:
When calling Entry.log a panic inside some of the
locking blocks could cause the whole logger to deadlock.

One of the ways this could happen is for a hook to cause
a panic, when this happens the lock is never unlocked and
the library deadlocks, causing the code that is calling
it to deadlock as well.

This changes how locking happens with unlocks at defer
blocks so even if a panic happens somewhere along the log
call the library will still unlock and continue to function.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -94,38 +94,45 @@ func (entry Entry) log(level Level, msg string) {
 	entry.Level = level
 	entry.Message = msg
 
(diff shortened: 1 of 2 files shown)


aybabtme:
👍

PR #698, opened 2018-01-31
Allow to print log level as integer

### LINKED ISSUE #355: Is it possible to configure the log levels to be ints rather than strings?
We export all our logs to logstash via filebeat, and all our other systems have logging libraries which use ints for the log level rather than strings. Can this be done with logrus?

jmdalmeida:
### Reason
To be able to have logs compliant with the Syslog Protocol ([RFC5424](http://www.ietf.org/rfc/rfc5424.txt) and [RFC3164](http://www.ietf.org/rfc/rfc3164.txt)), we need to print the log levels as integers.

### Changes
Adds a flag to both Text and JSON Formatters struct that, if set to true, will make the log levels print as integers (the default is the old behaviour, i.e. print the resolved level name).

### Issues
Fixes #355.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 logrus
+.idea/
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -33,8 +33,8 @@ type JSONFormatter struct {
 	// DisableTimestamp allows disabling automatic timestamps in output
 	DisableTimestamp bool
 
-	// IntLogLevels allows users to print log levels as integers instead of transformed strings
-	IntLogLevels bool
+	// EnableIntLogLevels allows users to print log levels as integers instead of transformed strings
+	EnableIntLogLevels bool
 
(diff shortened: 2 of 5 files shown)


jmdalmeida:
Needs more development for full compliance, will open another PR when it's ready.

jmdalmeida:
Maybe the way to go would be to have both: `level` for the level number, `level_name` for the respective name.

stevvooe:
@jmdalmeida I just can't see the reason to have the number by default, and we definitely can't switch `level` to be a number. It would just break way to many users for an uncommon use case.

Have you tried using a formatter to get the output you're looking for?

thaJeztah:
Agreed with Stephen; we may need to be careful on the assumptions of the level-numbers to be stable. Probably this could be handled in a custom formatter for those that need it.

Closing for now, also due to the project being mostly in maintenance mode (so we try to avoid expanding the public API), but feel free to continue the conversation.

PR #702, opened 2018-02-05
Field logger with hooks race

thundercat1:


thundercat1:
resolves https://github.com/sirupsen/logrus/issues/703

cozzbp:
+1

PR #705, opened 2018-02-10
add success logging level, close #699

### LINKED ISSUE #699: Success logging level
For applications like CLI tools, Success is an important logging level, along with Info, Error, Debug and Fatal. I had to implement this using a custom formatter and fields. 

Is there an interest in adding success as a level? I can work on a PR quickly if this is of interest to the communinty.

shahidhk:
This PR adds support for success logging level, useful mainly for CLI applications to denote an action has been completed.

The level is kept below `InfoLevel` so that default log level (which is Info) will also prints out success logs.

The actual `int` values for `InfoLevel` and `DebugLevel` changes by `+1`, I hope that is not an issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ attached, the output is compatible with the
 ```text
 time="2015-03-26T01:27:38-04:00" level=debug msg="Started observing beach" animal=walrus number=8
 time="2015-03-26T01:27:38-04:00" level=info msg="A group of walrus emerges from the ocean" animal=walrus size=10
+time="2015-03-26T01:27:38-04:00" level=success msg="Group emerged" animal=walrus
 time="2015-03-26T01:27:38-04:00" level=warning msg="The group's number increased tremendously!" number=122 omg=true
 time="2015-03-26T01:27:38-04:00" level=debug msg="Temperature changes" temperature=-4
 time="2015-03-26T01:27:38-04:00" level=panic msg="It's over 9000!" animal=orca size=9009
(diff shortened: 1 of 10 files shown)


eexit:
This is not [RFC compliant](https://tools.ietf.org/html/rfc5424#section-6.2.1).

dgsb:
@eexit indeed, but logrus is not compliant either. @shahidhk I'm not sure to understand what this level would bring as opposed to use the Info level.

shahidhk:
@dgsb Let's say a command X is being executed, which does actions A, B, and C thus achieving X. The log output is like this:

```
$ command X
INFO action A completed
INFO action B completed
INFO action C completed
INFO command X completed
```

Instead, if it could show:

```
$ command X
INFO    action A completed
INFO    action B completed
INFO    action C completed
SUCCESS command X completed
```

it provides a better UX denoting the command has completed successfully. If an error occurs in between, user is clear that even though certain actions have completed, the command itself did not complete.

dgsb:
@shahidhk I wonder if instead of declaring a new level for a very specific need using a dedicated field with its associated value would not be more practical ?
```
logrus.WithField("command_status", "success").Info("blablabla")
```

eexit:
> I wonder if instead of declaring a new level for a very specific need using a dedicated field with its associated value would not be more practical ?

Indeed, I don't understand why adding a context specific level would be needed to cover such case. Using the fields does perfectly the job. If you need to identify a specific log workflow, you can still create a contextualized logger off a mutated field logger:

```go
func CommandX(contextID string) {
    logger := logrus.WithFields(logrus.Fields{
        "command": "X",
        "context_id": contextID,
    })

    // Action A
    logger.Info("action A completed")
    // Action B
    logger.Info("action B completed")
    // ...

    logger.Info("command completed")
}
```

shahidhk:
> If you need to identify a specific log workflow, you can still create a contextualized logger off a mutated field logger

This is what I am doing in the existing workflow. But, having worked on CLI applications for quite some time, I came to the conclusion that reporting Success was as important as reporting Error/Fatal. But I agree that it can be done via the methods mentioned above also, mine was a very particular use case.

PR #718, opened 2018-02-25
Deferred creation of log file

mipimipi:
If logging shall be done into a log file, the file had to be create passed to SetOutput. This resulted in an empty log file if no log entry has been written.
With my contribution, the creation of the log file is deferred until the first log entry is made. Thus, if no entry is necessary, no log file is created.
Changes:

- I introduced the file name as new attribute in type Logger.
- I introduced the new function SetFilename in exported.go to set the file name
- I introduced the new function setOut in logger.go that creates the file based on the file name if it is not yet existing
- The only place where setOut is called, is in entry.write just before an entry is written.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -131,6 +131,10 @@ func (entry *Entry) write() {
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to obtain reader, %v\n", err)
 	} else {
+		// make sure that log file is created (in case logging shall be done
+		// into a file)
+		entry.Logger.setOut()
+		// write log entry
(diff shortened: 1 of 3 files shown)


dgsb:
Hello @mipimipi, thanks for your contribution. However, I'm kind of reluctant to merge this pr.

Most of the time empty log files are really not an issue but quite the contrary. So the added value of the 
feature vs the complexity ratio is not quite clear.
Alos in this the implementation, if the file can not be opened for any reason, we have no way to return an error upon a trace. This is quite unfortunate, because traces are important. I think most users would rather not start the application without a functional trace system.

PR #726, opened 2018-03-10
Fix run-on sentence

wla80:


README.md
```diff
@@ -495,7 +495,7 @@ logrus.RegisterExitHandler(handler)
 
 #### Thread safety
 
-By default Logger is protected by mutex for concurrent writes, this mutex is invoked when calling hooks and writing logs.
+By default, Logger is protected by mutex for concurrent writes, and such mutex is invoked when calling hooks and writing logs.
```
stevvooe: `By default, Logger is protected by mutex for concurrent writes. The mutex is held when calling hooks and writing logs.`

wla80: Hi @stevvooe , I like your version better. Just uploaded a new commit 2efdad6. Thanks

PR #730, opened 2018-03-16
Return the Entry from fireHooks().

### LINKED ISSUE #729: v1.0.5 breaks hooks that update the Entry
We've been using a Hook to add line number and file name information to the Entry for a long time.  Unfortunately https://github.com/sirupsen/logrus/pull/707 has broken our Hook.  Since it copies the Entry, our updates to it are lost when the fireHooks method returns.

Our hook used this pattern to avoid mutating the Entry while it may have been read by another thread:
```go
newData := make(log.Fields, len(entry.Data)+2)
for k, v := range entry.Data {
	newData[k] = v
}
newData["__file__"] = path.Base(frame.File)
newData["__line__"] = frame.Line
entry.Data = newData
```
but that no longer works.  I think that simply mutating the existing data map would be broken too.

This seems to work for a quick fix since it's the copy that gets logged out instead of the original Entry:
```
- entry.fireHooks()
+ entry = entry.fireHooks()
...
- func (entry Entry) fireHooks() {
+ func (entry Entry) fireHooks() Entry {
 	entry.Logger.mu.Lock()
 	err := entry.Logger.Hooks.Fire(entry.Level, &entry)
 	if err != nil { 
 		fmt.Fprintf(os.Stderr, "Failed to fire hook: %v\n", err)
 	}
+	return entry
 }
```
I'll put up a PR for that.

fasaxc:
Ensures that any modifications done by hooks are reflected.

Fixes #729

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -94,7 +94,7 @@ func (entry Entry) log(level Level, msg string) {
 	entry.Level = level
 	entry.Message = msg
 
-	entry.fireHooks()
+	entry = entry.fireHooks()
 
 	buffer = bufferPool.Get().(*bytes.Buffer)
 	buffer.Reset()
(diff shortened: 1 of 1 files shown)


dgsb:
The initial issue has been fixed, this interface change should not be needed anymore.

PR #735, opened 2018-03-21
Fix #691 

### LINKED ISSUE #691: how log double quotes to file without backslash?
The follow string is that I want to log to file:

    {"name": "test", "id": "123xx"}

When I log this string to file , no matter the json format or text format, logrus auto add a backslash('\') before double quotes. like following:
    {\"name\": \"test\", \"id\": \"123xx\"}
But I don't want backslash. How can I remove it?

Wang-Kai:
### I change %q to %s in appendValue function, it solve this issue perfectly .

```golang
package main

import (
	log "github.com/wang-kai/logrus"
)

func main() {
	log.WithFields(log.Fields{
		"json_string": `{"name": "test", "id": "123xx"}`,
	}).Info("I solve it")
}
```

Run
```
➜  log go run main.go
INFO[0000] I solve it    json_string={"name": "test", "id": "123xx"}
```

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -173,6 +173,6 @@ func (f *TextFormatter) appendValue(b *bytes.Buffer, value interface{}) {
 	if !f.needsQuoting(stringVal) {
 		b.WriteString(stringVal)
 	} else {
-		b.WriteString(fmt.Sprintf("%q", stringVal))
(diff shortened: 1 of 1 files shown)


Wang-Kai:
I do not think test is reasonable .

Issue #690, opened 2018-01-10
adding WithStack() function


kubistika:
I though about that in my projects that are using logrus,
I use to write the stack trace to important logs. 
The way I do that now is:
`logger.WithField("stack", string(debug.Stack())`

but that is not a nice way of doing that, because I need to choose a key for that any time I want to add the stack trace to a log entry. I think it would be nice to add a function WithStack(), like WithError()

jamesbloomer:
Some error reporting systems, such as GCP Error Reporting, use the existence of a stack to decide whether a log entry is an error.  Currently we have to do:

`logger.WithError(err).Error("This is an error\n", string(debug.Stack()))`

Which is a bit cumbersome.

gguridi:
I've added a small pull request because I think it would be a nice to have feature.

markphelps:
Closing per the project being put into [maintenance mode](https://github.com/sirupsen/logrus/commit/4ecd9a62bd57b6533e0d87ae23a6b10adc32b5ce)

PR #738, opened 2018-03-27
Adding Stack Trace generator as part of the logrus capabilities.

gguridi:
This pull request has been created to help #690.

The stack trace generator is based on the example one in the [runtime package](https://golang.org/pkg/runtime/#Frames). It's quite basic.

Added a StackTraceDepth to improve the information gathered upon generating the stack trace. Ignoring by default the runtime package calls. Added a standard format for the stack traces, it might be improved in the future. Added stack trace tests. Updated the README file with the stack trace information.

---
This change is [Reviewable](https://reviewable.io/reviews/sirupsen/logrus/738)


gguridi:
runtime.CallersFrames was added in Go 1.7 so the tests for Go 1.6 are failing. Anyone knows the best approach to keep backwards compatibility when you find these things? Must everything be changed to functions available in 1.6 onwards? Is it safe to drop compatibility with Go 1.6 as most vendoring tools use commit hashes to set the dependency.

joscarsson:
Nice! However, as the commonly used https://github.com/pkg/errors adds stack traces to the `error` objects themselves (through either `errors.Wrapf()`, `errors.WithStack()` and `errors.New()`), would it be possible to make the already existing `logrus.WithError()` log the stack trace from the `error` if it exists? If I understand it correctly, the current PR would log the stacktrace to the log line, which isn't necessarily the same as where the error originated from.

gguridi:
@joscarsson Sorry for coming back to this so late. Checked the https://github.com/pkg/errors but I seem unable to retrieve the stack trace alone from inside the error object. 

Using `WithError(err)` will force the formatter to use `data[k] = v.Error()` which will print altogether the error message and the stack trace from `pkg/errors`, but I don't see how I can do more without them exposing the property stack publicly.

flimzy:
@joscarsson This should be a trivial modification of this PR. It should be easy to modify `WithError` to set both the `error` and `stacktrace` fields, if the passed error satisfies the documented [`StackTracer`](https://godoc.org/github.com/pkg/errors#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper) interface.

flimzy:
This is a nice PR. It would save me a lot of work, adding stack traces to my logs.

I would suggest modifying it to optionally support `pkg/errors` stack traces, as well. I'd be happy to contribute that change, if that would be helpful.

I do believe it needs to be an optional feature, though, because some backends already have support for `pkg/errors` (in particular, the [logrus_sentry](https://github.com/evalphobia/logrus_sentry) driver does, since [I added it](https://github.com/evalphobia/logrus_sentry/pull/23) some time ago). So there is a potential for duplicate stack traces.

gguridi:
@flimzy I think this PR is no longer necessary, as #850 was merged and provides the stack trace functionality. Any improvement I think should go on that direction. I'm just keeping this open till someone confirms me that I'm not wrong and this PR no longer makes sense :)

markphelps:
Closing per the project being put into [maintenance mode](https://github.com/sirupsen/logrus/commit/4ecd9a62bd57b6533e0d87ae23a6b10adc32b5ce)

PR #742, opened 2018-04-10
fixes #615. panic with formatted message

### LINKED ISSUE #615: Incorrect panic msg during log.Panic()
have:
```
panic: (*logrus.Entry) (0xddc300,0xc42019e700)
```
want
```
panic: some error happened.
```
where `some error happened` is msg that was logged.

mkorolyov:
- fixes https://github.com/sirupsen/logrus/issues/615

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -109,7 +109,7 @@ func (entry Entry) log(level Level, msg string) {
 	// panic() to use in Entry#Panic(), we avoid the allocation by checking
 	// directly here.
 	if level <= PanicLevel {
-		panic(&entry)
+		panic(msg)
 	}
 }
 
--- a/entry_test.go
+++ b/entry_test.go
@@ -37,15 +37,7 @@ func TestEntryPanicln(t *testing.T) {
 
 	defer func() {
 		p := recover()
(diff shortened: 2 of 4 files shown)


mkorolyov:
anybody?

loren-osborn:
logrus `*Entry` already implements the `String()` method, so is already a `fmt.Stringer`, behaving exactly as you describe.

I understand the value of loose coupling, but I believe `fmt.Stringer` gives you that.

Otherwise this PR takes an inspectable object and makes it an opaque string. I think that is a step backward. Can you give a use-case where this is helpful?

nhinds:
logrus `*Entry` does not implement [`fmt.Stringer`](https://golang.org/pkg/fmt/#Stringer) because its [`.String()` method]() returns `(string, error)` instead of `string`.

Currently `.Panic(err)` outputs `panic: (*logrus.Entry) (0x123456,0x9876543210)` to stderr. If the Entry implemented `fmt.Stringer` (tested by removing `error` from the return type of `.String()` locally) it would output a more useful `panic: time="..." ... msg="..."`, but making such a change would break compatibility with other callers of `.String()`.

PR #749, opened 2018-05-02
SetEscapeHTML false for json_formatter

ckeyer:


markphelps:
Fixed in https://github.com/sirupsen/logrus/pull/1113. Thanks!

PR #752, opened 2018-05-12
Fix a race condition in TestLoggingWithHooksRace

dgsb:


dgsb:
@sirupsen did you have a chance to look at this pr ? the fix is pretty obvious and ensure the test reproductability.

PR #755, opened 2018-05-15
PERF: use buffer pool in json formatter

taylorchu:
```
benchmark             old ns/op     new ns/op     delta
BenchmarkLogrus-8     4163          4369          +4.95%

benchmark             old allocs     new allocs     delta
BenchmarkLogrus-8     36             31             -13.89%

benchmark             old bytes     new bytes     delta
BenchmarkLogrus-8     3027          2163          -28.54%
```

```
func BenchmarkLogrus(b *testing.B) {
	log := logrus.New()
	log.Formatter = &logrus.JSONFormatter{}
	log.Out = ioutil.Discard
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		log.
			WithField("foo1", "bar1").
			WithField("foo2", "bar2").
			Info("this is a dummy log")
	}
}
```

### DIFF
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -1,6 +1,7 @@
 package logrus
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 )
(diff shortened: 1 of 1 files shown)


taylorchu:
@sirupsen this idea is also in text_formatter. Could you take a look? thanks!

sirupsen:
@stevvooe @seiffert do you mind taking a look at this?

PR #762, opened 2018-05-29
feat: new methods to check enabled log level

### LINKED ISSUE #761: Allow log level checking to prevent useless call to logging methods
To check the log level, I had to read the source code to know if DebugLevel is greater or lesser then InfoLevel.

So it can be intersting to have some `IsXxxxxEnabled` methods on `exported`, `logger` en `entry` like:
- logger.IsDebugEnabled()
- logger.IsInfoEnabled()
- logger.IsWarnEnabled()
- logger.IsErrorEnabled()
- logger.IsFatalEnabled()
- logger.IsPanicEnabled()

Also, to respect the DRY pattern, I suggest to replace duplicated `if logger.level() >= DebugLevel` by these new methods. So if the business definition of "Debug is enabled" changes, we'll have to change it only once.

lhauspie:
Adding 6 methods on `exported`, `logger` and `entry`:
- IsDebugEnabled() bool
- IsInfoEnabled() bool
- IsWarnEnabled() bool
- IsErrorEnabled() bool
- IsFatalEnabled() bool
- IsPanicEnabled() bool

Replace duplicated `if logger.level() >= XxxxLevel` by a call to the new corresponding methods in `logger` and in `entry`

Closes #761


lhauspie:
up ?

entry.go
```diff
@@ -260,24 +260,48 @@ func (entry *Entry) Warningln(args ...interface{}) {
...
 }
 
 func (entry *Entry) Panicln(args ...interface{}) {
-	if entry.Logger.level() >= PanicLevel {
+	if entry.IsPanicEnabled() {
 		entry.Panic(entry.sprintlnn(args...))
 	}
 }
 
+func (entry *Entry) IsDebugEnabled() bool {
```
dgsb: I don't think we need to implements these methods at Entry level. This property is more related to the logger and the forward call is simple enough to not add them to the Entry interface.

logrus.go
```diff
@@ -140,4 +140,11 @@ type FieldLogger interface {
 	Errorln(args ...interface{})
 	Fatalln(args ...interface{})
 	Panicln(args ...interface{})
+
+	IsDebugEnabled() bool
```
dgsb: We can not change the FieldLogger, such an interface change may raise build breaks to some users.

logger.go
```diff
@@ -321,3 +321,27 @@ func (logger *Logger) AddHook(hook Hook) {
 	defer logger.mu.Unlock()
 	logger.Hooks.Add(hook)
 }
+
+func (logger *Logger) IsDebugEnabled() bool {
```
dgsb: I'd rather have a single method `IsLogLevelEnabled` which takes a Level parameter, there would be less methods to implement, to document and no modification would be required if someday we add a new trace level.

dgsb:
@lhauspie thanks for your contribution, I have requested a few changes.

logrus.go
lhauspie: hmmm... in which case we can raise build breaks to some users ?
If in there build process they pull the master branch of logrus before building ?

entry.go
lhauspie: Removed from `Entry`

logger.go
lhauspie: `IsXxxxEnabled` refactored to `IsLevelEnabled`

logrus.go
dgsb: Indeed, that the case I was talking about. We can not lightly introduces non backward compatible change to master.

PR #764, opened 2018-05-30
Revert the change introduced in #707 and do the proper fix

moriyoshi:
The problem addressed in #707 is that data race would occur because the dummy hook retains the entries passed to it as pointers to them instead of having their copies.

https://github.com/sirupsen/logrus/blob/0afea37/hooks/test/test.go#L18

Therefore, fixing the actual implementation code is so inappropriate that causes other problems like reported in #729.  The dummy code should have been fixed instead.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -113,12 +113,10 @@ func (entry Entry) log(level Level, msg string) {
 	}
 }
 
-// This function is not declared with a pointer value because otherwise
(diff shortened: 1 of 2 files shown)


moriyoshi:
Any possible discussion on this?

moriyoshi:
Hey anyone can take a look into this?

PR #775, opened 2018-06-21
Add GopherJS compatibility build tags

paralin:
This adds build tags to exclude the unix-specific stuff from a js build.


paralin:
Wow, looks like someone else did this and I haven't pulled since..

PR #776, opened 2018-06-21
Fix GopherJS build tags

paralin:
The GopherJS build tag is "js" not "gopherjs"

### DIFF
--- a/terminal_check_appengine.go
+++ b/terminal_check_appengine.go
@@ -1,4 +1,4 @@
-// +build appengine gopherjs
+// +build appengine js
 
 package logrus
 
--- a/text_formatter_js.go
+++ b/text_formatter_js.go
@@ -1,0 +1,11 @@
+// +build js
+
+package logrus
+
+import (
+	"io"
+)
+
+func (f *TextFormatter) checkIfTerminal(w io.Writer) bool {
+	return false
+}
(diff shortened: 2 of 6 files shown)


terminal_check_appengine.go
```diff
@@ -1,4 +1,4 @@
-// +build appengine gopherjs
+// +build appengine js
```
dgsb: Do we still need the js build constraint here if we have another file dedicated for the gopherjs environment ?

text_formatter_js.go
```diff
@@ -0,0 +1,11 @@
+// +build js
```
dgsb: Can we keep the same naming convention ? This file could be named terminal_check_js.go

text_formatter_other.go
```diff
@@ -0,0 +1,3 @@
+// +build !js
```
dgsb: I'm not sure this file is useful, this build constraint should already be covered by terminal_check_not_appengine.go file.

paralin:
Honestly @dgsb you can do whatever you want with the semantics and file naming here, I don't care. The main thing is just to make it compile and run properly under GopherJS - this was the hack I had to do to get it to work as a proof-of-concept, so there you go.

cheungpat:
@paralin Thanks for your input. I hope your fix can be merged, so I created another pull request (#798) with changes suggested by @dgsb.

Issue #241, opened 2015-09-03
Missing locks for reconfiguring loggers


rnapier:
The `std` logger has locks available for reconfiguring it via `SetOutput`, `SetLevel`, `SetFormatter`, etc. But if you create a new logger, the only access is via the raw fields, which creates race conditions (and crashes in my case) if you try to modify them while logrus is logging.

Ideally, all of the `Set...` and `Get...` functions should be moved onto Logger methods, and the top-level functions should just forward to them for `std`. For safety, the struct fields should be unexported, but this would be a nontrivial change to the interface, and make it a little harder to initially configure loggers (unless some other "LoggerConfig" were provided). The current scheme, however, is very dangerous.

Sirupsen:
You're completely right. I'm worried about the amount we'll break the standard interface, but on a path to 1.0 this is better done early rather than late.

Astone-Chou:
"Ideally, all of the Set... and Get... functions should be moved onto Logger methods"

Really! Really! Really! Need to got the lock, plz.

dmathieu:
We'll get to this issue at some point. In the mean time, feel free to open a PR if you want this to be fixed faster.

dgsb:
has been fixed in 0c03a05a

PR #779, opened 2018-06-27
[#241] Allow to set writer during logger usage.

daskol:
In this pull request I am trying to partially fix issue #241. Now, one could reset output writer,


dgsb:
Thanks for your contribution @daskol, can you also change the implementation of the global SetOutput function to just forward to the logger instance method ? That would make more sense to not have duplicated implementation.

daskol:
OK. I have already fixed this. Also, I amended the commit description and appended chanages to the commit in order to keep the commit history clean.

Issue #778, opened 2018-06-26
Allow user-defined entry.Time


sbrisson2:
Hi,

I'm currently writing a wrapper to go around logrus loggers.  This wrapper stores all logs in memory until it is explicitly flushed, at which point it logs to the underlying logger.  The intended use case is in tests, where I'm only interested in the logs for tests that fail.

The problem I'm having is that I would like the timestamps in the logs to reflect the time at which my program logs to the wrapper, not the time at which the wrapper is flushed.  I thought I could achieve this by modifying the `Time` field of an `entry` object.  However, it turns out that `entry.Time` is always set to `time.Now()` whenever one of the logging methods is called (see [here](https://github.com/sirupsen/logrus/blob/master/entry.go#L93)).

I was hoping that we could avoid overriding `entry.Time` in `entry.log()` if it's nonzero.

dgsb:
For your use case, you can use a dedicated field for your initial timestamp.

sbrisson2:
That is a possible solution, however it also requires us to have special handling for the custom field, either in our formatters, or in our log consumers.  I hope you can see why this is sub-optimal for us.

dgsb:
Fixed by #781

PR #781, opened 2018-06-29
Override time in log entries

sbrisson2:
This PR allows users to have control over the timestamps in their logs.  It solves #778.

The first commit prevents `Entry.log()` from using `time.Now()` if the entry already has a non-zero `Time` value.

The second commit adds `WithTime()` methods to the `FieldLogger` , `Entry`, and `Logger` types.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -90,7 +90,16 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 // race conditions will occur when using multiple goroutines
 func (entry Entry) log(level Level, msg string) {
 	var buffer *bytes.Buffer
-	entry.Time = time.Now()
+
+	// Default to now, but allow users to override if they want.
+	//
+	// We don't have to worry about polluting future calls to Entry#log()
+	// with this assignment because this function is declared with a
+	// non-pointer receiver.
+	if entry.Time.IsZero() {
+		entry.Time = time.Now()
+	}
+
 	entry.Level = level
 	entry.Message = msg
 
--- a/entry.go
+++ b/entry.go
@@ -83,7 +83,12 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 	for k, v := range fields {
 		data[k] = v
 	}
-	return &Entry{Logger: entry.Logger, Data: data}
+	return &Entry{Logger: entry.Logger, Data: data, Time: entry.Time}
+}
+
+// Overrides the time of the Entry.
+func (entry *Entry) WithTime(t time.Time) *Entry {
+	return &Entry{Logger: entry.Logger, Data: entry.Data, Time: t}
 }
 
 // This function is not declared with a pointer value because otherwise
--- a/exported.go
+++ b/exported.go
@@ -2,6 +2,7 @@ package logrus
 
 import (
 	"io"
+	"time"
 )
 
 var (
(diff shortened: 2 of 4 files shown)


dgsb:
We can not do that, because a same entry can be used to several times. With this patch we are going to end up have all these logs with the same timestamp.

sbrisson2:
I did forget to check for that when I opened this PR.  However, upon inspection (and confirmed by a test I wrote), that shouldn't be an issue.  Because `Entry.log` is declared with a non-pointer receiver, the assignment to `entry.Time` within it does not affect later logging calls.

The only time a user will be stuck with an entry that always logs with the same timestamp is if they explicitly create one with `WithTime`, which seems like expected behaviour.

Here's the test I wrote.  I chose not to include it in the PR because it involves a `time.Sleep` call (and also isn't particularly well-polished).

```go
func TestTimeOverrideMultipleLogs(t *testing.T) {
	var buffer, buffer2 bytes.Buffer
	var fields Fields

	logger := New()

	logger.Out = &buffer
	logger.Formatter = new(JSONFormatter)

	ll := logger.WithField("herp", "derp")
	ll.Info("foo")

	err := json.Unmarshal(buffer.Bytes(), &fields)
	assert.Nil(t, err)

	firstTime := fields["time"]

	logger.Out = &buffer2

	time.Sleep(5 * time.Second)
	ll.Info("bar")

	err = json.Unmarshal(buffer2.Bytes(), &fields)
	assert.Nil(t, err)

	secondTime := fields["time"]

	assert.NotEqual(t, firstTime, secondTime)
}
```

dgsb:
@sbrisson2 indeed the non-pointer receiver prevents the issue I was talking about to occur. Anyway, I would like to have somehow the test included in your pr, otherwise any future change to the receiver type to a pointer would introduce a bug undetected by the test suite.

sbrisson2:
@dgsb That certainly makes sense.  I can think of a few ways to add that test.

1. Simply accept a 2-second sleep in the tests.
2. Add a (non-exported) `Entry.timeFunc` field of type `func() time.Time`, which would default to `time.Now()` for new entries.  I would use the field to provide a mock implementation of `Now()` in the test.
3. Use a custom formatter in the test to print the log timestamps at a higher resolution, and shorten the sleep in the test, say to 100ms.

I'll proceed with option 3, and if you prefer another solution, let me know.

I'll also add a comment in `Entry.log()` explaining the importance of the non-pointer receiver.

dgsb:
Hello @sbrisson2, thanks for your implication. Don't bother with a mock, I'm totally ok with the solution 3.

sbrisson2:
@dgsb The test has been added.

logrus.go
```diff
@@ -113,6 +114,7 @@ type FieldLogger interface {
 	WithField(key string, value interface{}) *Entry
 	WithFields(fields Fields) *Entry
 	WithError(err error) *Entry
+	WithTime(t time.Time) *Entry
```
dgsb: @sbrisson2 is this interface change really needed for your use case. I'm kind of worried it could cause in any of the thousands importing package. It we can remove this interface change, it is fine with me. If you really need this method in an interface, you could declare in another interface extending this one. Sorry for the delay.

sbrisson2: I don't need the interface change.  I can just add the methods in the `Logger` and `Entry` types.  I'll update the PR.

sbrisson2:
Thanks!

PR #782, opened 2018-07-09
export Lock and Unlock for some custom operation

bigpigeon:
some operation log rotate need lock logger


Derlafff:
Any progress on this / any reason why it's not being merged?

PR #783, opened 2018-07-10
add public Logger#Log & make Entry#Log public

lowzj:
In some scenarios, I need to write logs into different target files.
I implements it by using `Hook`, the current workaround like this:
```go
type ConsoleHook struct {
	logger *logrus.Logger
	levels []logrus.Level
}

func (ch *ConsoleHook) Fire(e *logrus.Entry) error {
	if ch.logger.Level >= e.Level {
		switch e.Level {
		case logrus.PanicLevel:
			ch.logger.Panic(e.Message)
		case logrus.FatalLevel:
			ch.logger.Fatal(e.Message)
		case logrus.ErrorLevel:
			ch.logger.Error(e.Message)
		case logrus.WarnLevel:
			ch.logger.Warn(e.Message)
		case logrus.InfoLevel:
			chrus.logger.Info(e.Message)
		case logrus.DebugLevel:
			ch.logger.Debug(e.Message)
		}
	}
	return nil
}
```

This is not convenient and low efficency. More condition statements, more invoking of `fmt.Sprint`. 

So I submit this pull request, and the implementation will be:
```go
func (ch *ConsoleHook) Fire(e *logrus.Entry) error {
	ch.logger.Log(e.Level, e.Message)
	return nil
}
```

If I want to use `WithFiled`, then I can just write code like this:
```go
if ch.logger.Level >= e.Level {
	ch.logger.WithField("key", "value").Log(e.Level, e.Message)
}
```

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -178,9 +178,9 @@ func (entry Entry) HasCaller() (has bool) {
 		entry.Caller != nil
 }
 
-// This function is not declared with a pointer value because otherwise
+// Log is not declared with a pointer value because otherwise
 // race conditions will occur when using multiple goroutines
-func (entry Entry) log(level Level, msg string) {
+func (entry Entry) Log(level Level, msg string) {
 	var buffer *bytes.Buffer
 
 	// Default to now, but allow users to override if they want.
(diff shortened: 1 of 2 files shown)


dgsb:
If this really needed ? You could just use a map with the level as a key and function as a value, the implementation would be as concise as needed without needing to change the public api. What do you think ?

lowzj:
Yes, it's needed. I think that the implementation by using `map` is as same as using `switch...case` in complexity.
Regardless of which solution is more concise and efficient, another reason is: if I call the `Logger#Fatal` method, the program will exit and cannot recover after calling `Hook#Fire` to make the caller to output logs.

lowzj:
This error is not caused by the changed code.
```
go: golang.org/x/crypto@v0.0.0-20180904163835-0709b304e793: unrecognized import path "golang.org/x/crypto" (https fetch: Get https://golang.org/x/crypto?go-get=1: dial tcp 74.125.69.141:443: i/o timeout)
```

lowzj:
this [commit](https://github.com/sirupsen/logrus/commit/a6668e7a605443af72ef75804a9f9f24a99f79ae) has implement the feature, so closed.

PR #786, opened 2018-07-12
Remove case insensitive import collision

Ch3ck:


Ch3ck:
@dgsb here is the correct fix. I thought I'd fixed that earlier

Ch3ck:
@dgsb I was getting an inconsistency error because of the typo in the `Sirupsen` package reference

dgsb:
@Ch3ck I suspect the error comes from your local installation of logrus source files. Which command did you use to retrieve the package ?

PR #787, opened 2018-07-13
Add CLICOLOR support

nolith:
This implement CLICOLOR and CLICOLOR_FORCE check on terminal coloring
as defined in https://bixense.com/clicolors/

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -3,6 +3,7 @@ package logrus
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 	"sync"
@@ -34,6 +35,9 @@ type TextFormatter struct {
 
 	// Force disabling colors.
 	DisableColors bool
+
+	// Override coloring based on CLICOLOR and CLICOLOR_FORCE. - https://bixense.com/clicolors/
+	OverrideColors bool
 
 	// Disable timestamp logging. useful when output is redirected to logging
 	// system that already adds timestamps.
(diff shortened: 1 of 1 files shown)


dgsb:
Hello @nolith, I'd rather not silently change the meaning of the isTerminal field which may be used for other things than just output coloring.
You can achieve what you're trying to do at the application level by setting the ForceColors field of the TextFormatter instance depending on the setting of the same environment variable.

nolith:
Hi @dgsb, will you be open to considering a rework of this PR that will not change the `isTerminal` meaning?

Let me explain why I think adding `CLICOLOR` support is helpful.

**The meaning of those env variables is giving control back to the final user**, instead of forcing each developer to provide a custom way to allow controlling colored output.

Right now we have ~4 projects using logrus, we will have to implement such thing either 4 times or as an external library, and in any case, we'll need to keep it updated.

Let me know what do you think.

dgsb:
I've reopened the PR. I you do some rework, you must not change the default behaviour. Too much users would be impacted. You can add a new API to the TextFormatter to complete the initialisation based on the content of the environment.

text_formatter.go
```diff
@@ -78,6 +82,22 @@ func (f *TextFormatter) init(entry *Entry) {
 	}
 }
 
+func (f *TextFormatter) isColored() bool {
+	isColored := f.ForceColors || f.isTerminal
+
+	if f.OverrideColors {
```
nolith: the new `OverrideColor` will enable the `CLICOLOR*` env overrides, this will not change the default behavior

text_formatter.go
```diff
@@ -78,6 +82,22 @@ func (f *TextFormatter) init(entry *Entry) {
...
+		if force, ok := os.LookupEnv("CLICOLOR_FORCE"); ok && force != "0" {
+			isColored = true
+		}
+
+		if os.Getenv("CLICOLOR") == "0" {
+			isColored = false
+		}
+	}
+
+	return isColored && !f.DisableColors
```
nolith: I decided to give maximum priority to `DisableColors` also when `OverrideColors == true` as this is the only way for a developer to completely disable colored output

dgsb:
@nolith it looks good to me, you may want to change the name OverrideColors to indicates that it uses the environment to select the colored mode or not. If you could also add some unit test for `isColored` method to cover the various combinatories, that would perfect. Thanks for your contribution.

PR #788, opened 2018-07-16
Add a redactor mechanism to prevent leaks of credentials and confiden…

JLugagne:
I think It would be nice to have a redactor mechanism to hide credentials and confidentials data to prevent any leaks or compromising data in logs.

Here is a proposition to do it, I'm not sure it fit very well the architecture (as it is at the end of the process on the serialized data) but seems to be working.

I know I could have done it by overriding the JSONFormatter and TextFormatter but as it should be a standard feature (IMHO) I wanted to have it as is.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -145,6 +145,7 @@ func (entry *Entry) write() {
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to obtain reader, %v\n", err)
 	} else {
+		serialized = entry.Logger.Redactor(serialized)
 		_, err = entry.Logger.Out.Write(serialized)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


dgsb:
We already the system of hooks for implementing modification of the content of a log entry. I don't think we need another registration for a function pointer to alter the content of a log.

dgsb:
@JLugagne just getting back to you. Did you succeed to implement your feature with hooks ?

JLugagne:
Thank you @dgsb for getting back, I didn't tried the hook yet but I did succeed by doing my RedactorFormatter based on JSONFormatter, it is working fine, I will try hooks later.

dgsb:
We already have the needed logic in logrus to implement this.

Issue #715, opened 2018-02-21
Modify/Remove Hook


sanbornm:
How do you modify or remove a hook during runtime?

tubocurarin:
Browsing the codebase I couldn't see something that allows to remove or even modify them. But I would like to see that feature too, especially good for proper test-hook cleanup. :+1:

phoenix147:
i'd also like to see this feature to remove hooks during runtime, see #701

robertomoutinho:
+1 for that idea ...
Would be awesome to be able to reload a hook (change its config) without having to restart the application.

phoenix147:
Since there is no easy way to add this without breaking the api, i ended up using a logger instance instead of the package level logger and if a restart is needed, i just recreate it and add needed hooks again.

betrok:
Related(somewhat) pull request: #782
With exposed lock it will be possible to safely modify the Hooks slice.

artheus:
I am writing some unit tests, and I found this method to be useful. Though not a very efficient one if you want to use it in production runtime..

```golang
type LogHook struct {
	enabled bool
	entries []*logrus.Entry
}

func (h *LogHook) Clear() {
	h.entries = []*logrus.Entry{}
}

func (h *LogHook) GetEntries() []*logrus.Entry {
	return h.entries
}

func (h *LogHook) Levels() []logrus.Level {
	return []logrus.Level{logrus.ErrorLevel}
}

func (h *LogHook) Fire(entry *logrus.Entry) error {
	if h.enabled == false {
		return nil
	}

	h.entries = append(h.entries, entry)
	return nil
}

// Make sure hook implements the logrus.Hook interface
var _ logrus.Hook = (*LogHook)(nil)

var logHook *LogHook

func enableLogHook() {
	if logHook == nil {
		logHook = &LogHook{enabled: true}
		logrus.AddHook(logHook)
	}

	logHook.Clear()
}

func getLogHookEntries() []*logrus.Entry {
	if logHook == nil {
		return nil
	}

	return logHook.GetEntries()
}

func disableLogHook() {
	if logHook == nil {
		return
	}

	logHook.enabled = false
}
```

Issue #701, opened 2018-02-05
Possibility to delete Hooks?


phoenix147:
Hi!

First, thank you for this awesome logging library.

I've implemented my own hook for transmitting logging information via websockets.
This hook, which forwards the log messages into a channel, is only added, if the web application connects to the backend.

I think it would be best to close the channel and delete the hook after the connection is closed, but it seems this isn't implemented. Is this not necessary?
I tried to implement my own RemoveHook method, but since Hooks is a slice, it's rather difficult.

jonmchan:
With the new ReplaceHooks syntax, here is how I went about removing a hook and putting it back in afterwards:

```
				newHooks := make(logrus.LevelHooks)
				for k,v := range logger.Hooks {
					for k2,v2 := range v {
						varType := fmt.Sprintf("%T", v2)

						if varType != "*YOURHOOKNAME" {
							newHooks[k] = append(newHooks[k], v2)
						} else {
							fmt.Printf("Excluding/Removing Hook - %v[%v]: %T\n", k, k2, v2)
						}
					}
				}

				originalHooks := logger.ReplaceHooks(newHooks)
				// Do operation without the hook
				logger.ReplaceHooks(originalHooks)
```

Figure the code might be useful for someone since it took me 15 minutes to come up with it... (or maybe I just am too rusty at golang :P)

PR #808, opened 2018-08-22
ReplaceHooks method for logger

betrok:
Safe way to modify hooks. Related issues:  #715, #701

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -336,3 +336,9 @@ func (logger *Logger) AddHook(hook Hook) {
 	defer logger.mu.Unlock()
 	logger.Hooks.Add(hook)
 }
+
+func (logger *Logger) ReplaceHooks(hooks LevelHooks) {
+	logger.mu.Lock()
+	logger.Hooks = hooks
+	logger.mu.Unlock()
+}
--- a/logger.go
+++ b/logger.go
@@ -337,8 +337,11 @@ func (logger *Logger) AddHook(hook Hook) {
 	logger.Hooks.Add(hook)
 }
 
-func (logger *Logger) ReplaceHooks(hooks LevelHooks) {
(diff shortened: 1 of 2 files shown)


logger.go
```diff
@@ -336,3 +336,9 @@ func (logger *Logger) AddHook(hook Hook) {
 	defer logger.mu.Unlock()
 	logger.Hooks.Add(hook)
 }
+
+func (logger *Logger) ReplaceHooks(hooks LevelHooks) {
```
dgsb: Can we have the former hooks returned from this method ? It could practical in order to restore the former hooks.

dgsb:
Hello, thanks for your contribution. Can you have a look at the slight modification I have requested ?

betrok:
Ok, i added the return value.

dgsb:
@betrok thanks for your contribution

XieWeiXie:
When dealing with a large volume of logs, the ReplaceHook may cause CPU spikes.
image

betrok:
It has nothing to do with this pool request, check what your hooks do.

XieWeiXie:
> It has nothing to do with this pool request, check what your hooks do.

what should i do?

i want to change some filed in logger message , \r \n \t and so on。

this is my example hook。

type ReplaceHook struct {
	Old []string
	New string
}

func (hook *ReplaceHook) Fire(entry *logrus.Entry) error {
	var newMessage = entry.Message
	for _, i := range hook.Old {
		newMessage = strings.ReplaceAll(newMessage, i, hook.New)
	}
	entry.Message = newMessage
	return nil
}

func (hook *ReplaceHook) Levels() []logrus.Level {
	return logrus.AllLevels
}


it worked , but it caused cpu higher than no hook。

PR #810, opened 2018-08-28
Added trace log level.

### LINKED ISSUE #663: Trace log level
Hello

I am aware that this has been discussed before. It was also mentioned that Trace was not a common  requirement and this an edge case, further, it would confuse some people. I disagree that this is valid.

My use case is that I am working with a feed of financial data that if logged with Debug would be very verbose.

I would like to log with Trace and only enable Trace if I have a specific problem to find, for everything else, Debug would be fine.

Can we just add a Trace level. For those who don't understand it, just don't need to log to it :)

mkorolyov:
added trace level that below debug one.  fixes #663

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,9 +246,10 @@ A list of currently known of service hook can be found in this wiki [page](https
 
 #### Level logging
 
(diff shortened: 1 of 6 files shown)


dgsb:
duplicate of #844

Issue #814, opened 2018-09-01
Add module definition for logrus


dgsb:
go 1.11 has introduced the new concept of module used to ease the package versioning and dependency management: https://golang.org/doc/go1.11#modules
To ease our users in integrating logrus into their project, we should provide a module definition for logrus.

PR #817, opened 2018-09-06
Add go module support

dgodd:
Travis will only run the "go get" install lines if using go 1.10.x


dgsb:
@dgodd thanks for your contribution. Could you resync your branch with master. There are a lot of dependencies which I do not expect in the `go.mod`. I suspect they are all coming indirectly from airbrake that I've just removed in master,

dgsb:
Fixes #814

dgodd:
Thanks, @dgsb

PR #819, opened 2018-09-08
Fix travis on go 1.11 with modules

dgsb:


dgsb:
These commits have been merge-squashed on master

PR #823, opened 2018-09-17
Fix AppEngine builds

mdittmer:
`go build -tags appengine [...]` currently yields:

```
.../github.com/sirupsen/logrus/text_formatter.go:84:4: undefined: initTerminal
```

This change implements `initTerminal` for AppEngine builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,3 +26,26 @@ matrix:
         - go get golang.org/x/sys/windows
       script:
         - go test -race -v ./...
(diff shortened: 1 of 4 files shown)


mdittmer:
@dgsb PTAL. The test failures make me wonder whether a different `initTerminal()` implementation should be provided when `-tags appengine`?

mdittmer:
@dgsb done!

Issue #824, opened 2018-09-18
Building for Windows fails on MacOS and Linux without additional go get


a-h:
With the latest version of logrus, my cross-platform (Linux to Windows) build has stopped working.

Steps to reproduce:

* Write some software which uses logrus.
* Compile on MacOS and Linux with `go get ./...` then `go build` - works fine.
* Try `GOOS=windows go build`

Fails with `cannot find package "github.com/konsorten/go-windows-terminal-sequences"`.

Steps to workaround:

* Run `GOOS=windows go get ./...` then re-run the build to get Go to include the required Windows packages.

Possible improvements:

* Add `import _ "github.com/konsorten/go-windows-terminal-sequences"` in any file just to bring in the required package. (Doesn't work: `build constraints exclude all Go files in /Users/adrian/go/src/github.com/konsorten/go-windows-terminal-sequences`)
* Vendor the package until Go 1.12 when modules can replace it.

a-h:
This was changed in https://github.com/sirupsen/logrus/pull/822

dgsb:
Hello @a-h , that is unfortunate that this change broke your build. But I don't think you can expect that a go get ./... can retrieve all needed dependencies for all architectures. I think you workaround is exactly what should be done. What would you like we do to fix that ?

dgsb:
konsorten/go-windows-terminal-sequences#1

dgsb:
@a-h can you try if the build works for you in module mode ?

a-h:
OK, so to test this, I did the following. Installed Go 1.11 on a Mac and started a new project in a directory under the desktop. I deleted the `~/go/pkg/mod` directory to ensure that no modules were cached.

I created a `go.mod` file:

```go
module github.com/a-h/logtest

require github.com/sirupsen/logrus v1.0.6
```

Then created a simple "Hello World" main.go:

```go
package main

import "github.com/sirupsen/logrus"

func main() {
	logrus.Println("Hello")
}
```

Next, I ran `GOOS=windows go build` and the Go system downloaded required modules:

```
$ GOOS=windows go build
go: finding github.com/sirupsen/logrus v1.0.6
go: finding golang.org/x/crypto v0.0.0-20180910181607-0e37d006457b
go: finding golang.org/x/sys v0.0.0-20180925112736-b09afc3d579e
go: downloading github.com/sirupsen/logrus v1.0.6
go: downloading golang.org/x/crypto v0.0.0-20180910181607-0e37d006457b
[...]

PR #825, opened 2018-09-19
Add missing module dependency for windows build

dgsb:


dgsb:
When modules are available, fixes #824

Issue #642, opened 2017-09-29
JSONFormatter isn't handling functions as Field values


xandrewrampulla:
When using the JSONFormatter, if I accidentally pass a function as a Field value, the entire log message is lost.  It would be better if only that one field was lost (even better would be to log that the field couldn't be translated into JSON properly).

Simple example code
```go
package main

import (
	log "github.com/sirupsen/logrus"
)

func myfun() string {
	return ""
}

func main() {

	log.SetFormatter(&log.JSONFormatter{})

	log.WithFields(log.Fields{
		"animal": "walrus",
		"myfunc": myfun,
	}).Info("A walrus appears")
}
```

Current output
```Failed to obtain reader, Failed to marshal fields to JSON, json: unsupported type: func() string```

**Comment by dgsb** [comment]
Indeed this is annoying

**Comment by dimatkach** [comment]
I am still seeing this issue. 

```
      import (
         log "github.com/sirupsen/logrus"
      )
      type Foo struct {
         foo func() string
      }


     func main() {
        foo :=  Foo {
          func () string { return "foo" }
         }
   
        log.SetFormatter(&log.JSONFormatter{})
        log.WithField("foo", foo).Info("Foo!")
      }
```

It looks like the PR referenced above only fixes the case where field _itself_ is a func, but if it is a struct that has a func field somewhere inside of it, there is still a problem. The solution either needs to be more involved, and go recursively inside each of the struct.

linnaname:
I am still seeing this issue  in v1.7.0

Test Code:
`log.SetFormatter(&log.JSONFormatter{})`
 `req,_ := http.NewRequest("POST", "127.0.0.1",nil)`
`log.WithFields(log.Fields{
		"req":    req,
	}).Error("new request failed")`

Output:`Failed to obtain reader, failed to marshal fields to JSON, json: unsupported type: func() (io.ReadCloser, error)`

oxzi:
Today I came across the same error for a struct which has a function as a field type. As a workaround, a tag can be added to the field, so that this is not taken into account for the JSON,  `json:"-"`.

vishalkanaujia:
If you are logging HTTP request/response, use `httputil.DumpRequest`/ `httputil.DumpResponse`

PR #835, opened 2018-10-04
Move func reflect check to json formatter

zachbadgett:
Leave func handling up to the formatter.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"fmt"
 	"os"
-	"reflect"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


zachbadgett:
In regards to #832 and #642; no need to clutter our entry with more fields. Let the formatter decide how the field types are handled and replace invalid values with a message.

dgsb:
Hello @zachbadgett, thanks for your contribution, but I'm not sure if it's better to handle that at the formatter level rather than the Entry level.
We have another set of issues/pr (#365,#151,#207) related to fields for which it makes more sense to implement them at Entry level. I'd rather rationalize all this future work at entry level

zachbadgett:
IMO, which types to support would best to be left to the formatter. If a user wants to implement a custom formatter and wants to support grabbing details from a custom function in their application then they can.

Either way, I believe you should either move all normalization to the entry (which includes the error to string) or keep it all on the formatter.

dgsb:
IMO It's kind of different, an error can be stringified. The fact that the json formatter can't is an implementation detail of the json formatter.
While I'm not sure it does not make sense to stringify a function even if the TextFormatter can do it.

zachbadgett:
Kind of a silly example if someone really wanted to get information on a func: https://play.golang.org/p/zJiQT7AF-ni
I do understand your point about if a func can really be "stringified", the thing I was trying to get at is an entry should be type agnostic, let the formatters worry about which type to support.

Issue #813, opened 2018-08-30
Allow testing of calls to `Fatal*`


MrJoy:
If you were to add an `ExitFunc func(int)` field to the `Logger` struct, with a default implementation of `func(int code) { os.Exit(code) }`, it would become much easier to test code that calls `Fatal*`, by simply substituting the implementation with `func(int code) { panic(...) }` -- without having to resort to sub-process shenanigans that are complicated to implement and don't play well with `-cover`.  Is this something you would be willing to implement / accept a PR for?

dgsb:
That would be indeed very useful, such a PR would be very welcome.

caalberts:
I'd like to give this a try

dgsb:
Hi @caalberts, your contributions will be welcome.

caalberts:
I raised a PR on this, do let me know if there is any concern.

PR #837, opened 2018-10-06
Implements #813 - Add option to panic in `test.NewNullLogger`

caalberts:
Add a field in `Logger` for `Exit` implementation which defaults to `Exit()`.
This can be used in `NullLogger()` to replace calls to `os.Exit()` with `panic()`

### DIFF
--- a/hooks/test/test.go
+++ b/hooks/test/test.go
@@ -39,23 +39,11 @@ func NewLocal(logger *logrus.Logger) *Hook {
 
 }
 
-type TestOption func(logger *logrus.Logger)
-
-func FatalPanics(logger *logrus.Logger) {
-	logger.Exit = func(code int) {
-		panic(code)
-	}
-}
-
 // NewNullLogger creates a discarding logger and installs the test hook.
-func NewNullLogger(options ...TestOption) (*logrus.Logger, *Hook) {
+func NewNullLogger() (*logrus.Logger, *Hook) {
 
 	logger := logrus.New()
 	logger.Out = ioutil.Discard
-
-	for _, option := range options {
-		option(logger)
-	}
 
 	return logger, NewLocal(logger)
 
--- a/hooks/test/test.go
+++ b/hooks/test/test.go
@@ -39,11 +39,23 @@ func NewLocal(logger *logrus.Logger) *Hook {
 
 }
 
+type TestOption func(logger *logrus.Logger)
+
+func FatalPanics(logger *logrus.Logger) {
+	logger.Exit = func(code int) {
+		panic(code)
+	}
+}
+
 // NewNullLogger creates a discarding logger and installs the test hook.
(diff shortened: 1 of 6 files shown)


dgsb:
Hi @caalberts, thanks for your contribution. There is something important which needs to be changed in order to be backward compatible if we want to merge this PR in a v1 release.
With this implementation, the static declaration of a logger like
```go
log = &Logger{
      Out: os.Stderr,
      Formatter: new(JSONFormatter),
      Hooks: make(LevelHooks),
      Level: logrus.DebugLevel,
}
```
will be broken and execute a null pointer upon a Fatal call.
We could add instead a method Exit on Logger object which will then check the value of the Exit field and call appropriatly the default method if nil.

hooks/test/test.go
```diff
@@ -39,12 +39,24 @@ func NewLocal(logger *logrus.Logger) *Hook {
...
+
+func FatalPanics(logger *logrus.Logger) {
+	logger.Exit = func(code int) {
+		panic(code)
+	}
+}
+
 // NewNullLogger creates a discarding logger and installs the test hook.
-func NewNullLogger() (*logrus.Logger, *Hook) {
+func NewNullLogger(options ...TestOption) (*logrus.Logger, *Hook) {
```
dgsb: I'd rather not change this interface by adding a new parameter. We can just patch the resulting logger when needed by calling the same function on the returned object

hooks/test/test_test.go
```diff
@@ -71,3 +71,14 @@ func TestLoggingWithHooksRace(t *testing.T) {
 	entries := hook.AllEntries()
 	assert.Equal(100, len(entries))
 }
+
+func TestFatalWithPanic(t *testing.T) {
+	assert := assert.New(t)
+
+	logger, hook := NewNullLogger(FatalPanics)
```
dgsb: We can just call FatalPanics on the returned logger object rather than implementing that with a callback.

logger.go
```diff
@@ -32,6 +32,8 @@ type Logger struct {
 	mu MutexWrap
 	// Reusable empty entry
 	entryPool sync.Pool
+	// Function to exit the application, defaults to `Exit()`
+	Exit exitFunc
```
dgsb: The symbol Exit should rather be a method of the logger object to be backward compatible. I think users may also want to have their registered handlers to be called also, therefore we want to istead to make configurable the os.Exit call from logrus.Exit

caalberts:
Yeah I see what you mean. I can see how having a `Exit` method on the logger object will keep backward compatibility.

To keep the handlers executed, this `Exit` method will execute handlers before the final exit, be it `os.exit()` or `panic()`.

We still need to keep `logrus.Exit` function around though, because it is exported.

caalberts:
@dgsb Let me know if these changes are ok

dgsb:
@caalberts thanks for your contribution

PR #839, opened 2018-10-11
Prevent TextFormatter from modifying entry.Data

egurnov:
Current implementation of TextFormatter modifies entry.Fields. If a handler runs after the call to Format, it will see the prefixed fields created by prefixFieldClashes.

The proposed implementation creates a shallow copy of entry.Data. A similar approach is already used in JSONFormatter.

Example:
```go
package main

import (
	"fmt"
	"sync"
	"time"

	"github.com/sirupsen/logrus"
)

func main() {
	l := logrus.New()
	h := &Hook{}
	l.AddHook(h)

	l.WithFields(logrus.Fields{
		"hello": "world",
		"level": "highest",
	}).Info("hello")

	h.Wg.Wait()
}

type Hook struct {
	Wg sync.WaitGroup
}

func (h *Hook) Levels() []logrus.Level {
	return logrus.AllLevels
}

func (h *Hook) Fire(entry *logrus.Entry) error {
	h.Wg.Add(1)
	go func() {
		defer h.Wg.Done()
		time.Sleep(time.Second)
		fmt.Printf("Inside hook: %+v\n", entry.Data)
	}()
	return nil
}
```

Output with current implementation:
```
time="2018-10-11T10:49:44+02:00" level=info msg=hello fields.level=highest hello=world
Inside hook: map[fields.level:highest hello:world]
```

Output with proposed implementation:
```
time="2018-10-11T10:51:36+02:00" level=info msg=hello fields.level=highest hello=world
Inside hook: map[level:highest hello:world]
```

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -107,14 +107,18 @@ func (f *TextFormatter) isColored() bool {
 
 // Format renders a single log entry
 func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
-	prefixFieldClashes(entry.Data, f.FieldMap)
+	data := make(Fields, len(entry.Data))
+	for k, v := range entry.Data {
+		data[k] = v
+	}
+	prefixFieldClashes(data, f.FieldMap)
 
-	keys := make([]string, 0, len(entry.Data))
-	for k := range entry.Data {
+	keys := make([]string, 0, len(data))
+	for k := range data {
 		keys = append(keys, k)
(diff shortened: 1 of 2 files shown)


egurnov:
Closing because this has been fixed by a surprisingly similar commit https://github.com/sirupsen/logrus/commit/2cafb78db2ffe7e4592a07991f77f9105db898ab

PR #845, opened 2018-10-19
Make logrus.Level implement encoding.TextUnmarshaler

yanana:
Implemented `encoding.TextUnmarshaler` for `logrus.Level` to utilize the type in libraries that accept `encoding.TextUnmarshaler`.

---
This change is [Reviewable](https://reviewable.io/reviews/sirupsen/logrus/845)

### DIFF
--- a/logrus_test.go
+++ b/logrus_test.go
@@ -372,6 +372,19 @@ func TestParseLevel(t *testing.T) {
 	assert.Equal(t, "not a valid logrus Level: \"invalid\"", err.Error())
 }
 
+func TestUnmarshalText(t *testing.T) {
+	var u Level
+	for _, level := range AllLevels {
+		t.Run(level.String(), func(t *testing.T) {
+			assert.NoError(t, u.UnmarshalText([]byte(level.String())))
+			assert.Equal(t, level, u)
+		})
+	}
+	t.Run("invalid", func(t *testing.T) {
+		assert.Error(t, u.UnmarshalText([]byte("invalid")))
+	})
+}
+
 func TestGetSetLevelRace(t *testing.T) {
 	wg := sync.WaitGroup{}
 	for i := 0; i < 100; i++ {
--- a/logrus.go
+++ b/logrus.go
@@ -53,6 +53,18 @@ func ParseLevel(lvl string) (Level, error) {
 	return l, fmt.Errorf("not a valid logrus Level: %q", lvl)
 }
 
+// UnmarshalText implements encoding.TextUnmarshaler.
+func (level *Level) UnmarshalText(text []byte) error {
+	l, err := ParseLevel(string(text))
+	if err != nil {
+		return err
+	}
+
+	*level = Level(l)
+
+	return nil
+}
+
 // A constant exposing all logging levels
 var AllLevels = []Level{
 	PanicLevel,


dgsb:
Hello @yanana, thanks for your contribution. Do you mind to elaborate a little on what would be the use case ?
Adding a new API adds a maintenance cost, so we need to assess its usefulness.

yanana:
Hello @dgsb, thanks for your quick response. Especially, I'd like to utilize the API for use with https://github.com/kelseyhightower/envconfig. The library can convert environment variables to types implementing `encoding.TextUnmarshaler` or `encoding.BinaryUnmarshaler`.

logrus_test.go
```diff
@@ -372,6 +373,21 @@ func TestParseLevel(t *testing.T) {
 	assert.Equal(t, "not a valid logrus Level: \"invalid\"", err.Error())
 }
 
+func TestUnmarshalText(t *testing.T) {
+	u, _ := (interface{}(&(AllLevels[0]))).(encoding.TextUnmarshaler)
+	for _, level := range AllLevels {
+		t.Run(level.String(), func(t *testing.T) {
+			err := u.UnmarshalText([]byte(level.String()))
+			assert.Nil(t, err)
```
dgsb: You could use the NoError method to check that.

logrus_test.go
```diff
@@ -372,6 +373,21 @@ func TestParseLevel(t *testing.T) {
...
+			assert.Equal(t, &level, u)
+		})
+	}
+	t.Run("invalid", func(t *testing.T) {
+		err := u.UnmarshalText([]byte("invalid"))
+		assert.NotNil(t, err)
```
dgsb: You could use the Error method to check that.

logrus_test.go
```diff
@@ -372,6 +373,21 @@ func TestParseLevel(t *testing.T) {
 	assert.Equal(t, "not a valid logrus Level: \"invalid\"", err.Error())
 }
 
+func TestUnmarshalText(t *testing.T) {
+	u, _ := (interface{}(&(AllLevels[0]))).(encoding.TextUnmarshaler)
```
dgsb: You can just declare a Level variable here

logrus_test.go
```diff
@@ -372,6 +373,21 @@ func TestParseLevel(t *testing.T) {
...
+	for _, level := range AllLevels {
+		t.Run(level.String(), func(t *testing.T) {
+			err := u.UnmarshalText([]byte(level.String()))
+			assert.Nil(t, err)
+			assert.Equal(t, &level, u)
```
dgsb: I think you should test the value itself rather than comparing pointers here.

dgsb:
Hello @yanana, the need seems legit to me. However I would like some change in the associated unit test before merging it. Can you have a look at it ?

logrus_test.go
yanana: Would you tell me what the “Level variable” means? Suggesting to stop using `AllLevels`?

yanana:
Thanks for your dedication, @gdsb. I'm going to fix my code.

logrus_test.go
dgsb: Just something like:
```
var u Level
```

yanana:
Hello @dgsb, I've fixed all you pointed out and squashed commits. Would you please confirm again?

yanana:
Hello @dgsb, I've fixed all you pointed out and squashed commits. Would you please confirm again?

yanana:
Hello @dgsb, I've fixed all you pointed out and squashed commits. Would you please confirm again?

yanana:
Hello @dgsb, I've fixed all you pointed out and squashed commits. Would you please confirm again?

yanana:
Hello @dgsb, I've fixed all you pointed out and squashed commits. Would you please confirm again?

yanana:
Hello @dgsb, I've fixed all you pointed out and squashed commits. Would you please confirm again?

yanana:
Sorry for annoying consecutive comments due to GitHub's partial down. I've deleted duplicated comments.

PR #847, opened 2018-10-22
WIP: PLEASE LEAVE OPEN for now. Possible bugs? Refactoring highlighted a couple odd behaviors.

loren-osborn:
This isn't the PR I had mentioned wanting to submit a few days ago. While I was working on that, however, I discovered these two odd behaviors that are likely unintended bugs. Not knowing how it's actually intended to work, I updated the documentation, made the odd behaviour much more obvious, and marked them with `@todo` comments.

Could you look at these and tell me if there's a good reason for them I was unaware of?

Just FYI, the last two commits in my last PR are at a different email than this, as those weren't an email address github recognizes as belonging to me. Sorry for any confusion.


loren-osborn:
@dgsb, @stevvooe could one of you look at this and let me know if either/both of these are indeed bugs? (maintained current behavior for now.)

dgsb:
@loren-osborn this is indeed a bug, we must not blank the error field.
However, before you go too far in your refactoring, I'd rather keep this part simple because this is a corner case which should occur frequently (if not at all).
The heart of the fix was to not drop a trace when the user was a function as a value of a field.
The trace here is just informative for the user to be able to notice the error and to fix it.
We need to fix the bug which drop this trace upon Entry copy, but we don't need to be too fancy there.

loren-osborn:
I'm open to whatever fix you feel is most appropriate, My thinking was based on recent reading about issues stringifying errors too early. (Please see the errors library I linked to just above the definition of `errorSlice`.) The errors are only being *used* to move strings around, but as they are errors, I thought it important to treat them as such. 

Please advise.

loren-osborn:
This PR is replaced by #851. I will close it soon, but I'm currently using it for CI travis testing of another PR that I'm planning, so please **leave it open** for now.

dgsb:
@loren-osborn can we close this pr ?

PR #848, opened 2018-10-22
Add a DeferExitHandler function

geodimm:
Useful for running exit handlers in the same order as defer statements

### DIFF
--- a/alt_exit_test.go
+++ b/alt_exit_test.go
@@ -12,9 +12,6 @@ import (
 	"time"
 )
 
-func f1() {}
-func f2() {}
-
 func TestRegister(t *testing.T) {
 	current := len(handlers)
 


georgijd:
Hi,

Is there any chance you guys could have a look at this one? We've been using it for a while now and it's proven very useful.

Thanks

richpoirier:
@georgijd Can you describe the use case for controlling the order in which the exit handlers run? And why you couldn't just reorder your calls to `RegisterExitHandler()`?

georgijd:
Hi,

Thank you for your response! A use-case would be if a library which utilises logrus already registers one or more exit handlers. Users of that library might need to run those exit handlers after their own exit handlers which would not be achievable if it's only possible to append exit handlers.  I understand that in the majority of cases the exit handlers can be registered in the correct order, but sometimes that's just not possible. Regardless of whether the argument is compelling enough or not, it's always nice to allow people to control the order of such things. Especially when it comes to setup/cleanup procedures and there is little to no cost.

dgsb:
@richpoirier this is a common and expected behaviour similar to what is done by atexit. The current implementation is kind of unexpected.

richpoirier:
Ok. @georgijd could you just update the functions' docs so that it's clear that `Register` adds the handler to the end of the handlers list and `Defer` adds it to the beginning? And the unit tests should verify that behavior as well.

Thanks for the PR btw 👍

georgijd:
Hi,

Sorry for the delay. I've been quite busy lately. I've updated the diff with tests and new docstrings

PR #849, opened 2018-10-24
Add GELF to third party formatters

FabienM:
I had to create a [GELF](http://docs.graylog.org/en/2.4/pages/gelf.html) implementation for logrus a while ago.

I noticed there is no suggestion for gelf logging in official documentation, so here is my proposal.

Features:
* Logrus level to syslog level mapping
* Logrus entry times are converted to UNIX timestamps
* Logrus entry fields are prefixed with `_`, excepting standardized GELF fields (`version`, `host`, `short_message`, `full_message`, `timestamp` and `level`)


dgsb:
Hi @FabienM, nice work.

PR #851, opened 2018-10-25
WIP: Fix for field error handling bug

loren-osborn:
Initial commit just adds test. Want to verify works correctly with CI/CD.


loren-osborn:
@dgsb, this should be good to go... I intentionally kept it simple. Let me know if you'd like any changes.

loren-osborn:
@dgsb, @sirupsen, @stevvooe, want to let me know if this is what you wanted. It just adds a test for the newly-found bug, and then adds a fix for it.

The fix is simply string manipulation.

(FYI, my previous refactor of the big had been inaccurate... The old behavior was that if there was a new error, it appended **only the last** new error to the previous error list, otherwise, it emptied the error list.)

dgsb:
Hello @loren-osborn, thanks for your contribution. The fix looks correct to me. However I find the test case too complicated to be easily maintainable in the future. May I request a few changes on it ? Please look at the review.

entry_test.go
```diff
@@ -34,6 +38,9 @@ func TestEntryWithError(t *testing.T) {
 
 func TestEntryPanicln(t *testing.T) {
 	errBoom := fmt.Errorf("boom time")
+	logger := New()
+	logger.Formatter = &TextFormatter{DisableTimestamp: true, DisableColors: true}
```
dgsb: I would advise to use a json formatter, it is way easier to check a json output.

entry_test.go
```diff
@@ -43,17 +50,234 @@ func TestEntryPanicln(t *testing.T) {
 		case *Entry:
 			assert.Equal(t, "kaboom", pVal.Message)
 			assert.Equal(t, errBoom, pVal.Data["err"])
+			assert.Regexp(t, `^level=panic msg=kaboom err="boom time"`, outBuffer.String())
```
dgsb: The key order depends on the implementation and is not guaranteed, this is why a json formatter if way more convenient to check the output of a log.

entry_test.go
```diff
@@ -43,17 +50,234 @@ func TestEntryPanicln(t *testing.T) {
...
+	// Regular Expressions used later
+	keyValPartsPattern := regexp.MustCompile(`([^\s"=]*|"(?:[^\\"]|\\.)*")=([^\s"=]*|"(?:[^\\"]|\\.)*")(?:\s|$)`)
+	commaSepPattern := regexp.MustCompile(", ")
+	quotedCommaSepPattern := regexp.MustCompile(`", "`)
+	// This allows us to have multiple distinct tests, each with a new entry.
+	// While it is really a good idea to break tests out into separate
+	// independent units, we covered all the obvious cases with one multi-part
+	// test. (Multi-part tests share an entry and test the entry state after
+	// each change.)
+	for _, independantTestCase := range [][]struct {
```
dgsb: If there is a single element in the array, we can just remove a slice and a loop level, it will make the unit test easier to maintain.

entry_test.go
```diff
@@ -43,17 +50,234 @@ func TestEntryPanicln(t *testing.T) {
...
+	logger.Out = &outBuffer
 	entry := NewEntry(logger)
 	entry.WithField("err", errBoom).Panicln("kaboom")
 }
 
+// TestEntryFormattingError tests proper handling of any field formatting errors
+func TestEntryFormattingError(t *testing.T) {
+	logger := New()
+	// Turning off timestamp and color marks parsing the result easier:
+	logger.Formatter = &TextFormatter{
```
dgsb: We can use a json formatter to make the test easier to implement and to maintain.

entry_test.go
```diff
@@ -43,17 +50,234 @@ func TestEntryPanicln(t *testing.T) {
...
+		DisableTimestamp: true,
+		DisableColors:    true,
+	}
+	// Regular Expressions used later
+	keyValPartsPattern := regexp.MustCompile(`([^\s"=]*|"(?:[^\\"]|\\.)*")=([^\s"=]*|"(?:[^\\"]|\\.)*")(?:\s|$)`)
```
dgsb: with a json formatter we won't need a complicated regexp to parse the ouput of a trace

entry_test.go
```diff
@@ -43,17 +50,234 @@ func TestEntryPanicln(t *testing.T) {
...
+			entry = entry.WithFields(fieldGroupTestPart.Group)
+			// Everything below here is analizing the Entry state after
+			// WithFields() is called:
+
+			// Capture logged output:
+			entry.Info("baz")
+			finalFieldsStr = outBuffer.String()
+
+			// Parse the output into key-value pairs.
+			keyValPairs := keyValPartsPattern.FindAllStringSubmatch(outBuffer.String(), -1)
```
dgsb: a json formatter will make the test less brittle here

entry_test.go
loren-osborn: This is a **really** good suggestion I hadn't thought of.

entry_test.go
loren-osborn: Agreed

entry_test.go
loren-osborn: Looks very much the case... I'm a bit tangled up in another branch at the moment, but I **hope** to have something you can look at within the hour

entry_test.go
loren-osborn: Actually I put this loop here intentionally, to make independant test cases easier to add. What is there now is a single independant test, with multiple phases. This was due to my laziness. I’d rather break this into multiple independent tests then remove this outer loop. Let me see what I can do in that regard.

loren-osborn:
I am working on an additional commit to add more granular independent tests, rather than removing the outer loop.

entry_test.go
loren-osborn: Added more independant test cases to the outer loop. It no longer has just one item.

loren-osborn:
@dgsb, I believe I addressed all of your concerns with my test code. Please let me know if you would like me to change anything else.

dgsb:
@loren-osborn the fix is correct, but the test you've added is way too complicated to be usable when a modification on the source code cause its failure.
We really need to have simple (even simplistic) unit test so that we can easily identify what needs to be fixed when they are broken.

loren-osborn:
I’m sorry to learn my tests are so fragile. I will work on making simple, clear, concise tests for this

loren-osborn:
@dgsb, I don’t think I was able to simplify this very much, but think I made it a bit more clear. I also wasn’t able to find the fragility you alluded to. If you can tell me how the tests are failing, i’m happy to make them more robust.

loren-osborn:
Just a note, the three failing commits are **supposed to** be failing, since the fix isn’t applied until the final commit.

loren-osborn:
@dgsb, I took two more shots at clarifying, simplifying and fixing any fragility.

- I think it is definitely more clear than before.
- I did simplify the test some, by removing tests that are impossible to fail. but I also abstracted out some stuff, so I'm unsure if you feel it's an improvement.
- I'm still unsure how these tests are fragile. Please point me to something that breaks them, and I'm happy to improve them.

I'm open to any suggestions, but at this point I feel a bit like I'm going in circles. I'm happy with the state of the PR now, but please offer more specific feedback if you think it needs more work.

Thank you,
@loren-osborn

loren-osborn:
I just noticed I left `SlowTestWriter` in this set of commits... it’s part of other pending changes i’m working on... Please review everything else, and I will have `SlowTestWriter` removed from this test series shortly.

loren-osborn:
Rebased this on `master`.

@dgsb, I would really appreciate any suggestions on how to make this suitable for merging. I made the tests as clear and simple as I was able, and am unaware of anything making them fragile. Please point any specifics you'd like me to change, and I'm happy to.

entry_test.go
loren-osborn: While I did, as you suggested, switch to the json formatter in my new test(s), this was an existing test from 4 years ago. (My minor change, that I'm happy to revert, simply alternated the order of operations setting up the test. I believe I did this for readability.)

dgsb:
@loren-osborn the fix itself is something like 10 lines, and all the tests around are something like 400 hundreds lines. I think it is too wide and not needed for this fix. Can you write the tests with something like 50 lines ?
We need somehow to new test functions, one which ensures the error field is kept upon a call to WithTime and one which checks the error field is kept after another WithField with correct parameter is called.
I think that's really all we need here.

Issue #468, opened 2017-01-18
Concurrent map error on text formatter caused by prefixFieldClashes


Kegsay:
Version: `master/a283a10442df8dc09befd873fab202bf8a253d6a`

Stack:
```
fatal error: concurrent map read and map write

goroutine 5168081 [running]:
runtime.throw(0xabc559, 0x21)
        /usr/local/go/src/runtime/panic.go:566 +0x95 fp=0xc43134f2d8 sp=0xc43134f2b8
runtime.mapaccess1_faststr(0xa0c9c0, 0xc43116c510, 0xaa4f7c, 0x4, 0x9b2020)
        /usr/local/go/src/runtime/hashmap_fast.go:201 +0x4f3 fp=0xc43134f338 sp=0xc43134f2d8
github.com/Sirupsen/logrus.(*TextFormatter).Format(0xc420014280, 0xc42226ff00, 0xc42226ff00, 0xc43134f530, 0x5112db, 0xc42016c660, 0xc42226ff00)
        /home/goneb/go-neb/vendor/src/github.com/Sirupsen/logrus/text_formatter.go:91 +0x47e fp=0xc43134f4c8 sp=0xc43134f338
github.com/Sirupsen/logrus.(*Entry).Reader(0xc42226ff00, 0xc42ecda704, 0xc42226ff00, 0x0)
        /home/goneb/go-neb/vendor/src/github.com/Sirupsen/logrus/entry.go:44 +0x40 fp=0xc43134f540 sp=0xc43134f4c8
github.com/Sirupsen/logrus.Entry.log(0xc4200181c0, 0xc43116c510, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x4, 0xc42915ef00, ...)
        /home/goneb/go-neb/vendor/src/github.com/Sirupsen/logrus/entry.go:94 +0x200 fp=0xc43134f600 sp=0xc43134f540
github.com/Sirupsen/logrus.(*Entry).Info(0xc42226fec0, 0xc43134f7a8, 0x1, 0x1)
        /home/goneb/go-neb/vendor/src/github.com/Sirupsen/logrus/entry.go:129 +0x9a fp=0xc43134f668 sp=0xc43134f600
github.com/Sirupsen/logrus.(*Entry).Print(0xc42226fec0, 0xc43134f7a8, 0x1, 0x1)
[...]

PR #857, opened 2018-11-05
fix race condition caused by writing to entry.Data

h12w:
This PR should be able to fix https://github.com/sirupsen/logrus/issues/468, and it uses the same technique as JSONFormatter.

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -163,7 +163,7 @@ func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
 		timestampFormat = defaultTimestampFormat
 	}
 	if f.isColored() {
-		f.printColored(b, entry, keys, nil, timestampFormat)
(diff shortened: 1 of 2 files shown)


text_formatter.go
```diff
@@ -160,7 +163,7 @@ func (f *TextFormatter) Format(entry *Entry) ([]byte, error) {
 		timestampFormat = defaultTimestampFormat
 	}
 	if f.isColored() {
-		f.printColored(b, entry, keys, timestampFormat)
+		f.printColored(b, entry, keys, nil, timestampFormat)
```
dgsb: Should we give the new data array as parameter here ?

dgsb: Should we give the new data array as parameter here ?

h12w: Fixed.

h12w: Fixed.

PR #859, opened 2018-11-09
Enforce the new fully-lowercase import path with Go 1.4 mecanism

dolmen:
The original import path of was github.com/Sirupsen/logrus but it has
changed to a fully lowercase path in July 2017.
References:
- https://github.com/sirupsen/logrus/blob/7eeb7b7cbdeb0a582a5b7a1512d91f956f1674aa/README.md#case-sensitivity
- https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276

To help dependent project to detect the issue with mixing old and new
import path, let's use the Go 1.4 mecanism that enforces the import
path: https://golang.org/doc/go1.4#canonicalimports

This is a complement to go.mod for pre-Go-modules Go versions.

### DIFF
--- a/logrus.go
+++ b/logrus.go
@@ -1,4 +1,4 @@
-package logrus
+package logrus // import "github.com/sirupsen/logrus"
 
 import (
 	"fmt"


dgsb:
We can not merge that in v1 branch, it may break some build here and there.

dolmen:
The universe that depends on logrus is already broken because of the case change.
At least this change helps to diagnose those build failures.

dgsb:
We did not receive any issue about the case change lastly, so I don't think there are so many broken build out there because of that.
Anyway, I kind of agree with the change, but as I said we can not merely merge that in a v1 version, I do not want to add more instability to any build using logrus.

dolmen:
The situation is so broken that in fact in my company we lived for a long time without noticing the case change. We have been bitten by the issue more than one year after the case change.

This patch enforces that everyone that upgrades will come in the new post-case-change world instead of staying in the ignorance. And it gives the information to the go tooling to help the user to fix the issue.

dolmen:
@dgsb #1041 (Novemeber 2019) is an example of how the import path change still affects dependent projects recently.

dgsb:
the use of go.mod is the new standard for package management in go build. In this context the fix is not useful anymore, but it may harm users just building whatever packages are currently in the gopath.

Imunhatep:
```
$ go mod vendor
go: github.com/rancher/rke/pki imports
	github.com/docker/docker/pkg/archive imports
	github.com/Sirupsen/logrus: github.com/Sirupsen/logrus@v1.6.0: parsing go.mod:
	module declares its path as: github.com/sirupsen/logrus
	        but was required as: github.com/Sirupsen/logrus
```

PR #860, opened 2018-11-09
Fix hook example

xrstf:
The example code for adding a hook would never add a successfully built syslog adapter because the error check was inverted.


dgsb:
@xrstf thanks for your contribution

PR #863, opened 2018-11-16
Add Generic Log functions with level via argument

lugray:
I wanted to be able to use `Log` with a variable level, and it seemed like I should add `Logf` and `Logln` if I was going to add `Log`.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -240,16 +240,18 @@ func (entry *Entry) write() {
 	}
 }
 
-func (entry *Entry) Trace(args ...interface{}) {
-	if entry.Logger.IsLevelEnabled(TraceLevel) {
-		entry.log(TraceLevel, fmt.Sprint(args...))
+func (entry *Entry) Log(level Level, args ...interface{}) {
+	if entry.Logger.IsLevelEnabled(level) {
+		entry.log(level, fmt.Sprint(args...))
 	}
 }
 
+func (entry *Entry) Trace(args ...interface{}) {
+	entry.Log(TraceLevel, args...)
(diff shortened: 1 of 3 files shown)


lugray:
> @lugray Do you only need Log() on Entry? I would think you'd want it on Logger?

In my case I did only need it on `Entry`, but you're right, it makes sense to make the change on `Logger` too, so I did that.

An example of why you'd want this, and where I use it:
```go
const ignoreError logrus.Level = math.MaxInt32

func runAndLogGitCmd(ctx context.Context, cmd shell.Supervisor, defaultLevel logrus.Level, levelOverrides map[string]logrus.Level)
	stdout, stderr, err := cmd.RunAndGetOutput()
	if err == nil {
		return stdout, nil
	}

	gitErr := parseGitCmdError(string(stderr), err)

	level := defaultLevel
	if levelOverrides != nil {
		if l, ok := levelOverrides[gitErr.ID()]; ok {
			level = l
		}
	}

	if level == ignoreError {
		return
	}

	log(ctx, gitErr).WithFields(gitErr.LogFields()).Log(level, "error while running git command")
}
```
Some git errors in our app should be fully ignored, others should be downgraded to `WarnLevel`, most should be `ErrorLevel`, but some commands are generally allowed to fail and so called with a `defaultLevel` of `WarnLevel`.

richpoirier:
@lugray Maybe add a `TestLog()` test case to logrus_test.go and then we'll be good to merge.

lugray:
Added test - thanks for the review on this.

PR #864, opened 2018-11-16
Remove colored output on windows

### LINKED ISSUE #862: Colorized output on windows 7 broken
When printing output with a TextFormatter to cmd or powershell on windows 7, the control sequences are rendered to the output.

``←[36mINFO←[0m[0000] Replay took: 4.0003ms                         ←[36msource←[0m=Badger``

This was already mentioned in #820 but probably never investigated.

Workaround: Just disable colors with 

```
log.SetFormatter(&logrus.TextFormatter{
		DisableColors: true,
})
```

ceriath:
Fixes #862

Currently the textformatter on windows outputs 

``←[31mERRO←[0m[0000] test windows``

when coloring is not disabled explicitly.
However, windows up to windows 8.1 does not support colored output on cmd entirely. Windows 10 added support for it, which is off by default and has to be enabled via registry or environment variable. Therefore i suggest removing colored output on windows entirely to make the output usable again.

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -102,7 +103,7 @@ func (f *TextFormatter) isColored() bool {
 		}
 	}
 
-	return isColored && !f.DisableColors
+	return isColored && !f.DisableColors && (runtime.GOOS != "windows")
 }
 
 // Format renders a single log entry
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -91,7 +91,7 @@ func (f *TextFormatter) init(entry *Entry) {
 }
 
 func (f *TextFormatter) isColored() bool {
-	isColored := f.ForceColors || f.isTerminal
+	isColored := f.ForceColors || (f.isTerminal && (runtime.GOOS != "windows"))
 
 	if f.EnvironmentOverrideColors {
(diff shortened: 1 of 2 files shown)


text_formatter.go
```diff
@@ -102,7 +103,7 @@ func (f *TextFormatter) isColored() bool {
 		}
 	}
 
-	return isColored && !f.DisableColors
+	return isColored && !f.DisableColors && (runtime.GOOS != "windows")
```
dgsb: Hello, I would like that the ForceColor field of the text formatter takes precedence on the environment check.
I expect that then this flag is set to true the user knows what he is doing.

text_formatter_test.go
```diff
@@ -443,7 +444,11 @@ func TestTextFormatterIsColored(t *testing.T) {
 				os.Setenv("CLICOLOR_FORCE", val.clicolorForceVal)
 			}
 			res := tf.isColored()
-			assert.Equal(subT, val.expectedResult, res)
+			if runtime.GOOS == "windows" {
```
dgsb: I think we need to updated the test  for windows environment where isColored always return false except when ForceColor is set to true.

dgsb:
@ceriath thanks for your contribution, I've added a few comments on you change. Let me know what you think.

ceriath:
@dgsb i agree, i changed it so setting ForceColor or the environment variables takes precedence. With the environment variables users who explicitly turned on colored output have an easy way to also enable it on logrus without recompiling the application

dgsb:
@ceriath that's exactly what I had in mind.

PR #870, opened 2018-12-05
Skip func pointer type value in fields

smacker:
Before there was introduced a fix for JSONFormatter when func type value
passed as Field. This commit does the same but for pointer to func.

Ref:
https://github.com/sirupsen/logrus/issues/642
https://github.com/sirupsen/logrus/pull/832


dgsb:
@smacker thanks for your contribution

PR #874, opened 2018-12-15
do not clear error formatting informative field

dgsb:
Based on work from @loren-osborn in #851


loren-osborn:
@dgsb, sorry for not getting back to this. I was sidetracked on other projects.

PR #883, opened 2019-01-08
syslog build constraints encapsulation

devintegral:
Logrus users have not to use build constraints with "github.com/sirupsen/logrus/hooks/syslog" import from here on.


devintegral:
PR #884 preferred.

Issue #887, opened 2019-01-13
Give the library user a method to customize the reportcaller output


GAZ082:
Hey. When I enable it i get an absolute path  to the code line:

`"file":"/home/gabriel/go/src/someprogram/package/subpackage/subpackage.go:303`

I just need the last few charactes, ie: `subpackage.go:303`

You guys know some way to get this that way?

Thanks!

richpoirier:
@GAZ082 unfortunately no, there's no way to do that currently.

quancore:
This definitely needs to implement otherwise, outputs seem very ugly. I am wondering what is hindering you to just truncate last part of caller method path?

GAZ082:
Ignorance? Haha. Have no idea how to work around this.

dgsb:
There is no actual way to do that from a library user standpoint. The formatting is of file data fields is done at the fomatter level which is not customizable.
We need to think how we would like to provide a library user to customize the output of such information.
The usual way is to use a hooks, but it is not feasible with this implementation.

GAZ082:
@dgsb no problem David, than you for being so nice. Looking forward this feature!

tossp:
I submitted a solution(#898 )  that I am using, which is expected to meet your needs, but I did not do performance testing.

GAZ082:
Hey, looks handy! Is there any way I can load this up in my local version of logrus? I'm sure some git command to suck your commit up and merge it in my local version?

dgsb:
@GAZ082 I've added a callback to the json formatter in order to give the user a chance to format the runtime.Frame himself. Would that help you ?

dgsb:
it's here #909

dgsb:
The same customization has been added to the text formatter #911

GAZ082:
Thank you @dgsb ! Been looking into the commit and could not figure it out how to implement it. Any example would be welcomed!

dgsb:
I've added an example in #913 
I'm going to merge that soon

Issue #867, opened 2018-11-20
Adding Config Options to SetReportCaller


karthik-ramanathan-3006:
Currently, `SetReportCaller` is a flag that can be toggled on and off. 
When set to true, by default, it returns the first frame that does not belong to the `Logrus` package.
The request is to make this configurable. It must be possible for the user to specify the level of the frame (counting from the last `Logrus` package) to be returned.

**Use case:** I have written a wrapper over `Logrus`. All calls to the logger go through this wrapper. When I set `SetReportCaller` to true, the File Name, Function Name and Line Numbers of the Wrapper are returned.
I would like to log the File Name, Function Name and Line Numbers of the next calling function in the stack.

montanaflynn:
I'd also like to specify which log levels to add this to, I mainly only want to see the caller for errors.

sepehrmm:
https://github.com/sirupsen/logrus/issues/972
https://github.com/sirupsen/logrus/pull/973

tossp:
Related PR: #989

frittentheke:
This is not stale ... there is a PR open.

frittentheke:
This is not stale ... there is a PR open.

frittentheke:
> This is not stale ... there is a PR open.

8-)

dclendenan:
hi bots - there's still a PR open

PR #898, opened 2019-01-27
add skipPackageNameForCaller

tossp:
It's just an implementation. It can only be used, but it's not perfect. I hope there's a better solution.😋

ex:
```golang
logrus.AddSkipPackageFromStackTrace("someprogram/package/subpackage")
logrus.AddSkipPackageFromStackTrace("tsl/core/log")
logrus.AddSkipPackageFromStackTrace("github.com/go-xorm/xorm")
```

#887

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -15,7 +15,7 @@ var (
 	bufferPool *sync.Pool
 
 	// qualified package name, cached at first use
-	logrusPackage string
+	skipPackageNameForCaller = make(map[string]bool, 1)
 
 	// Positions in the call stack when tracing to report the calling method
 	minimumCallerDepth int
@@ -38,6 +38,11 @@ func init() {
 
 	// start at the bottom of the stack before the package-name cache is primed
 	minimumCallerDepth = 1
+}
+
+// Set the global qualified package name.
+func SetSkipPackageNameForCaller(name string) {
+	skipPackageNameForCaller[name] = true
 }
 
 // Defines the key when adding errors using WithError.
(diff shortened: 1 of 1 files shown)


eightnoteight:
this should help in resolving #867 too

entry.go
```diff
@@ -40,6 +40,11 @@ func init() {
 	minimumCallerDepth = 1
 }
 
+// Set the global qualified package name.
+func SetSkipPackageNameForCaller(name string) {
```
lavoiesl: Mere suggestion:
I feel this is more of an `Add`. `Set` sounds like setting _a_ value.
For the user, `SkipPackageFromStackTrace` might make more intuitive sense than the "Caller" internal.

```suggestion
func AddSkipPackageFromStackTrace(name string) {
```

entry.go
```diff
@@ -40,6 +40,11 @@ func init() {
 	minimumCallerDepth = 1
 }
 
+// Set the global qualified package name.
```
lavoiesl: Add the examples here as well.

entry.go
```diff
@@ -163,18 +168,21 @@ func getCaller() *runtime.Frame {
...
-
 		// now that we have the cache, we can skip a minimum count of known-logrus functions
 		// XXX this is dubious, the number of frames may vary store an entry in a logger interface
-		minimumCallerDepth = knownLogrusFrames
+		if len(skipPackageNameForCaller) != 0 {
+			minimumCallerDepth = knownLogrusFrames + 1
+		} else {
+			minimumCallerDepth = knownLogrusFrames
+		}
+		skipPackageNameForCaller[getPackageName(runtime.FuncForPC(pcs[0]).Name())] = true
```
lavoiesl: I’d prefer having a local `logrusPackage` variable, this is quite hard to read.

entry.go
```diff
@@ -163,18 +168,21 @@ func getCaller() *runtime.Frame {
 
 	// cache this package's fully-qualified name
 	callerInitOnce.Do(func() {
-		logrusPackage = getPackageName(runtime.FuncForPC(pcs[0]).Name())
-
 		// now that we have the cache, we can skip a minimum count of known-logrus functions
 		// XXX this is dubious, the number of frames may vary store an entry in a logger interface
-		minimumCallerDepth = knownLogrusFrames
+		if len(skipPackageNameForCaller) != 0 {
```
lavoiesl: Can you explain what is going on here? 
I have the impression `minimumCallerDepth = knownLogrusFrames` should remain the same.

eightnoteight: Hi @lavoiesl @tossp ,
won't this be unused since we are already including the logrus package in the ignore map?

dgsb:
I'm not such fan of the interface, I wonder if it would not be more practical for the user to provide a callback function returning a boolean which would allow to implement any kind of logic.
Not sure what would be the parameter of the callback function though. Maybe just the current frame and and its index.

PR #907, opened 2019-02-20
Infinite recursion

### LINKED ISSUE #906: infinite recursion in Level.String()
There is an infinite recursion in `Level.String()` when `MarshalText()` calls `Level.String()` via the `%q`

```go
func TestLogLevelString(t *testing.T) {
    var loggerlevel logrus.Level
    loggerlevel = 32
    t.Logf("test: %s", loggerlevel.String())
}
```

I can make an proposal in a PR

noushavandijk:
This PR fixes #906.

### DIFF
--- a/logrus.go
+++ b/logrus.go
@@ -74,7 +74,7 @@ func (level Level) MarshalText() ([]byte, error) {
 		return []byte("panic"), nil
 	}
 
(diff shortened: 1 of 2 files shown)


richpoirier:
Thanks for the fix @noushavandijk! Good catch.

FYI @dgsb

dgsb:
indeed thanks @noushavandijk

PR #911, opened 2019-02-27
Add a CallerPrettyfier callback to the text formatter

dgsb:


DYDYSY:
when TextFormatter and ForceColors is true, The CallerPrettyfier function will execute 2 times.
this may not be necessary.

PR #918, opened 2019-03-08
Added Entry.OverrideLoggerLevel

xaionaro:
We need a way to set different logging levels for different entries.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -8,6 +8,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -55,6 +56,9 @@ type Entry struct {
 
 	// Time at which the log entry was created
 	Time time.Time
+
+	// OverrideLoggerLevel allows to set another effective Logger.Level for this Entry
+	OverrideLoggerLevel Level
 
 	// Level the log entry was logged at: Trace, Debug, Info, Warn, Error, Fatal or Panic
 	// This field will be set on entry firing and the value will be equal to the one in Logger struct field.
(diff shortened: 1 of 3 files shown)


dgsb:
Hi @xaionaro, what would be the use case ?

xaionaro:
@dgsb: we have a middleware that processes only 5% of requests in logging level "debug". The rest requests should be passed in logging level "error". Moreover other goroutines (which are not service any requests) should work in logging level "error", too.

dgsb:
Why don't you use two loggers instead ?

xaionaro:
Well, yes, it's possible. However we would like to do not duplicate the code that sets up the logger ([DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)). It initializes hooks and does other settings and it will be required to pass-through two the same loggers through the code -- looks a little ugly in our case.

However, of course, I will understand if this PR will be rejected :)

xaionaro:
There's still tag `need-more-info` on this PR. Can I help to get the info somehow?

dgsb:
Well, I'm not really fund of adding such a specific api for a single person. Because we will then have to support and maintain it for a long.
I really suggest you try to use a logger per level, because this is how the library has been defined.

xaionaro:
Well, it's required for a project I do in a company that I work for, so it's not quite a single person request.

However, of course, I will understand if you will reject this PR.

gguridi:
I also think that this scenario can be solved by creating more than one logger and there's no need for this PR.

If someone needs to control verbosity per message, or have special cases where different messages must be printed or not, it's part of your business logic and not what a logging library is supposed to do.

We all have specific cases and requirements from our products owners that are complex but that doesn't mean that we must make a library deviate from its purpose to accommodate them.

xaionaro:
> I also think that this scenario can be solved by creating more than one logger and there's no need for this PR.

I hardly remember the problem, but:
Let's for example there's a middleware which setups a logging context for an HTTP request and has access only to an Entry. This middleware for example understands GET-parameter `x_log_level=trace` (or `x_log_level=debug`). So what is required to do in this case? Am I understand the suggestion is to create new logger from scratch (so all settings of the initial `Entry` will be lost)? Or it's suggested to manually perform a deep copy of the logger on each request?

It is really bad when a dependency dictates how I should implement my tool.

PR #919, opened 2019-03-11
Add WithContext

tgwizard:
Add `WithContext(ctx context.Context)` method, and a `Context` property, on `Entry`. Contexts aren't that useful for the logging itself, but it can be very useful for hooks you may have configured.

Example: You have Bugsnag configured, which store request metadata in the context, and you want to notify Bugsnag on any error-level logs. Without passing along the context through the log entry, the Bugsnag report will lack a lot of interesting detail.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -2,6 +2,7 @@ package logrus
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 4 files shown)


dgsb:
Context are pretty useful indeed.
Thanks for your contribution @tgwizard

tgwizard:
Thanks @dgsb @lavoiesl!

PR #920, opened 2019-03-11
Add CHANGELOG for v1.4.0

tgwizard:
I've gone through the commits in https://github.com/sirupsen/logrus/compare/v1.3.0...master, this seems to be the things that changed (except for examples and README updates - should they be here?)

I assume creating a new version is simplest to do in the GitHub UI, with tag `v1.4.0` pointing to to the merge commit for this PR?


tgwizard:
Thanks @richpoirier!

PR #921, opened 2019-03-13
Rewrite if-else-if-else to switch statements

sosiska:
This commit rewrites some if-else-if-else chains as a switch.
Based on Go style guide:
https://golang.org/doc/effective_go.html#switch

Found by https://go-critic.github.io/overview#ifElseChain-ref

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -95,11 +95,10 @@ func (f *TextFormatter) isColored() bool {
 	isColored := f.ForceColors || (f.isTerminal && (runtime.GOOS != "windows"))
 
 	if f.EnvironmentOverrideColors {
-		if force, ok := os.LookupEnv("CLICOLOR_FORCE"); ok && force != "0" {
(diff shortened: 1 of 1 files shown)


dgsb:
I'm not such fan of mixing if and switch control structure.
I'm not such fan either of using switch for something else than  value or range value comparison.

sosiska:
> I'm not such fan of mixing if and switch control structure.
> I'm not such fan either of using switch for something else than value or range value comparison.

I can return if-else, but in isColored() func we can squash 2 checks into one with `||`, because they doing one thing - `isColored = false`.

sosiska:
Ping.

sosiska:
Ping

dgsb:
@sosiska yes we can merge the two branches together if you want

sosiska:
> @sosiska yes we can merge the two branches together if you want

I do not quite understand you 🤔, but i like current variation of code state.

PR #923, opened 2019-03-14
Make Writer() methods inherit their receiver's log level

dmlambea:
Calls to Writer() method in logger/entry should inherit its logger/entry's level instead of defaulting to InfoLevel.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -7,9 +7,9 @@ import (
 	"strings"
 )
 
-// Writer at INFO level. See WriterLevel for details.
+// Writer at the Logger's level. See WriterLevel for details.
 func (logger *Logger) Writer() *io.PipeWriter {
-	return logger.WriterLevel(InfoLevel)
+	return logger.WriterLevel(logger.GetLevel())
 }
 
 // WriterLevel returns an io.Writer that can be used to write arbitrary text to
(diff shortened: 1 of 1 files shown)


writer.go
```diff
@@ -7,15 +7,15 @@ import (
...
+	return logger.WriterLevel(logger.GetLevel())
 }
 
 func (logger *Logger) WriterLevel(level Level) *io.PipeWriter {
 	return NewEntry(logger).WriterLevel(level)
 }
 
 func (entry *Entry) Writer() *io.PipeWriter {
-	return entry.WriterLevel(InfoLevel)
+	return entry.WriterLevel(entry.Logger.GetLevel())
```
lavoiesl: Shouldn't this honour `entry.Level` if it's not `nil`?
```suggestion
	level := entry.Level
	if level == nil {
		level = entry.Logger.GetLevel()
	}
	return entry.WriterLevel(level)
```

dmlambea: Not sure of this. Unless I'm missing something, I've seen in entry.go source that Level only gets populated when the log(...) method is called, which means that Level is just a copy of Logger's Level for the last time the log(...) method was called, or nil if it never happened before. One can force entry's Level to have a value, of course, but it would be lost after a call to log(...) so I don't see any benefit of setting this fields.

Regarding the Writer() code, I see that checking for the entry's value to be non-nil, otherwise going for Logger's value, makes no practical difference to the resulting value of Level. On the other hand, the addition of the "if" clause for no added value increments the cognitive complexity of the method, thus making the code unnecessarily complicated, IMHO.

lavoiesl: Ah, you're right.

lavoiesl (approved):

PR #927, opened 2019-03-25
Test with Go 1.12

ceh:
Add Go 1.12 to Travis CI build matrix.

AppVeyor doesn't seem to support Go 1.12 out of the box yet according to https://www.appveyor.com/docs/windows-images-software/#golang, deferring for later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,21 @@ matrix:
         - go get golang.org/x/sys/windows
       script:
         - go test -race -v ./...
+    - go: 1.12.x
+      env: GO111MODULE=on
+      install:
+        - go mod download
+      script:
+        - go test -race -v ./...
+    - go: 1.12.x
+      env: GO111MODULE=off
+      install:
+        - go get github.com/stretchr/testify/assert
+        - go get golang.org/x/crypto/ssh/terminal
+        - go get golang.org/x/sys/unix
+        - go get golang.org/x/sys/windows
+      script:
+        - go test -race -v ./...
     - go: 1.10.x
       install:
         - go get github.com/stretchr/testify/assert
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,8 @@
+installtestdeps:
+	go get github.com/stretchr/testify/assert
+	go get golang.org/x/crypto/ssh/terminal
+	go get golang.org/x/sys/unix
+	go get golang.org/x/sys/windows
+
+.PHONY: \
```
cjgu: Isn't .PHONY supposed to be at the top of the file?

ceh: No, I don't think so. See for example https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

.travis.yml
```diff
@@ -6,10 +6,7 @@ matrix:
   include:
     - go: 1.10.x
       install:
-        - go get github.com/stretchr/testify/assert
```
lavoiesl: Instead of introducing a new Makefile, which doesn't do much, how about adding a `before_install block` that is common to all blocks? https://docs.travis-ci.com/user/customizing-the-build/

ceh: Won't that block be executed for the go module builds as well? I didn't add one because I don't wan't a `go get` to interfere or pollute the namespace before `go mod download` is executed.

lavoiesl: Ah, you're right. Good thinking.

Makefile
```diff
@@ -0,0 +1,8 @@
+installtestdeps:
```
dgsb: I'd rather not add a makefile for just a single phony target.

ceh:
Updated the PR to not use a Makefile.

dgsb:
@ceh thanks for your contribution

Issue #928, opened 2019-03-25
Dropping dependency on golang.org/x/crypto


tandr:
This library is pulled for only one check - to see if output is a terminal or not.

Would it be possible just to inline one used function from it? This would allow us (and all people who pulls it in) to drop the dependency on it as well. The import with "crypto" in it triggers additional attention from our legal department, and we have to explain that it does not have anything "cryptographic" in it - just a check for is it a terminal or not.

Ideally, `ssh/terminal` should be not under crypto , but I will try to ping golang folks about it directly.

lavoiesl:
I like this idea; I opened a PR.

lavoiesl:
Unfortunately, the implementation is os-dependent and it’d be quite many files to bring over to get rid of ssh/terminal.

I suggest leaving it as is for now.

tandr:
8 files to bring over. Password-reading related stuff can be removed.

https://github.com/golang/go/issues/31044 got reference for https://github.com/golang/go/issues/13104, but honestly, it does not look very promising short term.

tandr:
Actually... there is a whole bunch of file in logrus starting with term_*, we can try to copy these functions   there.

dgsb:
I'd rather not, it would be hard to keep up with updates

tandr:
There are no updates since ... 2011 and 2015

PR #929, opened 2019-03-26
Inline the use of IsTerminal to remove crypto dependency

lavoiesl:
golang.org/x/crypto is quite a big package

Fixes https://github.com/sirupsen/logrus/issues/928


lavoiesl:
Nevermind, this is much more complicated because the `IsTerminal` implementation is os-dependent.

PR #930, opened 2019-03-26
Got rid of IsTerminal call to reduce external dependencies

tandr:
Addresses #928

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ matrix:
     - go: 1.10.x
       install:
         - go get github.com/stretchr/testify/assert
-        - go get golang.org/x/crypto/ssh/terminal
         - go get golang.org/x/sys/unix
         - go get golang.org/x/sys/windows
       script:
@@ -22,7 +21,6 @@ matrix:
       env: GO111MODULE=off
       install:
         - go get github.com/stretchr/testify/assert
-        - go get golang.org/x/crypto/ssh/terminal
         - go get golang.org/x/sys/unix
         - go get golang.org/x/sys/windows
       script:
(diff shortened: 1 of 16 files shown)


terminal_check_aix.go
```diff
@@ -1,9 +1,9 @@
-// +build !appengine,!js,!windows,aix
+// +build aix
```
lavoiesl: Can this, at least, be moved to a subpackage with a `doc.go` explaining where it comes from and why it was done this way?

tandr: I am not sure I follow, sorry. What should I move into doc.go ?

lavoiesl: Move all the files you just added to a `terminal` package (sub folder) and create a `doc.go` in there to add some comments.

tandr: Right now all the functions exposed by this terminal_* group (I added only 2 files, but could've done with 1 in retrospect) are not exported (`checkIfTerminal`). If we want to keep it that way, they will need to go into "internal/terminal" instead, so even if they will become un-importable from any other projects. This change scope will grow. We can do it in a separate MR or do it here. How do you want to proceed ?

lavoiesl: You can create a `IsTerminal` function in the package, which would be public, just like it is in the `ssh/terminal` package. That way, no need to change any other file.

tandr: We need to change other files to adjust imports. I moved all of them under import/terminal, please take a look

tandr: I will add an additional review since there is more than one way to solve it

tandr: Created https://github.com/sirupsen/logrus/pull/932 that be closer to what you have asked

tandr:
#932 solves it in a maybe a less structured (_some_ of the responsibilities of detecting platform moved, not all of them), but definitely less intrusive way

PR #932, opened 2019-03-26
Drop x/crypto package dependency

### LINKED ISSUE #928: Dropping dependency on golang.org/x/crypto

tandr:
Closes #928

### DIFF
--- a/internal/terminal/terminal_check_bsd.go
+++ b/internal/terminal/terminal_check_bsd.go
@@ -1,0 +1,13 @@
+// +build darwin dragonfly freebsd netbsd openbsd
+
+package terminal
+
+import "golang.org/x/sys/unix"
+
+const ioctlReadTermios = unix.TIOCGETA
+
+func IsTerminal(fd int) bool {
+	_, err := unix.IoctlGetTermios(fd, ioctlReadTermios)
+	return err == nil
+}
+
--- a/internal/terminal/terminal_check_unix.go
+++ b/internal/terminal/terminal_check_unix.go
@@ -1,0 +1,13 @@
+// +build linux aix
+
+package terminal
+
+import "golang.org/x/sys/unix"
+
+const ioctlReadTermios = unix.TCGETS
+
+func IsTerminal(fd int) bool {
+	_, err := unix.IoctlGetTermios(fd, ioctlReadTermios)
+	return err == nil
+}
+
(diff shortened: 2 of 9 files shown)


internal/terminal/terminal_check_bsd.go
```diff
@@ -0,0 +1,12 @@
...
+package terminal
+
+import "golang.org/x/sys/unix"
+
+const ioctlReadTermios = unix.TIOCGETA
+
+func IsTerminal(fd int) bool {
+	_, err := unix.IoctlGetTermios(fd, ioctlReadTermios)
+	return err == nil
+}
```
lavoiesl: Newline

internal/terminal/terminal_check_unix.go
```diff
@@ -0,0 +1,12 @@
...
+package terminal
+
+import "golang.org/x/sys/unix"
+
+const ioctlReadTermios = unix.TCGETS
+
+func IsTerminal(fd int) bool {
+	_, err := unix.IoctlGetTermios(fd, ioctlReadTermios)
+	return err == nil
+}
```
lavoiesl: Newline

tandr: added

internal/terminal/terminal_check_bsd.go
tandr: added

tandr:
Would it be possible to slap a tag on it once it is merged in? Or it is not sufficiently big change to bump  it to v1.4.1 ?

lavoiesl:
A patch bump would be in order, yes. I’m just waiting for someone else to review.

dgsb:
@tandr that looks good to me.

tandr:
@dgsb David, would you be able to tell when do you guys expect v1.4.1 get tagged? Reason I am asking is that or I need to pull it now and then second time when you tag it, or if you expect it tomorrow or in very near future, I can just wait for it instead.

thaJeztah:
+1 on tagging a v1.4.1 for this 🤗

dgsb:
@tandr and @thaJeztah we can not yet, we do not want to make public the new terminal interface.

Issue #933, opened 2019-03-28
use of internal package github.com/sirupsen/logrus/internal/terminal not allowed


pacmessica:
breaking change was introduced yesterday. I see the following error on all my builds:

```
../../../github.com/Sirupsen/logrus/terminal_check_notappengine.go:9:2: use of internal package github.com/sirupsen/logrus/internal/terminal not allowed
```

artagel:
Same issue.

dgsb:
It's kind of unexpected.
Can you provide more context, or even better a code sample using logrus which exhibits the build break ?
Which go version are you using ?
On a sidenote,  I highly recommend that you use tagged version instead of the tip master.

lavoiesl:
I merged #935 for the time being, just so users aren't blocked, but let's keep this one open to figure out why.

artagel:
I’ll be able to get full details when I get back home. 
I’m using https://github.com/vouch/vouch-proxy
You can’t specify the tag version when using go get, so I ended up pulling the repo and checking out the tag in my gopath src. This obviously isn’t a long term solution.

dgsb:
@artagel you may want to use dep https://golang.github.io/dep/ for managaing your project dependencies.
Or you use the next gen dependency manager through modules.
So that you don't rely on master anymore.

artagel:
Yeah, I think this is an issue golang needs to solve better.  Dep is being abandoned as far as I've read, and they are trying to put out something new.

dgsb:
@pacmessica or @artagel can you provide more feedback on how to reproduce the issue ? because our temporary forces us to expose an API we do not want to make public.

artagel:
Sure, 
The push you guys did yesterday fixed the issue for now...  
I still have a copy of the master from before that fix...this is what happens:
But using vouch-proxy as an example:
git clone https://github.com/vouch/vouch-proxy
cd vouch-proxy
go get -v 
go run main.go 
../../Sirupsen/logrus/terminal_check_notappengine.go:9:2: use of internal package not allowed

The problem seems to be the use of the word internal in your import.
Check out: 
https://golang.org/cmd/go/#hdr-Internal_Directories
https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit

tandr:
Well... usage of internal was intentional. At this point we can as well put it under main directory and get rid of exported IsTerminal altogether.

@artagel 
[...]

PR #935, opened 2019-03-28
Move terminal package

pacmessica:
fixes issue where terminal_check_notappengine.go can't access terminal
package since terminal package is in an internal package

https://github.com/sirupsen/logrus/issues/933


lavoiesl:
Merging to unblock users

PR #937, opened 2019-03-28
Test more platforms

tandr:
Test 3 architecture minus "go modules mode" on Go 1.10 - modules do not exist on Go 1.10


dgsb:
@tandr thanks for your contribution. The travis file is much more readable now.

PR #941, opened 2019-03-29
return new entry for Entry.WithContext

hrxu01:
Just like Entry.WithError,  Entry.WithField and Entry.WithFields, return a new entry for Entry.WithContext. If we don't return a new entry, the old entry with context will be released to entry pool, this may cause some unexpected errors.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -103,8 +103,7 @@ func (entry *Entry) WithError(err error) *Entry {
 
 // Add a context to the Entry.
 func (entry *Entry) WithContext(ctx context.Context) *Entry {
-	entry.Context = ctx
-	return entry
+	return &Entry{Logger: entry.Logger, Data: entry.Data, Time: entry.Time, err: entry.err, Context: ctx}
 }
 
 // Add a single field to the Entry.


entry.go
```diff
@@ -103,8 +103,7 @@ func (entry *Entry) WithError(err error) *Entry {
 
 // Add a context to the Entry.
 func (entry *Entry) WithContext(ctx context.Context) *Entry {
```
ntwarijoshua: Why is the receiver a pointer since you dont manipulate it?

hrxu01: Keep consistent with Entry.WithError, Entry.WithField and Entry.WithFields methods in the same file.

ntwarijoshua: ok

tandr: That pointer means memory barriers will be different when creating a new Entry. I would argue that the rest of these With* calls should have receiver passed by value.

hrxu01: > That pointer means memory barriers will be different when creating a new Entry. I would argue that the rest of these With* calls should have receiver passed by value.

I think having receiver passed by value or pointer are both ok in this case,  just make sure to keep consistent. For the present logic, we create a new Entry inside With* methods and copy some but not all fields from receiver, pointer receiver might be slightly better (less and cleaner code).

PR #944, opened 2019-04-02
remove field if val is empty string for func and file field 

hrxu01:
Changes are made in text_formatter.go.

Users can set CallerPrettyfier to customize  function and file fields, but if users want only one of these two fields, there is no way.

I make some changes so that if users set one field to empty string, logrus will not output that field.

And the json_formatter.go already had this logic.


dgsb:
@hrxu01 thanks for your contribution

PR #948, opened 2019-04-05
Guard unsafe imports behind !appengine

### LINKED ISSUE #947: terminal_check_solaris.go doesn't have a build tag and is breaking the build on AppEngine
https://github.com/sirupsen/logrus/blob/9b3cdde74fbe9443d704467498a7dcb61a79de9b/terminal_check_solaris.go#L1

I think this file should have a build tag like:
```go
// +build solaris
```
similar to https://github.com/sirupsen/logrus/blob/9b3cdde74fbe9443d704467498a7dcb61a79de9b/terminal_check_unix.go#L1

I believe this is the reason logrus currently fails to build on AppEngine standard projects (where we can't use `golang.org/x/sys/unix`).

Hexcles:
"unsafe" (in this case, imported via "golang.org/x/sys/unix") cannot be
used on AppEngine, so we need to add the "!appengine" build constraint
to files that use it (namely a few terminal_check_*.go).

Fixes #947.

### DIFF
--- a/terminal_check_unix.go
+++ b/terminal_check_unix.go
@@ -1,4 +1,5 @@
 // +build linux aix
+// +build !appengine
 
 package logrus
 
@@ -10,4 +11,3 @@ func isTerminal(fd int) bool {
 	_, err := unix.IoctlGetTermios(fd, ioctlReadTermios)
 	return err == nil
 }
-
--- a/terminal_check_windows.go
+++ b/terminal_check_windows.go
@@ -1,4 +1,4 @@
-// +build !appengine,!js,windows
+// +build windows
 
 package logrus
 
--- a/terminal_check_bsd.go
+++ b/terminal_check_bsd.go
@@ -1,4 +1,5 @@
 // +build darwin dragonfly freebsd netbsd openbsd
+// +build !appengine
 
 package logrus
 
@@ -10,4 +11,3 @@ func isTerminal(fd int) bool {
 	_, err := unix.IoctlGetTermios(fd, ioctlReadTermios)
 	return err == nil
 }
-
--- a/terminal_check_solaris.go
+++ b/terminal_check_solaris.go
@@ -1,3 +1,6 @@
+// +build solaris
+// +build !appengine
+
 package logrus
 
 import (


terminal_check_unix.go
```diff
@@ -1,4 +1,5 @@
 // +build linux aix
+// +build !appengine
```
dgsb: I think this single change is enough to fix your build issue, can you try ?

Hexcles: This would fix the build problem on *my* machine, because I'm running Linux.

However, if one wants to build and deploy AppEngine projects from various BSD or Solaris, they'd need the other two `// +build !appengine` constraints in the corresponding files.

terminal_check_windows.go
```diff
@@ -1,4 +1,4 @@
-// +build !appengine,!js,windows
+// +build windows
```
Hexcles: Now this change is truly not necessary, but I believe that's a nice-to-have drive-by cleanup. Feel free to revert it (by accepting this suggestion) if you don't like it.

```suggestion
// +build !appengine,!js,windows
```

Hexcles:
@dgsb what do you reckon?

terminal_check_windows.go
dgsb: We need to keep the js tag, it is used to identify gopherjs build

terminal_check_unix.go
dgsb: I guess the tag when building for an appengine are always linux and appengine ?
The host build doesn't matter really here, we just want to differentiate standard linux build for linux+appengine target.
Or am I wrong somewhere ?

dgsb:
@Hexcles did you try my suggestion to only add the `!appengine` build constraint in the `terminal_check_unix.go` file ?
I would also like that you merge master in your branch so that you can benefit a full cross compilation for all target supported by go 1.12 compiler in travis.

Hexcles:
@dgsb sorry this thread has been going to my personal email which I haven't been keeping close track of.

I tried your suggestion on both Linux and macOS, and it does work. Only `terminal_check_unix.go` needs to be modified. I thought `terminal_check_bsd.go` would also need to be modified when I deployed from macOS, but apparently that's not the case from my experiment. It seems like the Go AppEngine toolchain is more like a crossbuild setup targeting Unix/Linux.

I also rebased the change on top of the current master.

Hexcles:
ping @dgsb

thaJeztah:
Closing; this should be fixed (at least in master)

PR #956, opened 2019-04-23
Update x/sys/unix to fix AIX support

Helflym:
The current version of golang.org/x/sys/unix is breaking AIX build.


dgsb:
@Helflym thanks for your contribution

PR #962, opened 2019-05-10
Update Example to Append to the log file not to overwrite

Clivern:
update README example to append to the log file not overwrite

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ func main() {
   log.Out = os.Stdout
 
   // You could set this to any `io.Writer` such as a file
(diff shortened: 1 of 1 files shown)


dgsb:
@Clivern thanks for your contribution. But this change is unneeded, the example is quite correct. After writing to the file the current offset is already advanced by the amount of written data. Concurrent write on the same file descriptor are also protected by a mutex.

dgsb:
silly me, the example is quite incorrect, and your example version much more informative. Thanks again.

PR #973, opened 2019-05-23
add caller skip

dongshimou:
can help me skip some stack

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -164,7 +164,7 @@ func getPackageName(f string) string {
 }
 
 // getCaller retrieves the name of the first non-logrus calling function
-func getCaller() *runtime.Frame {
+func getCaller(skip int) *runtime.Frame {
 
 	// cache this package's fully-qualified name
 	callerInitOnce.Do(func() {
@@ -187,6 +187,10 @@ func getCaller() *runtime.Frame {
 
 		// If the caller isn't part of this package, we're done
 		if pkg != logrusPackage {
+			if skip != 0 {
+				skip--
+				continue
+			}
 			return &f
 		}
 	}
@@ -218,7 +222,7 @@ func (entry Entry) log(level Level, msg string) {
 	entry.Level = level
 	entry.Message = msg
 	if entry.Logger.ReportCaller {
-		entry.Caller = getCaller()
+		entry.Caller = getCaller(entry.Logger.CallerSkip)
 	}
 
 	entry.fireHooks()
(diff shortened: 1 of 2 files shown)


rentiansheng:
Thanks. I also need this feature. No need to implement

kync:
> Why not `depth := runtime.Callers(minimumCallerDepth + skip, pcs)` rather than using if statement?

better:
``` callerInitOnce.Do(func() {
		pcs := make([]uintptr, 2)
		_ = runtime.Callers(0, pcs)
		logrusPackage = getPackageName(runtime.FuncForPC(pcs[1]).Name())

		// now that we have the cache, we can skip a minimum count of known-logrus functions
		// XXX this is dubious, the number of frames may vary
		minimumCallerDepth = knownLogrusFrames + skip
	})```

**Comment by updogliu** [comment]
I am interested in this as well. Anything blocking it moving forward?

**Comment by xqzhang2015** [comment]
Hi @dongshimou,

One proposal, adding these codes to align with other params:

<pre>
func (logger *Logger) SetCallerSkip(skip int) {
	logger.mu.Lock()
	defer logger.mu.Unlock()
	logger.SetCallerSkip = skip
}

func SetCallerSkip(skip int) {
	std.SetCallerSkip(skip)
}
</pre>

**Comment by xqzhang2015** [comment]
BTW, maybe the system-level logger maybe not enough. One option is to define func at call-level, like

- DebugDepth(depth int, args ...interface{})
- InfoDepth(depth int, args ...interface{})
- ...

**Review by iamvivekv — APPROVED** [review]

**Comment by dvdalilue** [comment]
This is a feature I would like to have but It seems there is no intention to include new features in the project.

I ended up using the `CallerPrettyfier` field in the `JSONFormatter`:
```go
formatter := &logrus.JSONFormatter{
        // ...
	CallerPrettyfier: callerFn,
	// ...
}

func callerFinder(f *runtime.Frame) (function string, file string) {
	path, err := os.Getwd()
	if err == nil {
		pcs := make([]uintptr, 50)
		_ = runtime.Callers(0, pcs)
		frames := runtime.CallersFrames(pcs)

		for next, again := frames.Next(); again; next, again = frames.Next() {
			if strings.HasPrefix(next.File, path) {
				return next.Function, fmt.Sprintf("%s:%d", next.File, next.Line)
			}
		}
	}

	return f.Function, fmt.Sprintf("%s:%d", f.File, f.Line)
}
```
I know it's not beautiful, and I would prefer a feature for it, but it works to get the information I need.

jjeejj:
I need this feature, how is it now？

alramdein:
Yo, any updates from this feature?

PR #974, opened 2019-05-23
Fix output coloring for terminals on windows

kevpar:
Recent versions of Windows support ANSI VT100 escape codes to e.g. allow
changing color of output. This support must be opted into by a call to
SetConsoleMode with ENABLE_VIRTUAL_TERMINAL_PROCESSING. Logrus already
has code to make this call (via the sequences package in
terminal_check_windows.go). However, color was never used due to the
check for Windows in text_formatter.go. This check was added due to the
possibility of running on old versions of Windows which don't support
this new console mode.

This change resolves these issues by attempting to set the console mode,
if it fails, we return false from checkIfTerminal, which will cause
colored output to not be used. This should allow color to work properly
on new versions of Windows, without affecting behavior on older
versions.

Signed-off-by: Kevin Parsons <kevpar@microsoft.com>

### DIFF
--- a/terminal_check_windows.go
+++ b/terminal_check_windows.go
@@ -10,11 +10,12 @@ import (
 	sequences "github.com/konsorten/go-windows-terminal-sequences"
 )
 
-func initTerminal(w io.Writer) {
+func initTerminal(w io.Writer) error {
 	switch v := w.(type) {
 	case *os.File:
-		sequences.EnableVirtualTerminalProcessing(syscall.Handle(v.Fd()), true)
+		return sequences.EnableVirtualTerminalProcessing(syscall.Handle(v.Fd()), true)
 	}
+	return nil
 }
 
 func checkIfTerminal(w io.Writer) bool {
@@ -28,7 +29,7 @@ func checkIfTerminal(w io.Writer) bool {
 		ret = false
 	}
 	if ret {
-		initTerminal(w)
+		ret = (initTerminal(w) == nil)
 	}
 	return ret
 }
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -88,7 +88,7 @@ func (f *TextFormatter) init(entry *Entry) {
 }
 
 func (f *TextFormatter) isColored() bool {
-	isColored := f.ForceColors || (f.isTerminal && (runtime.GOOS != "windows"))
+	isColored := f.ForceColors || f.isTerminal
 
 	if f.EnvironmentOverrideColors {
 		if force, ok := os.LookupEnv("CLICOLOR_FORCE"); ok && force != "0" {
--- a/text_formatter_test.go
+++ b/text_formatter_test.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"runtime"
 	"sort"
 	"strings"
 	"testing"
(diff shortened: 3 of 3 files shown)


cup:
I dont think an incremental fix like this is the best solution.

the best solution would be one that support color for all windows versions.

we know this is possible as other projects are already doing it:

~~~go
package main
import "github.com/labstack/gommon/log"
func main() {
   log.EnableColor()
   log.Warn("abcde")
}
~~~

kevpar:
> I dont think an incremental fix like this is the best solution.
> 
> the best solution would be one that support color for all windows versions.
> 
> we know this is possible as other projects are already doing it:
> 
> ```go
> package main
> import "github.com/labstack/gommon/log"
> func main() {
>    log.EnableColor()
>    log.Warn("abcde")
> }
> ```

We could use a package like `github.com/mattn/go-colorable` (which is what labstack/gommon uses). It parses the output itself for VT100 sequences and calls the corresponding Windows console APIs to achieve the same effect. Though I do think using the native VT100 support is better on platforms that support it.

I know taking this change is not a complete solution for colored output across all versions of Windows. However, this is a small change, and will at least enable colored output on new versions (it looks like this is supported starting in Anniversary Update, mid 2016). This change doesn't prevent anyone from making a later change to use `go-colorable` on old versions of Windows, if that's desirable.

cup:
@kevpar 

My view is a repo with 11,000 stars should have equal or greater support than a repo with 200 stars. This repo is the most popular Go logging package, by far. With that comes some duty I say of at least attempting wide support.

I think you know that it would be a disservice to merge a request like this when the author knows full well that a better solution is available.

jstarks:
@cup This change will improve the majority of Windows users' experience at no cost to users that are on old versions of Windows. Adding support for old versions of Windows will add complexity and code size with zero benefit to most users. Within 18 months, all versions of Windows in mainstream support by Microsoft will support the flag that this change depends on.

@dgsb Would you mind taking a look? We would like to ensure color output is enabled on Windows in Docker, containerd, crictl, etc.

cup:
@jstarks Windows 7 still has 38% market share by some measurements:

https://netmarketshare.com/operating-system-market-share.aspx?id=platformsDesktopVersions

thats more than Windows 8, all Mac and Linux **combined**.

If you want to hand wave that away thats your business, but youre only kidding yourself, not me.

jstarks:
@cup Thanks for sharing your perspective. Our uses of logrus are all in container-related software, such as Docker, which does not work and will never work on Windows 7. So you can understand why we are not focused on such old versions of Windows.

Even outside our own use cases, it's not obvious to me that the OS distribution of the users of logrus matches that of the general Windows population.

cup:
@jstarks I dont see that you are a collaborator on this repo, or even a contributor:

https://github.com/sirupsen/logrus/commits?author=jstarks

so it appears that you are talking about of turn. This is not a microsoft repo, so it would appear you have no decision making abilities in regards to this issue. If thats not the case I apologize, but your aggressiveness is making it seem that way.

jstarks:
I am not intending to be aggressive or offensive. I am just trying to advocate for this change, which is important to our users. I am sorry if I was careless in my tone.

You're right that I have no authority here. It appears you don't either, so I think I will leave this conversation for now until someone who does has time to review the change. Thanks!

kevpar:
@freeformz Looks like you are the Logrus maintainer now? Could you please take a look at this? :)

This change adds value and greatly improves the output on modern Windows systems. While it doesn't address colored output on older versions of Windows, it also doesn't detract from the ability to address that in the future, if such a change is desired.

cup:
I will repeat my original objection.

Some would consider this a half done patch, which given that existing drop-in solutions are already available for older Windows versions. That makes it unacceptable in my mind, and even selfish.

Issue #959, opened 2019-05-03
[Feature Request] Add an pad option log level text


lynncyrin:
### What

~I would like a `TextFormatter` option that outputs in a 5 character / 4 character + 1 padding space output mode. I'm tentatively calling it `FourAndFiveCharacterTruncation` (feedback requested on that name)~

Taking in the feedback provided in the comments, I would instead like a `TextFormatter` option that outputs the level text with padding up to the max length. I'm tentatively calling it `PadLevelText`. It would look like so

```
INFO    some information
WARNING some warning
DEBUG   some debug information
```

### Why

The above example is in contrast to the current default behavior of

```
INFO some information
WARN some warning
DEBU some debug information
```

And the `DisableLevelTruncation` behavior of

```
INFO some information
WARNING some warning
DEBUG some debug information
```

(`FATAL`, `PANIC`, and `TRACE` not explicitly mentioned above for the sake of conciseness, they would all output at 5 characters just like `DEBUG`)

### Who

This feature request is essentially the "pad" option that mentioned in https://github.com/sirupsen/logrus/pull/407#issuecomment-274781317

I intend to create the PR for this myself, if you all don't mind ^^ Haven't had an OSS contribution in a too-long now. I'll start on that if I get >=1 👍 react from a Contributor

lynncyrin:
@dgsb, @sirupsen, @lavoiesl would any of you happen to have an opinion about this feature request? I'd love to help out with this package!

dgsb:
Indeed the truncated level looks weird to me.
But rather than pad to an arbitrary value and truncate beyond, why not pad to the largest level string representation ?
wdyt ?

lynncyrin:
Padding to the largest level string representation would look good and be less work to implement. I'm for it! @dgsb 👍

---

(_tangentially:_ I think my ideal end state would be pad to the largest string representation and change the cli text of `WARNING` to `WARN`. But the warn change is a different matter, and possibly a breaking change, so I'll consider that separately.)

lynncyrin:
For the sake of completeness, here's what a pad option would look like

```
INFO    some information
WARNING some warning
DEBUG   some debug information
```

I'll edit the OP to reflect this instead (...)

gauravaditya:
Hi,

Just wondering if the "PadLevelText" option on text formatter is not yet released, though the docs say it is usable. Let me know if i am missing something.
Thanks

lynncyrin:
[...]

PR #977, opened 2019-06-05
Add an pad option log level text

lynncyrin:
Closes https://github.com/sirupsen/logrus/issues/959

I spent a few hours working on the tests. They're not perfect, but I don't think they can be improved substantially without doing a large refactor (which wouldn't be appropriate for me to do in this PR).

The implementation of the option itself is very straightforward, thanks to `Sprintf`.

I added a bunch of comments that possibly don't belong in the source code? If it's non-ideal having them there, then I can remove them.

The PR should be pretty comprehensive, but totally let me know if there's something else I missed ✨

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ time="2015-03-26T01:27:38-04:00" level=fatal method=github.com/sirupsen/arcticcr
 ```
 Note that this does add measurable overhead - the cost will depend on the version of Go, but is
 between 20 and 40% in recent tests with 1.6 and 1.7.  You can validate this in your
-environment via benchmarks:
+environment via benchmarks: 
 ```
 go test -bench=.*CallerTracing
 ```
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ The built-in logging formatters are:
     [github.com/mattn/go-colorable](https://github.com/mattn/go-colorable).
   * When colors are enabled, levels are truncated to 4 characters by default. To disable
     truncation set the `DisableLevelTruncation` field to `true`.
-  * When outputting to a TTY, its often helpful to visually scan down a column where all the levels are the same width. Setting the `PadLevelText` field to `true` enables this functionality, by adding padding to the level text.
+  * When outputting to a TTY, its often helpful to visually scan down a column where all the levels are the same width. Setting the `PadLevelText` field to `true` enables this behavior, by adding padding to the level text.
   * All options are listed in the [generated docs](https://godoc.org/github.com/sirupsen/logrus#TextFormatter).
 * `logrus.JSONFormatter`. Logs fields as JSON.
   * All options are listed in the [generated docs](https://godoc.org/github.com/sirupsen/logrus#JSONFormatter).
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ time="2015-03-26T01:27:38-04:00" level=fatal method=github.com/sirupsen/arcticcr
 ```
 Note that this does add measurable overhead - the cost will depend on the version of Go, but is
 between 20 and 40% in recent tests with 1.6 and 1.7.  You can validate this in your
-environment via benchmarks: 
+environment via benchmarks:
 ```
 go test -bench=.*CallerTracing
 ```
@@ -354,6 +354,7 @@ The built-in logging formatters are:
     [github.com/mattn/go-colorable](https://github.com/mattn/go-colorable).
   * When colors are enabled, levels are truncated to 4 characters by default. To disable
     truncation set the `DisableLevelTruncation` field to `true`.
+  * When outputting to a TTY, its often helpful to visually scan down a column where all the levels are the same width. Setting the `PadLevelText` field to `true` enables this functionality, by adding padding to the level text.
   * All options are listed in the [generated docs](https://godoc.org/github.com/sirupsen/logrus#TextFormatter).
 * `logrus.JSONFormatter`. Logs fields as JSON.
   * All options are listed in the [generated docs](https://godoc.org/github.com/sirupsen/logrus#JSONFormatter).
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ The built-in logging formatters are:
     [github.com/mattn/go-colorable](https://github.com/mattn/go-colorable).
   * When colors are enabled, levels are truncated to 4 characters by default. To disable
     truncation set the `DisableLevelTruncation` field to `true`.
-  * When outputting to a TTY, its often helpful to visually scan down a column where all the levels are the same width. Setting the `PadLevelText` field to `true` enables this behavior, by adding padding to the level text.
+  * When outputting to a TTY, it's often helpful to visually scan down a column where all the levels are the same width. Setting the `PadLevelText` field to `true` enables this behavior, by adding padding to the level text.
   * All options are listed in the [generated docs](https://godoc.org/github.com/sirupsen/logrus#TextFormatter).
 * `logrus.JSONFormatter`. Logs fields as JSON.
   * All options are listed in the [generated docs](https://godoc.org/github.com/sirupsen/logrus#JSONFormatter).
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -57,6 +57,10 @@ type TextFormatter struct {
 	// Disables the truncation of the level text to 4 characters.
 	DisableLevelTruncation bool
 
+	// PadLevelText Adds padding the level text so that all the levels output at the same length
+	// PadLevelText is a superset of the DisableLevelTruncation option
+	PadLevelText bool
+
 	// QuoteEmptyFields will wrap empty fields in quotes if true
 	QuoteEmptyFields bool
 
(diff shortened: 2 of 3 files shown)


lynncyrin:
cc @dgsb for code review!

text_formatter.go
```diff
@@ -217,9 +221,12 @@ func (f *TextFormatter) printColored(b *bytes.Buffer, entry *Entry, keys []strin
 	}
 
 	levelText := strings.ToUpper(entry.Level.String())
-	if !f.DisableLevelTruncation {
+	if !f.DisableLevelTruncation && !f.PadLevelText {
 		levelText = levelText[0:4]
 	}
+	if f.PadLevelText {
+		levelText = fmt.Sprintf("%-7s", levelText)
```
dgsb: I'd rather that to not be hard coded, so that when/if change the available trace levels this should not need to be updated.
This could be done in an init function by iterating on AllLevels to find the largest one.

lynncyrin: I had that same idea while writing this! I'll add that 👍

lynncyrin:
I got a little bit distracted, but haven't forgotten about this ^^ the requested changes are pretty clear, I'll add them Soon ™️

text_formatter.go
lynncyrin: @dgsb I took a short break from this to do other stuff, but now I have returned!

I added https://github.com/sirupsen/logrus/pull/977/commits/693469de8f5d1e17ff6f3d8b479f36e5f97c5b7d and https://github.com/sirupsen/logrus/pull/977/commits/8ba442aca65c8333a2c78c566e064cc0e8ad1f88 since ya last looked at this PR 👀. Those commits add the dynamic max length thing!

lynncyrin:
> I've never used colored output. Does this work for logs that are e.g. piped to a file?

You only get colored output (and by extension, padded level text) when logging to a terminal
 https://github.com/sirupsen/logrus/blob/8ba442aca65c8333a2c78c566e064cc0e8ad1f88/text_formatter.go#L105-L106

https://github.com/sirupsen/logrus/blob/8ba442aca65c8333a2c78c566e064cc0e8ad1f88/text_formatter.go#L191-L192

text_formatter.go
```diff
@@ -217,9 +232,18 @@ func (f *TextFormatter) printColored(b *bytes.Buffer, entry *Entry, keys []strin
...
 
 	levelText := strings.ToUpper(entry.Level.String())
-	if !f.DisableLevelTruncation {
+	if !f.DisableLevelTruncation && !f.PadLevelText {
 		levelText = levelText[0:4]
 	}
+	if f.PadLevelText {
+		// Generates the format string used in the next line, for example "%-6s" or "%-7s".
+		// Based on the max level text length.
+		formatString := "%-" + strconv.Itoa(f.levelTextMaxLength) + "s"
```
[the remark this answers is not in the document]
lynncyrin: It's wrapped in functionality that makes sure it's only called once

https://github.com/sirupsen/logrus/blob/8ba442aca65c8333a2c78c566e064cc0e8ad1f88/text_formatter.go#L86

https://github.com/sirupsen/logrus/blob/8ba442aca65c8333a2c78c566e064cc0e8ad1f88/text_formatter.go#L185

docs: https://golang.org/pkg/sync/#Once

I don't precisely know the boundaries of the "once" here, but I would guess it's "once for entire execution of the program.

therealplato:
> > I've never used colored output. Does this work for logs that are e.g. piped to a file?
> You only get colored output (and by extension, padded level text) when logging to a terminal

Thanks, am I correctly following that the existing padding approach is not supported outside of colored output? I was wondering why this was applied to color only

lynncyrin:
> Thanks, am I correctly following that the existing padding approach is not supported outside of colored output? I was wondering why this was applied to color only

Primarily because the existing `DisableLevelTruncation` flag is nested inside the `printColored` function, and `PadLevelText` is (conceptually) an extension of `DisableLevelTruncation`.

https://github.com/sirupsen/logrus/blob/8ba442aca65c8333a2c78c566e064cc0e8ad1f88/text_formatter.go#L221-L246

It sounds like what you're looking for is the ability to enable either `DisableLevelTruncation` or `PadLevelText` without colored output? If so, you should totally create an issue for that 📝

text_formatter_test.go
```diff
@@ -172,6 +172,97 @@ func TestDisableLevelTruncation(t *testing.T) {
...
+			}
+
+			// Assertion: the level text should still contain the string representation of the level
+			if !strings.Contains(strings.ToLower(logLineWithPadding), val.level.String()) {
+				t.Errorf("log line \"%s\" should contain the level text \"%s\" when padding is enabled", logLineWithPadding, val.level.String())
+			}
+
+			// Assertion: the level text should be in its padded form now
+			if val.paddedLevelText != "" && !strings.Contains(logLineWithPadding, val.paddedLevelText) {
+				t.Errorf("log line \"%s\" should contain the padded level text \"%s\" when padding is enabled", logLineWithPadding, val.paddedLevelText)
```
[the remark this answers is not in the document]
lynncyrin: ✔️  https://github.com/sirupsen/logrus/pull/977/commits/dcce32597dee9bd34880bf1e7a42a328eaec13d1

text_formatter.go
```diff
@@ -79,12 +84,22 @@ type TextFormatter struct {
...
+	levelTextMaxLength int
 }
 
 func (f *TextFormatter) init(entry *Entry) {
 	if entry.Logger != nil {
 		f.isTerminal = checkIfTerminal(entry.Logger.Out)
 	}
+	// Get the max length of the level text
+	for _, level := range AllLevels {
+		levelTextLength := len(level.String())
```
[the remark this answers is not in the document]
lynncyrin: ✔️  https://github.com/sirupsen/logrus/pull/977/commits/af6ed964ef540511b4982e50cee85e6fc9715763

README.md
```diff
@@ -354,6 +354,7 @@ The built-in logging formatters are:
     [github.com/mattn/go-colorable](https://github.com/mattn/go-colorable).
   * When colors are enabled, levels are truncated to 4 characters by default. To disable
     truncation set the `DisableLevelTruncation` field to `true`.
+  * When outputting to a TTY, its often helpful to visually scan down a column where all the levels are the same width. Setting the `PadLevelText` field to `true` enables this behavior, by adding padding to the level text.
```
richpoirier: `it's`

text_formatter.go
```diff
@@ -217,9 +233,18 @@ func (f *TextFormatter) printColored(b *bytes.Buffer, entry *Entry, keys []strin
 	}
 
 	levelText := strings.ToUpper(entry.Level.String())
-	if !f.DisableLevelTruncation {
+	if !f.DisableLevelTruncation && !f.PadLevelText {
 		levelText = levelText[0:4]
 	}
```
therealplato: @richpoirier if you want to avoid multi-bool conditions, while keeping present behavior (someone is probably parsing the output and expecting it to remain consistent) here's a possibility:
```go
levelText = fmt.Sprintf(formatString, strings.ToUpper(entry.Level.String())
if !f.PadLevelText {
  levelText = strings.TrimSuffix(levelText, " ")
}
if !f.DisableLevelTruncation {
  levelText = levelText[0:4] // this also breaks for many non-english scripts btw
}
```

lynncyrin:
> Have we considered padding by default when `DisableLevelTruncation` is turned on? Seems like a sane default to me and would simplify this.
> I'm trying to avoid adding more and more conditionals to the code. (if !f.DisableLevelTruncation && !f.PadLevelText ... if f.PadLevelText).

I would prefer that, but I don't have to context required to know if that's a safe change to make.

If there's someone with write access who would approx me making that change, then I'll change this PR to do that.

README.md
lynncyrin: ✔️  https://github.com/sirupsen/logrus/pull/977/commits/539b8af839f189b659e6c7f62b4524a460fbae20

text_formatter.go
lynncyrin: I'll do this if it's still relevant after I hear back from either @richpoirier or @dgsb about this comment https://github.com/sirupsen/logrus/pull/977#issuecomment-506567331

richpoirier:
@lynncyrin @therealplato is right, I'm sure there are users out there parsing the output, unfortunately so probably a bad idea.

text_formatter.go
lynncyrin: Alright! I'll change the PR to do this in a bit

text_formatter.go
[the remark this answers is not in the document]
dgsb: Yes it is for the full execution time, conceptually it is a mutex with a counter.

dgsb:
> @lynncyrin @therealplato is right, I'm sure there are users out there parsing the output, unfortunately so probably a bad idea.

Agree, we shouldn't change the output for existing users with a given TextFormatter configuration

dgsb:
@lynncyrin thanks for your contribution, sorry for the delay on my side

lynncyrin:
Its no problem, @dgsb thanks for working with me on this!

PR #982, opened 2019-06-24
Add DisableEscapeHTML parameter to JSONFormatter

Dakatan:
Hi!

I use logrus, but http request and response xml is escaped by json encoder.
So, i add this parameter.

Logging sample.

1. `DisableEscapeHTML`=`false` or `default`
```text
{"level":"panic","msg":"","time":"0001-01-01T00:00:00Z","xml":"\u003cxml\u003e\u0026\u003c/xml\u003e"}
```

2. `DisableEscapeHTML`=`true`
```text
{"level":"panic","msg":"","time":"0001-01-01T00:00:00Z","xml":"<xml>&</xml>"}
```

Thank you!

### DIFF
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -51,6 +51,9 @@ type JSONFormatter struct {
 
 	// PrettyPrint will indent all json logs
 	PrettyPrint bool
+
+	// DisableEscapeHTML allows disabling HTML escape.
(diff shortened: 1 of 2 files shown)


markphelps:
Fixed in #1113. Thanks!

PR #983, opened 2019-06-26
Add a dynamic business log path

ltachi1:
Add log path, since each module may be stored under different directories, dynamic assignment is needed when calling


dgsb:
Hello @ltachi1, can't we use here a regular field ?

PR #986, opened 2019-07-08
Add error callback for get sentry error with a callback

RashadAnsari:


flimzy:
I think the callback should be optional. Unless I'm overlooking something, this code will panic when no callback is defined.

logger.go
```diff
@@ -39,8 +39,12 @@ type Logger struct {
 	entryPool sync.Pool
 	// Function to exit the application, defaults to `os.Exit()`
 	ExitFunc exitFunc
+	// ErrorCallback is called when any error occurs in Logrus
```
freeformz: If I'm mistaken ErrorCallback is only called when a formatter or hook returns an error and there are other places where errors could happen no?

hooks/syslog/syslog.go
```diff
@@ -4,10 +4,8 @@ package syslog
 
 import (
 	"fmt"
-	"log/syslog"
-	"os"
-
 	"github.com/sirupsen/logrus"
+	"log/syslog"
```
freeformz: please remove this formatting change, github.com/sirupsen/logrus imports should be formatted at the end with a space between them and std lib imports.

exported.go
```diff
@@ -25,6 +25,11 @@ func SetFormatter(formatter Formatter) {
 	std.SetFormatter(formatter)
 }
 
+// SetErrorCallback sets the error callback for end user.
```
freeformz: Suggestion:
```go
// SetErrorCallback sets the Error ErrorCallback for the standard logger
```

Or something along those lines.

logger.go
```diff
@@ -349,3 +354,9 @@ func (logger *Logger) ReplaceHooks(hooks LevelHooks) LevelHooks {
 	logger.mu.Unlock()
 	return oldHooks
 }
+
```
freeformz: Please add a godoc comment for this exported function. Something along the lines of:

```go
// SetErrorCallback sets the ErrorCallback function of the logger
```

logger.go
freeformz: It would be good to also tell developers using logrus what the default is (it does nothing, but I can't assume that).

entry.go
```diff
@@ -245,7 +244,8 @@ func (entry *Entry) fireHooks() {
 	defer entry.Logger.mu.Unlock()
 	err := entry.Logger.Hooks.Fire(entry.Level, entry)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Failed to fire hook: %v\n", err)
+		err := fmt.Errorf("Failed to fire hook: %v\n", err)
```
freeformz: Suggestion: pass the original error, not a wrapper string error. fmt.Errorf'ing here nukes things like pkg/errors and IIRC go 1.13's error wrapping. If it's important to differentiate hook errors from formatting errors I recommend using the go1.13 error formatter verb (https://golang.org/doc/go1.13#error_wrapping) and maybe implement a HookError and FormatterError types with Unwrap methods as per: https://golang.org/pkg/errors/#Unwrap

In either case error messages shouldn't start with capitals.

entry.go
```diff
@@ -254,11 +254,13 @@ func (entry *Entry) write() {
 	defer entry.Logger.mu.Unlock()
 	serialized, err := entry.Logger.Formatter.Format(entry)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Failed to obtain reader, %v\n", err)
+		err := fmt.Errorf("Failed to obtain reader, %v\n", err)
```
freeformz: See above.

entry.go
freeformz: PS: supporting go1.13 formatting may require the go1.13 implementation to be extracted to build tag protected implementations.

logger.go
```diff
@@ -78,12 +82,13 @@ func (mw *MutexWrap) Disable() {
...
-		Level:        InfoLevel,
-		ExitFunc:     os.Exit,
-		ReportCaller: false,
+		Out:           os.Stderr,
+		Formatter:     new(TextFormatter),
+		Hooks:         make(LevelHooks),
+		Level:         InfoLevel,
+		ExitFunc:      os.Exit,
+		ReportCaller:  false,
+		ErrorCallback: func(e error) {},
```
freeformz: Should the default hook implement current behavior of printing the error to stderr? That seems to me like the best option.

freeformz:
@flimzy The default callback is a no-op `func(_ error) {}`, so it would not panic.

PR #988, opened 2019-07-10
add logatee test package

itzamna314:
I'm not sure where this belongs in logrus, if at all. This is a nice utility package for projects that use `*logrus.Logger` instances. It provides nicely-formatted test output, and makes it easy to inspect log messages from tests. Its been really useful for me.

The tests are failing because the import paths are pointed to my fork. If there's interest in adopting this package I'll clean up the import paths. I'm also open to suggestions of somewhere else where this utility could live.

### DIFF
--- a/logatee/README.md
+++ b/logatee/README.md
@@ -1,0 +1,40 @@
+# Logatee
+A timid logrus. Useful for testing code that uses logrus instances.
+
+# Usage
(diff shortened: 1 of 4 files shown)


freeformz:
Thank You! I think this would make a **great separate** package.

I'd be happy to add a link to it from the readme if you want to open a PR to do that, but ATM it's not something *I'm* interested in adding to logrus directly if it can stand alone.

PR #1008, opened 2019-08-27
Add a getter for the Entry.err

VadimKulagin:
Hello!

I want to implement my Formatter and for this I need access to the `Entry.err` field. 
But, unfortunately, this is not possible now, since this field is private.

I ended up adding a getter for this field:
```go
// Err returns an internal field formatting error.
func (entry *Entry) Err() string {
	return entry.err
}
```

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -201,6 +201,11 @@ func (entry Entry) HasCaller() (has bool) {
 		entry.Caller != nil
 }
 
+// Err returns an internal field formatting error.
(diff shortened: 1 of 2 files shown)


VadimKulagin:
Are there any chances for merge?

pashaosipyants:
@markphelps 
Still relevant enhancement, not stale.
Any updates/thoughts?

Issue #463, opened 2017-01-08
Custom Log Levels


yoiang:
Hello logrus folks!
It was mentioned in issue #121 that being able to make your own custom log levels would be a feature by 1.0, wanted to make a place to track the feature's progress.

Thanks for the great work!

stevvooe:
@yoiang Do you have a proposal for how this would work?

yoiang:
hey @stevvooe! Do you mean what would the API look like or how would it actually be implemented in logrus' guts technically? I am not expert with either but can take some time and figure it out depending on what you're asking!

stevvooe:
@yoiang So, the issue is that we already have values that have an ordering. Allowing users to define these may have an impact on how that ordering interacts. Without going into details, bolting this on is likely pretty complex.

@sirupsen may have a more detailed plan for how this might work.

cretz:
I suggest simply exposing `logrus.Entry.Log(Level, string)` (and `logrus.Entry.Logf(Level, string, ...interface{})`) and adding a `logrus.SetCustomLevelName(Level, string)` and `logrus.GetCustomLevelName(Level) string, bool` which are simply backed by a thread-safe map. Then `logrus.Level.String()` can use the latter before defaulting to "unknown".

Then we can all simply create levels via:

```go
const TraceLevel = logrus.DebugLevel + 1

func init() {
  logrus.SetCustomLevelName(TraceLevel, "trace")
}
```

This of course does not solve issues where people may want their level in between other levels. The way to solve that with a backwards compatible impl is to keep the `Level` type and make a new interface and what not:

```go
type Leveler interface {
  Cmp(other Leveler) int
  LevelString() string
}

func (level Level) Cmp(other Leveler) int {
  switch other := other.(type) {
    case Level:
      if level > other {
        return 1
      } else if level < other {
        return -1
      }
      return 0
    default:
      return -other.Cmp(level)
  }
}

func (level Level) LevelString() string { return level.String() }
```

Then of course have `logrus.Entry.Log(Leveler, string)` and `logrus.Entry.Logf(Leveler, string, ...interface{})`. Granted it may have a bit of performance overhead introducing a type switch in every log statement, but it's a thought. I think the first proposal is good enough to solve people's notice/trace wants (or just go ahead and acknowledge there's a level below debug for many common use cases).

stevvooe:
@cretz Would it be sufficient just to add a `Trace` level to logrus? I'm really not sure about the effect bolting something on like this.

[...]

PR #1009, opened 2019-08-27
Add the ability to override log level values

michael-bud:


michael-bud:
Not sure who the maintainers are anymore but could this be looked at please?

@dgsb @sirupsen @freeformz @freeformzSFDC

freeformz:
@michael-bud Please describe the use case(s) a little so that we can do our best to understand the problem(s) being solved by this change.

michael-bud:
We have a number of different technologies across our architecture which when get fed into a centralised logging system. In make it easier to search for things of 'warn' log level across Go, Python, PHP, Java etc. projects we try to standardise the log level values (e.g. all the same case) but this requires some flexibility in the terms used; similarly we use the functionality that already exists in logrus that allows for renaming of certain fields to standardise those too.

I don't know any of the people who did the +1s on the original PR OP; they may have other use cases too.

json_formatter.go
```diff
@@ -43,6 +54,21 @@ type JSONFormatter struct {
 	// }
 	FieldMap FieldMap
 
+	// LogLevelMap allows users to customize the values used for the various log levels
+	// As an example:
+	// formatter := &JSONFormatter{
+	//   	LogLevelMap: LogLevelMap{
+	// 		 PanicLevel:  1,
```
freeformz: This example isn't valid. The value on the right side needs to be a string, so "1", "2", "3", ...

michael-bud:
(a) Yep, as you've pointed out, it's only really possible by replacing the whole JSON formatter with a custom one
(b) The reason I didn't do it as a custom function was to keep consistency with where we're doing this elsewhere in the JSON formatter, specifically where we allow customisation of the field names for things like the message, log level and timestamp.

json_formatter.go
```diff
@@ -20,6 +20,17 @@ func (f FieldMap) resolve(key fieldKey) string {
 	return string(key)
 }
 
+// LogLevel map allows customization of the values used to describe log levels
+type LogLevelMap map[Level]string
```
freeformz: Please rename to `LevelMap` and fix the good to correspond.

json_formatter.go
freeformz: Please make the example valid. Better yet, make an `ExampleLevelMap` function in the [test](https://golang.org/pkg/testing/#hdr-Examples) file.

freeformz:
@michael-bud That makes sense. I requested a few (minor) changes that I'd like to see before merging.

freeformz:
@michael-bud So...
After I posted that, and was reading #463, I started to wonder if this is solely a function of the JSONFormatter or not. So now I'm not so sure about my statements above.

For instance Level currently hard defines the string definition used for the stock levels. Maybe that should be more flexible and could help finish off #463. Thoughts ?

michael-bud:
@freeformz I'd rather not try and work that into this as I think it will make this significantly more complex than it needs to be and #463 looks like quite a massive change whereas this is relatively minor. The use cases are also quite different.

lnenad:
+1, i think having the ability to define custom levels is a necessary thing.

michael-bud:
@freeformz This PR is waiting for a response from you I think?

iamolegga:
Hi, is this ready to be merged? I guess there are just minor fixes left, really looking forward for this feature 👍

zx8:
Would really love to see this merged, but progress seems to have stalled.

I'm more than happy to pick it up and make the changes necessary to get it merged, if @michael-bud / @freeformz can confirm the current state of things?

PR #1011, opened 2019-08-31
Field Filters

pradykaushik:
Allow filtering of fields while formatting.

One might want to filter out some fields based on some criteria. One such criteria might be to not 
include fields with empty values.

Defined `FieldFilter` to be a function that accepts a key (string) and a value (interface{}) and returns boolean indicating whether to allow the field or not.
Return `true` if field should be allowed.
Return `false` if field should be filtered out.

Retrofitted `TextFormatter` and `JSONFormatter` to compose `[]FieldFilter` and use them (if any) while formatting. Included a `FieldFilter` called `IgnoreEmtpyValueFieldFilter` that can be used to filter out fields that have an empty value (zero value).

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,21 @@ requestLogger.Info("something happened on that request") # will log request_id a
 requestLogger.Warn("something not great happened")
 ```
 
+#### Filtering Fields
+
+It might sometimes be required that some fields be filtered out of logs for various reasons.
+For example, one might want to filter out fields that contain empty values.
+
+A `FieldFilter`can be added while instantiating formatters.
+```go
+formatter := &TextFormatter{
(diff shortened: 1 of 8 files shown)


pradykaushik:
@sirupsen and @dgsb for review.

freeformz:
This is all possible with hooks: https://play.golang.org/p/bampaVhZMmf

pradykaushik:
> Thank you for your interest, but this is all possible _now_ via hooks: https://play.golang.org/p/bampaVhZMmf

Great! Thanks for your response.

pradykaushik:
@freeformz can I close this PR as it has already been addressed?

pradykaushik:
> @freeformz can I close this PR as it has already been addressed?

Yes this PR can be closed.

freeformz:
Please close it. I can't (yet).

PR #1019, opened 2019-09-08
Fire all hooks, even in case of an error

flimzy:
The current behavior aborts logging with an error, the first time a hook returns an error.

For me this meant that a misconfiguration of loggly resulted in sentry not getting logs.  This PR changes that behavior, such that all hooks will now be called, even in the case of an error with one of them.


freeformz:
The semantics of Fire can't be changed in the v1.X.X release series because that would change the [logrus.Hook](https://godoc.org/github.com/sirupsen/logrus#Hook) interface and break everyone who uses Hooks.

It's debatable whether or not it's a good idea to have errors cancel processing the rest of the hooks. As you point out a misconfigured hook or one that starts erroring for a random reason greatly impacts your ability to continue logging properly.

My suggestion would be to extend Hook a little and implement something like*:

```go
type Continuer interface {
  Continue() bool
}

// If the error returned from Fire is a Continuer and Continue() is true then the 
// rest of the error will be logged and the remaining Hooks will be called.
```

IMO it's important not to remove Fire's error though. Some hooks will filter message/data contents to remove PII info and if those hooks were to fail for some reason, it could be a security incident to log the unfiltered data.

*Modulo naming, which is hard.

flimzy:
> The semantics of Fire can't be changed in the v1.X.X release series because that would change the logrus.Hook interface and break everyone who uses Hooks.

Do you think this PR changes the semantics? If so, how?

Is there any documented mention that the first error aborts logging? As I see it, the current behavior is an undocumented bug (unless it is documented somewhere--that documentation needs to be made more explicit), and this change just fixes a bug. It doesn't change any interfaces or calling conventions.

> IMO it's important not to remove Fire's error though. Some hooks will filter message/data contents to remove PII info and if those hooks were to fail for some reason, it could be a security incident to log the unfiltered data.

This change doesn't remove the error. Were you under the impression that it did?

flimzy:
I think I see what you're talking about now...

> The semantics of Fire can't be changed...

LevelHooks is part of the public API. I didn't realize it was. I'll adjust my PR.

flimzy:
I've pushed a new implementation of this PR, which retains the original `LevelHooks.Fire()` API.

It adds a new, unexported type, to report all errors. If we think others will call `LevelHooks.Fire()` directly, and may want access to all errors, that type should perhaps be exported. I'm not sure if that's a meaningful use case, though, so maybe wait until someone requests it?

freeformz:
Continuing to run the remaining hooks after a hook returned an error would also be a change that is too large, at least IMO, for anything more than a full version release (v1.4.2 -> v2). It's a significant change to the behavior of hooks. And it's a possible security problem (see my comment above).

I think it's necessary for hook authors and/or users of hooks to decide if hook processing should stop after the first error or not. I'm also not convinced one way or the other the best way to go about that.

Related to this there is also #986, which is related, but different.

Maybe something combined that addresses everything detailed in both issues would be workable?

flimzy:
Thanks for the additional comments.

How would you feel about a configuration option to enable this behavior?  I'm not sure how that would look, but if you agree with the approach on principle, I'll work on it.

If not, I think at minimum it's appropriate to document that the first hook failure will abort logging. I hope that such a doc change PR won't be controversial :)

freeformz:
Please do the doc PR. FWIW: I thought it was already documented, but re-reading the godoc it appears that I was wrong.

I feel that a PR to add something like "FireAllHooks", defaulting to false (maintains backwards compatibility), with appropriate documentation detailing default and true behavior is a decent option. It should also play well with #986.

flimzy:
I've updated this PR now with the new configuration flag (off by default). Please let me know your thoughts.

entry.go
```diff
@@ -245,7 +245,13 @@ func (entry *Entry) fireHooks() {
 	defer entry.Logger.mu.Unlock()
 	err := entry.Logger.Hooks.Fire(entry.Level, entry)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Failed to fire hook: %v\n", err)
+		if merr, ok := err.(multiErr); ok {
+			for _, e := range merr {
```
freeformz: Why not do formatting in `multiErr.Error()`?

hooks.go
```diff
@@ -21,14 +21,38 @@ func (hooks LevelHooks) Add(hook Hook) {
...
 }
 
+// multiErr may contain multiple errors.
+type multiErr []error
+
+func (e multiErr) Error() string {
+	if e == nil || len(e) == 0 {
+		return ""
+	}
+	return e[0].Error()
```
freeformz: I would expect something more along the lines of: https://play.golang.org/p/P6EyQ4prr3s

It's debatable if adding the \n is good though.

entry.go
flimzy: You mean move the `"Failed to fire hook: "` bit into multiErr's `Error()` method?

hooks.go
flimzy: That's a good point. I really wish there were a standardize way to handle multiple errors in Go.

This makes me think it may actually be a good idea, even now, to export the `multiErr` type, so that consumers of the library can get at the underlying error values, if they want to.

hooks.go
```diff
@@ -21,14 +21,38 @@ func (hooks LevelHooks) Add(hook Hook) {
...
+	}
+	return e[0]
+}
+
 // Fire all the hooks for the passed level. Used by `entry.log` to fire
-// appropriate hooks for a log entry.
+// appropriate hooks for a log entry. In case of an error, the first error
+// encountered will be returned, but all hooks will fire.
 func (hooks LevelHooks) Fire(level Level, entry *Entry) error {
+	merr := make(multiErr, 0)
```
freeformz: append handles nil slices, so this isn't necessary: https://play.golang.org/p/9M1TC9lnkcQ

logger.go
```diff
@@ -29,6 +29,11 @@ type Logger struct {
 	// Flag for whether to log caller info (off by default)
 	ReportCaller bool
 
+	// FireAllHooks indicates that all hooks should be fired, even in case of an
+	// error. When off (default), when an error occurs firing one of the hooks,
```
freeformz: Maybe?
```
When false (default), the first hook that returns an error will stop any unfired hooks from firing and that error will be returned. When true, all hooks are fired and any errors are collected and returned.
```

Does this mean that mutliErr should be mentioned and exported?

i.e. I'm asking myself how would I process / handle the errors in the case of true and the documentation doesn't really give me any guidance.

entry.go
freeformz: No, that should probably stay here. I meant the ranging. I guess I could argue either case, so not sure.

flimzy: Oh right, I see what you mean. It's a good point.

This, and your other comments, have me really thinking about the best way to handle these multiple errors.

flimzy:
So I'm digging into the error proposals for Go, and the closest relevant one I think is [this](https://gist.github.com/gregwebs/5a14a83970d607411310a3e0281d55be). It's only a proposal, and probably won't be accepted as is, at least not for a long time, but it's simple, and I think it might be the right approach. Basically, it suggests implementing a specific interface:

```
type ErrorGroup interface {
       Errors() []error
}
```

I could make `multiErr` satisfy this interface, without exporting `multiErr` directly. Then it's just a matter of documenting how to access the multiple errors, probably with a brief example.

flimzy:
I've updated the PR once again with some changes to multi-error handling. I don't know if any or all of it is a good idea. I welcome your thoughts :)

freeformz:
Nice. I like the use of the interface to define the behavior of Errors().
There are a few minor nits I'd like to clean up and I'd like to take a look at supporting go1.13 error wrapping if possible, but they don't need to be done in this PR IMO.

flimzy:
> I'd like to take a look at supporting go1.13 error wrapping if possible

What, specifically, would you like to support here?  I added `Unwrap()` to the `multiErr` type, but it only returns the first error, since the Go 1.13 errors don't have any explicit support for multiple errors. There's probably more that could be done, but it wasn't obvious to me which direction to take.

freeformz:
I think it's worth *experimenting* with the %w formatting verb as well, but like I said, that isn't something that needs to happen in this PR, IMO.

PR #1024, opened 2019-09-19
Add terminal_check_js.go file, for compatibility with GopherJS

flimzy:
This allows logrus to compile with GopherJS.

The test suite still doesn't work with GopherJS, due to the use of `runtime.Caller()` in the test-specific `getPackage()` method.

If running the test suite against GopherJS is desirable, I suggest doing that in a separate PR (and probably adding it to .travis.yml, too).  I'm happy to take on that task, if it would be considered useful.

### DIFF
--- a/terminal_check_bsd.go
+++ b/terminal_check_bsd.go
@@ -1,4 +1,5 @@
 // +build darwin dragonfly freebsd netbsd openbsd
+// +build !js
 
 package logrus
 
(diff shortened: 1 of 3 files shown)


flimzy:
> Do you think it would make sense to add a GOOS=js (or is js an GOARCH?) run to the travis testing matrix?

It's probably a good idea, but it's not nearly as trivial as this PR, as it requires setting up a test environment with nodejs, etc.

It would also be good to target Go/WASM in the CI pipeline. Also should probably be a separate PR.

freeformz:
SGTM

PR #1047, opened 2019-10-23
fix race conditions on entry

### LINKED ISSUE #1046: Race condition in entry.Data on calling WithFields() method
```
fatal error: concurrent map iteration and map write
runtime.mapiternext(0xc000bbee98)
	/usr/local/go/src/runtime/map.go:858 +0x579 fp=0xc000bbedd8 sp=0xc000bbed50 pc=0x1011289
vendor/github.com/sirupsen/logrus.(*Entry).WithFields(0xc0001065b0, 0xc0009d3f50, 0xc0009d3f50)
	vendor/github.com/sirupsen/logrus/entry.go:117 +0xef fp=0xc000bbef08 sp=0xc000bbedd8 pc=0x1479b0f```

lwsanty:
closes #1046

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -113,6 +113,8 @@ func (entry *Entry) WithField(key string, value interface{}) *Entry {
 
 // Add a map of fields to the Entry.
 func (entry *Entry) WithFields(fields Fields) *Entry {
+	entry.Logger.mu.Lock()
+	defer entry.Logger.mu.Unlock()
 	data := make(Fields, len(entry.Data)+len(fields))
 	for k, v := range entry.Data {
 		data[k] = v
--- a/entry_test.go
+++ b/entry_test.go
@@ -134,7 +134,7 @@ func TestEntryWithIncorrectField(t *testing.T) {
 
 	fn := func() {}
 
-	e := Entry{}
+	e := &Entry{Logger: New()}
 	eWithFunc := e.WithFields(Fields{"func": fn})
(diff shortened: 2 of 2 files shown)


blocknonip:
This is still a bug. It occurs with any use of Entry and a hook (i.e. logrustash).

markphelps:
Thank you @blocknonip!!. Sorry for the stalebot spam. I thought it would only label/comment on issues, not PRs too.

flimzy:
This lock is far too aggressive. It locks the entire logger, rather than a single entry as it should.

dgsb:
Agree, we should find a better solution than this one.

markphelps:
@flimzy my apologies for merging this so quickly. would you want to work on a PR to lock only the entry? If not I can early this coming week.

flimzy:
I looked at it briefly before posting my comment, and it wasn't immediately obvious the proper solution.  It looks like we need to add a mutex to the `Entry` type, and lock it in `WithFields` and in `fireHooks`, but my first attempt lead to a deadlock.

It's possible that we have other overly-aggressive locks which could be relaxed, too, but it's tricky to do that with confidence, without tests covering all possible races (which are incredibly hard to test).

I'll try again if I have time, but you may get to it faster than me.

markphelps:
@flimzy ok I created #1122 to track. I'll also try taking a look

PR #1058, opened 2019-10-28
chore: added ifFail func

ZergsLaw:


dgsb:
@ZergsLaw Hello, can you open an issue first so that we can discuss the new API you want to implement

ZergsLaw:
Hello, I wanted to add a convenient functionality to work with linter, namely:
```
defer log.WarnInFail(rs.body.Close)
func (l log) WarnInFail(fn func() error)  {
  if err := fn(); err != nil {
    log.Warn(err)
  }
}
```
or 
```
db, err := db.Open()
if err != nil {y:i}.
	return err
}
log.WarnIfFail(db.Close)
```

However, my tests were local, but in CI I didn't, that's why I decided to close PR, to understand the reason of the test crash, as well as to understand the logger better.

Translated with www.DeepL.com/Translator

P.S. Sorry for the inconvenience :)

Issue #181, opened 2015-05-14
recover panic into a log message


stevvooe:
I may be missing something but there doesn't seem to be a way to recover from a panic and emit a panic-level message without causing another panic.

This is useful for handling panics uniformly inside of a hook.

skoushan:
So it appears that there's still no way to emit a panic-level message without causing another panic?

aybabtme:
I think it's a good idea, would be happy to review any PR if someone wants to implement it.

pmalekn:
Any progress? I'd love to see that as well

adzimzf:
Is this the same with [this one](https://github.com/sirupsen/logrus/issues/1048).?

pmalekn:
So it seems there's no magical solution to this.

Go documentation states the following:

> // The recover built-in function allows a program to manage behavior of a
> // panicking goroutine. Executing a call to recover inside a deferred
> // function (but not any function called by it) stops the panicking sequence
> // by restoring normal execution and retrieves the error value passed to the
> // call of panic. **If recover is called outside the deferred function it will
> // not stop a panicking sequence**. In this case, or when the goroutine is not
> // panicking, or if the argument supplied to panic was nil, recover returns
> // nil. Thus the return value from recover reports whether the goroutine is
> // panicking.
> func recover() interface{}

So the only option left is either to roll your own:

```go
defer func() {
	if r := recover(); r != nil {
		log.WithField("panic", r).Error("we panicked!")
	}
}()
```

or define a helper either in your code base or in `logrus`:

```go
func RecoverPanic() {
	if r := recover(); r != nil {
		log.WithField("panic", r).Error("we panicked!")
	}
}

func main() {
	defer RecoverPanic()
	// code...
}
```

PR #1066, opened 2019-11-06
feat: add panic recover for Issue #181

abc950309:
Add recover functions like raven client for Issue #181.


markphelps:
Thank you for your contribution! Unfortunately the project is being put into [maintenance mode](https://github.com/sirupsen/logrus/commit/4ecd9a62bd57b6533e0d87ae23a6b10adc32b5ce) which means we will not be adding new features.

PR #1072, opened 2019-11-28
Fix entry data bleed when using WithContext and WithTime

taywrobel:
Creates a copy of the `Data` map when using `WithContext` to create a
child Entry.  Without this, the data map of the parent Entry,
which is exposed in the Entry struct, is shared between a parent
and all children instances.

This can create bugs of shared or overwritten data when a parent
Entry is used to make children in differing contexts, and behaves
differently than `WithField` and its derivatives which does make
a copy of the `Data` map.

Additionally implements the same logic for `WithTime`, for API
consistency in behavior.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -103,7 +103,11 @@ func (entry *Entry) WithError(err error) *Entry {
 
 // Add a context to the Entry.
 func (entry *Entry) WithContext(ctx context.Context) *Entry {
-	return &Entry{Logger: entry.Logger, Data: entry.Data, Time: entry.Time, err: entry.err, Context: ctx}
+	dataCopy := make(Fields, len(entry.Data))
+	for k, v := range entry.Data {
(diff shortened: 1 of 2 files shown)


taywrobel:
@dgsb Looks like you're one of the more active maintainers here.  If this LGTY (or another maintainer) I can add an entry to the changelog if you'd like me to, or you're welcome to.

This change maintains API compatibility but technically breaks backwards compatibility if a user is reliant on the bug to share data amongst entities.  However, I don't think that's a significant enough risk to justify a major version release as strict semver would dictate, so I was thinking a patch release to 1.4.3 would be best?

taywrobel:
@freeformzSFDC Any chance you can review this and get it published?  We're needing to do a pretty questionable workaround for this right now, and having an upstream fix available would be 🍬

PR #1074, opened 2019-11-29
Add Audit log level

glb:
Audit logs are intended for security audits -- things you want to log all the time that are not errors.

Hoping other people will find this useful.


steakunderscore:
We are using logrus and I was about to implement pretty much exactly this. @sirupsen is this type of feature something that you would like to see in logrus?

I reviewed the change and I think a large portion of the minor changes could be split out in to separate PR so this PR is only concerning audit logs. Otherwise the change LGTM.

glb:
@steakunderscore thanks! Could you explain a bit more what you mean about splitting out the minor changes to a separate PR? There are a couple of things in the readme that were a result of vscode auto-formatting but otherwise I think this is the minimal change required to implement / test / document audit logging?

markphelps:
Closing per the project being put into [maintenance mode](https://github.com/sirupsen/logrus/commit/4ecd9a62bd57b6533e0d87ae23a6b10adc32b5ce)

PR #1075, opened 2019-12-06
Allow HTML escaping to be disabled in JSONFormatter

harrisonhjones:
This code change allows users of JSONFormatter to disable HTML escaping. A test was also added.


markphelps:
Fixed in #1113. Thanks!

harrisonhjones:
You are welcome. Thanks for merging.

PR #1088, opened 2020-01-12
Simplify checkIfTerminal for Windows

tklauser:
Instead of relying on `EnableVirtualTerminalProcessing` from
`github.com/konsorten/go-windows-terminal-sequences` which just calls
`GetConsoleMode`, sets `ENABLE_VIRTUAL_TERMINAL_PROCESSING` and calls
`SetConsoleMode` with the new modified mode, implement it directly inside
`checkIfTerminal`. This also avoids the duplicate call to `GetConsoleMode`
and gets rid of an external dependency.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,6 @@ module github.com/sirupsen/logrus
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/konsorten/go-windows-terminal-sequences v1.0.3
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/stretchr/testify v1.2.2
 	golang.org/x/sys v0.0.0-20190422165155-953cdadca894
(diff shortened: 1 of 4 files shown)


tklauser:
Rebased on latest master. Would be nice to get this merged in order to get rid of an additional dependency (which is unused on every platform but Windows). Is there anything else needed from my side? /cc @dgsb

thaJeztah:
nice! Was recently working on something similar (didn't open a PR for that one yet; thanks for reminding me 😂 https://github.com/moby/term/compare/master...thaJeztah:deprecate_isconsole?expand=1)

thaJeztah:
@dgsb ptal

dgsb:
hello we are indeed getting rid of a dependency but we get another instead which seems a way bigger package. What is the advantage of that ?

tklauser:
> hello we are indeed getting rid of a dependency but we get another instead which seems a way bigger package. What is the advantage of that ?

`github.com/sirupsen/logrus` already depends on `golang.org/x/sys` for all other platforms than Windows. So it would streamline this dependency for all platforms. Also, `x/sys/windows` is maintained by the Go project and the official replacement for the `syscall` package, see https://github.com/golang/go/blob/2b70ffe9307c0992e28513ba25081d767b5937b2/src/syscall/syscall.go#L21-L25, so it's well maintained.

I agree the dependency itself is bigger but the Go linker will strip away any unused symbols, so in terms of binary size we should only pay for functionality we actually use.

thaJeztah:
`golang.org/x/sys` is quite commonly used in projects (and, as mentioned, already a dependency for the linux builds), so from my perspective should not be a blocker.  As discussed with @tklauser in https://github.com/moby/term/pull/10#discussion_r421383584, implementing using `"syscall"` instead is possible (if there's concerns about adding this dependency for windows-only projects), but may not be "recommended" given that the `syscall` package is in maintenance mode.

dgsb:
Ok, makes sense

PR #1092, opened 2020-01-23
Properly close the writer and wait for all writes to finish

adieu:
I use `logger.Writer()` as `io.Writer` for the output of my program and found that sometimes when the program exit too fast, some output wouldn't be displayed even if I close the writer before exit.

After reading `writer.go` I found that when we call `writer.Close()` we don't wait for `entry.writerScanner()` which runs in another goroutine to finish.

This commit adds `WriterLevelWithClose` which returns a `closeFunc` to properly close the writer and reader.

Eventually we might have to modify `Writer` and `WriterLevel` to return the `closeFunc` too which is not backward compatible or we might wrap the writer and return an `io.WriteCloser` instead.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -49,10 +49,11 @@ func (entry *Entry) WriterLevel(level Level) *io.PipeWriter {
 	case PanicLevel:
 		printFunc = entry.Panic
 	}
+	done := make(chan struct{})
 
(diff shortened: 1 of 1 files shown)


medoraemon:
好好好

PR #1106, opened 2020-03-05
requesting merge with corrected updates

zachwylde00:
you guys code awesomely  by the way so thank you regardless of which way you decide.


dgsb:
Sorry, but what's the point of this MR ?

zachwylde00:
well I had a system going until circle.ci fizzled out on me that would
possibly allow for  A Hard Fork IN Bitcoin the data showed a simulation
event of a disregarded PNG with value and simulated using it, producing a
hard fork with a split value and 21000000 more bitcoin, considering India
accepting bitcoin into the banking system that wouldn't affect the price
for more than a couple of days possibly two weeks at most also the fork
would allow for a us interest considering 76% of what my data produces goes
to them via the SOCIALSECURITYADMINISTRATION for the future of our
country's elderly and incapable. but I must now find an equally affective
ci.  or build it myself.

PR #1107, opened 2020-03-05
Inline logger.Log[f]/Info[f] and other methods.

xaurx:
golang wasn't inlining any of logger.Logf/Log/Info/Infof/Warn/Warnf
methods.

Seems every call to the function adds nodes and compiler believes the
function is more heavier than it really is.
As a result I had to put check for log level right in the above
mentioned functions + replace logger.IsLevelEnabled() call with the
condition from it (yes, it affects result as well!).

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -140,24 +140,34 @@ func (logger *Logger) WithTime(t time.Time) *Entry {
 	return entry.WithTime(t)
 }
 
+func (logger *Logger) logf(level Level, format string, args ...interface{}) {
+	entry := logger.newEntry()
+	entry.Logf(level, format, args...)
+	logger.releaseEntry(entry)
+}
+
 func (logger *Logger) Logf(level Level, format string, args ...interface{}) {
-	if logger.IsLevelEnabled(level) {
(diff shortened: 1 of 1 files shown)


markphelps:
@xaurx would you mind posting the compiler output before and after you change to show the non-inlining and then inlining versions?

What do you mean by:
>  (yes, it affects result as well!)

Thanks!

xaurx:
> @xaurx would you mind posting the compiler output before and after your change to show the non-inlining and then inlining versions?

What compiler output you mean? I was looking at generated ASM code when worked on it.

> What do you mean by:
> > (yes, it affects result as well!)

I mean that if you take some of the functions which are not inlined by compiler and where logger.IsLevelEnabled() is used and replace the call to logger.IsLevelEnabled() with the very same if() statement from it - the functions in question will start to be inline'able.

Hope it is more clear now.

dgsb:
With this patch we are duplicating every where the level check. This is something which has already gone back and forth. I'm not sure the inlinable-ility is worth the loss in maintainability/readability.

Issue #954, opened 2019-04-17
Race Condition with SetReportCaller()


jeremyje:
It looks like there's a data race condition between `SetReportCaller` and internal reads from `entry.go`.

 https://github.com/sirupsen/logrus/blob/master/entry.go#L220 and https://github.com/sirupsen/logrus/blob/master/entry.go#L200 are 2 probable culprits here.

```
Step #8 - "Build: Test": ==================
Step #8 - "Build: Test": WARNING: DATA RACE
Step #8 - "Build: Test": Write at 0x00c00003c208 by goroutine 67:
Step #8 - "Build: Test":   github.com/sirupsen/logrus.(*Logger).SetReportCaller()
Step #8 - "Build: Test":       /go/pkg/mod/github.com/sirupsen/logrus@v1.4.1/logger.go:341 +0x85
Step #8 - "Build: Test":   github.com/GoogleCloudPlatform/open-match/internal/serving/testing.createOpenMatchServer()
Step #8 - "Build: Test":       /go/pkg/mod/github.com/sirupsen/logrus@v1.4.1/exported.go:31 +0xf6
Step #8 - "Build: Test":   github.com/GoogleCloudPlatform/open-match/internal/serving/testing.NewMiniMatch()
Step #8 - "Build: Test":       /workspace/internal/serving/testing/minimatch.go:65 +0x6d
Step #8 - "Build: Test":   github.com/GoogleCloudPlatform/open-match/test/e2e/frontend.TestFrontendStartup()
Step #8 - "Build: Test":       /workspace/test/e2e/frontend/frontendapi_test.go:13 +0x76
Step #8 - "Build: Test":   testing.tRunner()
Step #8 - "Build: Test":       /usr/local/go/src/testing/testing.go:865 +0x163
Step #8 - "Build: Test": 
Step #8 - "Build: Test": Previous read at 0x00c00003c208 by goroutine 62:
Step #8 - "Build: Test":   github.com/sirupsen/logrus.Entry.log()
Step #8 - "Build: Test":       /go/pkg/mod/github.com/sirupsen/logrus@v1.4.1/entry.go:220 +0x1c2
Step #8 - "Build: Test":   github.com/sirupsen/logrus.(*Entry).Log()
Step #8 - "Build: Test":       /go/pkg/mod/github.com/sirupsen/logrus@v1.4.1/entry.go:268 +0x160
[...]

PR #1116, opened 2020-03-19
Resolve race condition with SetReportCaller() and Entry

admacleod:
I came across this issue in a project I am working on and discovered that there was already an issue for it (#954).

I added a test to prove that the race was fixed and ensure that it doesn't come back.


markphelps:
Thank you @admacleod !

medoraemon:
good

medoraemon:
have a see

Issue #1105, opened 2020-03-05
Add function parameters


Azer0s:
Adding function parameters to logging can improve speed greatly. Let me give an example:

```go
type InputFunction func()[]interface{}

func (logger *Logger) LogFn(level Level, fn InputFunction) {
	if logger.IsLevelEnabled(level) {
		entry := logger.newEntry()
		entry.Log(level, (fn())...)
		logger.releaseEntry(entry)
	}
}
```

One would, of course, also have to write all the exposed methods (`InfoFn`, `DebugFn`, `TraceFn`, etc.).

The usage of this feature could look like so:

```go
log.TraceFn(func()[]interface{} {
	return []interface{}{strings.ReplaceAll(fmt.Sprintf("\n%s", hex.Dump(buf)), "\n", "\n\t")}
})
```

This function would only be called if the trace level is actually enabled.

Azer0s:
If you deem this feature sensible, I could implement it because it is something I actually really need and don't want to write as an abstraction **just** for my project.

PR #1127, opened 2020-04-10
Add loggers that take functions as arguments

Azer0s:
**This is connected to #1105** 

Having functions as log parameters can greatly improve performance in some cases.

E.g.: a big output is logged on trace level, when the current log-level is warn. Prior to the change, the log message would've always been generated. Now, we can pass a function and the log message is only generated when the corresponding log level is actually active. 

### Example

```go
log.TraceFn(func()[]interface{} {
	return []interface{}{strings.ReplaceAll(fmt.Sprintf("\n%s", hex.Dump(buf)), "\n", "\n\t")}
})
```

### DIFF
--- a/example_function_test.go
+++ b/example_function_test.go
@@ -1,0 +1,26 @@
+package logrus_test
+
+import (
+	"fmt"
+	log "github.com/sirupsen/logrus"
+	"testing"
+)
+
+func TestLogger_LogFn(t *testing.T) {
+	log.SetFormatter(&log.JSONFormatter{})
+	log.SetLevel(log.WarnLevel)
+
+	log.InfoFn(func() []interface{} {
+		fmt.Println("This is never run")
+		return []interface{} {
+			"Hello",
+		}
+	})
+
+	log.ErrorFn(func() []interface{} {
+		fmt.Println("This runs")
+		return []interface{} {
+			"Oopsi",
+		}
+	})
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 logrus
 vendor
+
+.idea/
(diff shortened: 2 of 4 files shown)


dgsb:
Hello @Azer0s. Thanks for your contribution, this looks indeed interesting.

Azer0s:
Hi @dgsb! Thanks! Do you think there are some further improvements to be made or is this ready for review?

dgsb:
@Azer0s it's ready to be merged, I'm just wondering the reason of the glitch with travis

Azer0s:
@dgsb Hmm...could you maybe restart the travis build?

Azer0s:
@dgsb Any updates on this?

example_function_test.go
```diff
@@ -0,0 +1,26 @@
...
+	log "github.com/sirupsen/logrus"
+	"testing"
+)
+
+func TestLogger_LogFn(t *testing.T) {
+	log.SetFormatter(&log.JSONFormatter{})
+	log.SetLevel(log.WarnLevel)
+
+	log.InfoFn(func() []interface{} {
+		fmt.Println("This is never run")
```
dgsb, line 14: Can we improve the test to actually ensure that the function is indeed not called ?

example_function_test.go
```diff
@@ -0,0 +1,26 @@
...
+		return []interface{} {
+			"Hello",
+		}
+	})
+
+	log.ErrorFn(func() []interface{} {
+		fmt.Println("This runs")
```
dgsb, line 21: Can we improve the test to actually ensure that the function is indeed called ?

dgsb:
@Azer0s I've added small comments on the testing part, do you have time to improve that ?

Issue #654, opened 2017-10-24
The newline characters in the log message are not processed correctly


ahrtr:
When there are new line characters or any other special characters are in the log message, then they are just displayed as "\n" directly in the log file.  For example,
time="2017-10-23T17:22:27+08:00" level=warning msg="hello1,\nhello2,\nhello3"

It isn't what I want. The reason is that the new lines characters are escaped by the doube quote.  So I'd suggest to enhance the text_formatter so that the user can decide whether to wrap the log message in double quote or not. 

Another suggestion is add a configuration item for text_formatter to control the key fields(i.e., time, level, msg) as well, so that the user can decide whether to display the leading key fields or not.  For example, the above example would be displayed as below if the user disables the leading key fields,
2017-10-23T17:22:27+08:00 warning hello1,
hello2,
hello3

Actually I have already fixed it in my forked repository as below,
https://github.com/ahrtr/logrus

jflyup:
This so called feature make me go back to std log.

qinze118:
@ahrtr hello,I also encountered this problem.i'm try to use func Println,but not resolved.how to resolved  with you ?

qinze118:
@ahrtr thanks!  I have solved the problem with your code

reedwade:
For what it's worth, it would rate as a bug to me if the current behaviour were to change as it would make the logs unusable except by the humans.

jinweizhiyuan:
> When there are new line characters or any other special characters are in the log message, then they are just displayed as "\n" directly in the log file. For example,
> time="2017-10-23T17:22:27+08:00" level=warning msg="hello1,\nhello2,\nhello3"
> 
> It isn't what I want. The reason is that the new lines characters are escaped by the doube quote. So I'd suggest to enhance the text_formatter so that the user can decide whether to wrap the log message in double quote or not.

How to solve this problem? Is there an example?

linrl3:
Could you try using `DisableQuote` for your logger like this?
```
formatter := &logrus.TextFormatter{}
formatter.DisableQuote = true
logrus.SetFormatter(formatter)
```
It works for me.

Issue #608, opened 2017-07-31
New lines are escaped now


pkopac:
Hi!
With a recent update I've noticed that my debug data dumps in logs get fully escaped. Previously, I had _nicely formatted_ data `structs` in output (using [spew](https://github.com/davecgh/go-spew)`.Sdump`), so I could actually read them.

**Code:**
* data is a complex data structure with arrays, pointers, many custom types, optional fields etc.
`log.Debug(spew.Sdump(data))`

And now I get:
`(*pkg.Data)(0xc423e8cb00)({\n UUID: (string) \"\",\n Property:`...

I've browsed through recent updates, but can't find a way to turn this off other then fixing version of logrus to some previous one. Please advise. The data dumps are unreadable now.

dmathieu:
This is probably because we hardened the string escaping with `%q` instead of the previous unsafe escaping.
See #580

I'm afraid there isn't much we can do here. This could actually be considered as a bug which has been fixed. Being able to send new lines in your log could make them harder to read, and actually reduce the purpose of structured logging, where you would only log specific meaningful fields not dump a whole json object.

JelteF:
This makes sense when logging to a log file, but when logging for development this is really annoying. Stack traces that are logged are now unreadable.

berniedurfee-ge:
Yes please on this one. Having a switch of some sort would be very nice.

himanshug:
Another use case I have where I am printing json serialized objects in the debug logs. In the printed json, all double quotes are escaped too making it totally unreadable.

reefbarman:
we have run into the need to print the stack traces in our logs as well and the removal of the new lines on stdout makes them very hard to read, is this going to be considered?

hmalphettes:
For human consumption, I am using https://github.com/antonfisher/nested-logrus-formatter which does not escape the text content.

For example:
```
Mar 18 09:32:26.740 [INFO]  [Content:## Generated by my-tool
# Read the policy
path "sys/policies/acl/a-policy" {
  "capabilities" = [ "read" ]
}

path "secret/*" {
  "capabilities" = ["read","list"]
}
path "auth/token/create/a-role" {
  "capabilities" = ["update"]
  "denied_parameters" = {"*"=[]}
}
] [Environment:dev]
```

I hope this helps.

nzt4567:
A switch would definitely be handy... 

[...]

PR #1134, opened 2020-04-23
Adds flag to disable quotes in TextFormatter

thlacroix:
Relates to https://github.com/sirupsen/logrus/issues/654 and https://github.com/sirupsen/logrus/issues/608

In some cases, we don't want the log values to be quoted or escaped, so this is adding a simple flag in TextFormatter to disable quotes (which also disables escaping) in the field values.

### DIFF
--- a/text_formatter.go
+++ b/text_formatter.go
@@ -37,6 +37,9 @@ type TextFormatter struct {
 	// Force quoting of all values
 	ForceQuote bool
 
+	// DisableQuote disables quoting for all values
+	DisableQuote bool
+
 	// Override coloring based on CLICOLOR and CLICOLOR_FORCE. - https://bixense.com/clicolors/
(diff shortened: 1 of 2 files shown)


dgsb:
What is the expected behaviour between forcequote and disablequote are both set to true ?

thlacroix:
`DisableQuote` has the least priority, so if `ForceQuote` is also set, quotes will be used. If `QuoteEmptyFields` is also set, quotes will be used with empty fields.

This way it impacts less the current logic, and by default works as before.

dgsb:
Can we have one more unit test to check this behaviour ?

thlacroix:
Of course. I've added a few test cases.

dgsb:
thanks for your contribution @thlacroix

PR #1136, opened 2020-04-28
Change CRLF line endings to LF

ialidzhikov:
Currently there is a file with CRLF line endings which is sometimes causing diff on projects vendoring `github.com/sirupsen/logrus`.


dgsb:
Thanks for your contribution @ialidzhikov

PR #1145, opened 2020-05-28
Add an API to plug a custom buffer free item mangement system

### LINKED ISSUE #1124: When the transient high traffic impacts, the memory usage of the internal buffer pool of Logrus is high.
In our production environment, when the transient high traffic impact occurred, we found that the memory occupied by the buffer pool inside Logrus could not be released quickly. 
```
// entry.go 246 - 255
...
	entry.fireHooks()

	buffer = bufferPool.Get().(*bytes.Buffer)
	buffer.Reset()
	defer bufferPool.Put(buffer)
	entry.Buffer = buffer

	entry.write()

	entry.Buffer = nil
...
```
Objects in the buffer pool are not cleaned up when released.

dgsb:
Fixes #1124

### DIFF
--- a/buffer_pool.go
+++ b/buffer_pool.go
@@ -1,0 +1,52 @@
+package logrus
+
+import (
+	"bytes"
+	"sync"
+)
+
+var (
+	bufferPool BufferPool
+)
+
+type BufferPool interface {
+	Put(*bytes.Buffer)
+	Get() *bytes.Buffer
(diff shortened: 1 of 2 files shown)


edoger:
@dgsb Thank you very much for your work. We have decided to test the grayscale after the next Logrus version is released. Thanks again ~

Issue #1149, opened 2020-06-01
log just for specific user


masoodbayati:
Hi, How can i have debug log just for specific user?

dgsb:
There is nothing in logrus for doing that, you need to handle that at package consumer side.

mohsensamiei:
Hi guys
In [this pull request](https://github.com/sirupsen/logrus/pull/1150)
I want to force some log entry to logging regardless of its level
So we can pass force for custom users

dgsb:
Why not using a specific logger with a different level for a specific user instead ?

mohsensamiei:
> Why not using a specific logger with a different level for a specific user instead ?
> We try to avoid changing the package public interface for specific narrow use case in order to reduce the maintenance load.

because with this approach we need to have if block for switch between loggers
but I got idea, we could have a function for switch between loggers 

thanks bro

dgsb:
> but I got idea, we could have a function for switch between loggers

Indeed, that's the way to go

PR #1150, opened 2020-06-01
feat: implement force logging entry

mohsensamiei:
Hi guys
As issue: [#1149](https://github.com/sirupsen/logrus/issues/1149)
I want to force some log entry to logging regardless of its level 
So we can pass force for custom users


dgsb:
We can use a different logger instance for this use case instead of changing the package public api

PR #1171, opened 2020-09-08
Reduce log output by logging into a buffer

abenettt:


hooks/test/test_test.go
```diff
@@ -33,6 +34,7 @@ func TestAllHooks(t *testing.T) {
 
 	hook = NewGlobal()
 
+	logrus.SetOutput(&bytes.Buffer{})
```
thaJeztah: It's better to use `io.Discard` for these if it's only to suppress the output. We should probably also restore the previous output (`t.Cleanup`) because this is modifying the global "stdLogger").


I was working on something similar to this PR; let me push those changes.

Issue #1158, opened 2020-07-20
Bug: Inconsistent levels when creating Entry from Logger


iamolegga:
Example: https://play.golang.org/p/zZODKE5p6SD

Is this a bug?

markphelps:
@iamolegga that does look odd, I added some more output here: https://play.golang.org/p/Xe6QQx1igjo

You can see the entry is actually logging at the right level, ie it doesnt output 'debug' level messages since the root logger is configured to log at 'info' and above... however when you call `entry.Level` it still outputs `level: panic` which is incorrect.

```
logger level: info, entry level: panic
time="2009-11-10T23:00:00Z" level=info msg="I should be printed" foo=bar
logger level: info, entry level: panic
```

So I'm inclined to say that yes, it is a bug in that `entry.Level` does not seem to report the actual 'inherited level' that it is logging at.

iamolegga:
I believe this isn't fixed yet

iamolegga:
not fixed

iamolegga:
not fixed

iamolegga:
not fixed

iamolegga:
not fixed

iamolegga:
not fixed

iamolegga:
not fixed

iamolegga:
not fixed

iamolegga:
not fixed

iamolegga:
not fixed

edoger:
@iamolegga The library is already inactive, maybe you should try another solution! Just like me.

PR #1174, opened 2020-09-14
fix Bug: Inconsistent levels when creating Entry from Logger #1158

randomSignal:
fix: Bug: Inconsistent levels when creating Entry from Logger #1158


dgsb:
This is an breaking behavioral change.

PR #1175, opened 2020-09-15
support caller manually

cgxxv:
Support caller for someone who wants to locate the runtime.caller precisely and has a customed log themself based on the logrus.


thaJeztah:
Looks like this had a bad rebase; closing in favor of https://github.com/sirupsen/logrus/pull/1215

- rebased version in https://github.com/sirupsen/logrus/pull/1538

PR #1186, opened 2020-10-03
Fix typo log.InfoLevel->logrus.InfoLevel

dmaciel:


README.md
```diff
@@ -311,7 +311,7 @@ that severity or anything above it:
 
 ```go
 // Will log anything that is info or above (warn, error, fatal, panic). Default.
-log.SetLevel(log.InfoLevel)
+log.SetLevel(logrus.InfoLevel)
```
thaJeztah, line 314: The code above (at line 268 https://github.com/sirupsen/logrus/pull/1186/files#diff-04c6e90faac2675aa89e2176d2eec7d8R268) uses an alias for logrus, so the logrus package is imported as "log"; 

```go
```

thaJeztah (changes requested):
I don't think this change is correct (see my comment)

README.md
dgsb: Indeed the change is incorrect

thaJeztah: No problem, thanks for contributing!

(I agree it's somewhat confusing because it's not in the same "block" in the readme)

PR #1205, opened 2020-11-20
Extending FieldLogger interface to support IsLevelEnabled and SetLevel

### LINKED ISSUE #1204: FieldLogger doesn't support IsLevelEnabled
Hi team, given that FieldLogger is the abstraction of Logger and Entry, can we also add this function in FieldLogger interface? Thanks!

koote:
This PR adds IsLevelEnabled function to FieldLogger interface, fix #1204

### DIFF
--- a/logrus.go
+++ b/logrus.go
@@ -168,14 +168,18 @@ type FieldLogger interface {
 	Fatalln(args ...interface{})
 	Panicln(args ...interface{})
 
-	SetLevel(level Level)
-	IsLevelEnabled(level Level) bool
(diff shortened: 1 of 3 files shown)


logrus.go
```diff
@@ -176,6 +176,12 @@ type FieldLogger interface {
 	// IsPanicEnabled() bool
 }
 
+type Ext2FieldLogger interface {
```
dgsb, line 179: I'd rather not have this new interface.
If a package user needs that, they can easily define it at use side

PR #1210, opened 2020-11-30
Uses spaces

moremay:


dgsb:
Hello @moremay thanks for your contribution, however this function can easily be implemented outside of logrus and its resulting string given back at log function.
The API of logrus is already a bit bloated and we are reluctant to add new configuration option.

PR #1215, opened 2020-12-29
Add WithCaller and WithCallerAt for overriding the caller detection

nelsonxb:
Seems like there's already a few PRs along these lines (e.g. #973, #1175)... but I forgot to check for those before implementing this 😅

Anyway, I think this approach should be more flexible and useful. I don't think the other approaches would _quite_ work for my purposes... or at least, they'd be incredibly awkward.

The new methods `Event.WithCaller`, `Logger.WithCaller` allow for setting the caller with an existing frame, and exists mostly as a Just-In-Case™. The other new methods `Event.WithCallerAt`, `Logger.WithCallerAt` are the main source of interest (at least for me), and make it simple to log a message attached to the caller of the function that needs to log. For example:

```go
func DoSomething() error {
  // ...
  handleResult(result, err)
}

func handleResult(result string, err error) {
  if err != nil {
    // ...
    logger.WithCallerAt(1, err) // The error happened in the calling function
  }

  // ...
}
```

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -105,7 +105,7 @@ func (entry *Entry) WithContext(ctx context.Context) *Entry {
 	for k, v := range entry.Data {
 		dataCopy[k] = v
 	}
-	return &Entry{Logger: entry.Logger, Data: dataCopy, Time: entry.Time, err: entry.err, Context: ctx}
+	return &Entry{Logger: entry.Logger, Data: dataCopy, Time: entry.Time, err: entry.err, Context: ctx, Caller: entry.Caller}
 }
 
 // Add a single field to the Entry.
@@ -139,7 +139,7 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 			data[k] = v
 		}
 	}
-	return &Entry{Logger: entry.Logger, Data: data, Time: entry.Time, err: fieldErr, Context: entry.Context}
+	return &Entry{Logger: entry.Logger, Data: data, Time: entry.Time, err: fieldErr, Context: entry.Context, Caller: entry.Caller}
 }
 
 // Overrides the time of the Entry.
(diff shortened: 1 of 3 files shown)


Sytten:
@sirupsen Can this be added, it is quite common when writing adapters that the caller is just wrong and currently it gets overriden if the adapter tries to set it manually.

nelsonxb:
Honestly, I doubt that this PR will get merged, as upstream logrus isn’t accepting new PRs. Though it’d be great if this PR could be left open regardless, so those who do need this feature can find it.

If you do wish to use my fork, you can still use the import of _github.com/sirupsen/logrus_, by adding a _replace_ statement to your go.mod file. This won’t work very well if you’re a library expecting to be imported, but it might help some users if they want it that badly.

In the meantime, I’ll be updating my fork tonight to the latest revision of upstream; and I’ll try to keep doing that every now and then. Logrus doesn’t change much (likely due to the whole maintenance mode thing), so it shouldn’t be too far out of date at any given time.

shallowclouds:
This is what I need!

But I think passing caller via `Entry` is enough, it should be avoided to imported new methods for `Entry` as `logrus` is not developing actively at present.

nelsonxb:
If any of the maintainers confirm that removing the WithCaller methods would get this merged, I’ll happily oblige.

But failing that, I feel that particularly WithCallerAt is far too useful to drop out-of-hand. Not only does it just do nothing if we don’t want the performance penalty, but it allows a caller to not have to fiddle around with reflect for a relatively simple use case.

edoger:
I really hope this feature can be merged! Feeling sleepy.

Lvzhenqian:
One year has passed, is there any news about this merger?  :(

LAShZ:
Is there any message about this commit or other similar ones? It's seems that I still have the problem of setting caller depths when use a wrapper of logrus :<

PR #1223, opened 2021-02-04
Fix Data map goroutine access

314159:


314159:
Oops.  I'm playing and I didn't mean to make a PR with you.  Sorry.

PR #1225, opened 2021-02-11
Improving syslog hook

aluvare:
The syslog library has been changed to prevent crashes if the syslog server is down, writes doesnt degrade if syslog server is slow.

### DIFF
--- a/hooks/syslog/syslog.go
+++ b/hooks/syslog/syslog.go
@@ -4,50 +4,95 @@ package syslog
 
 import (
 	"fmt"
-	"log/syslog"
+	//"log/syslog"
+	"github.com/GolangResources/syslog/syslog"
 	"os"
+	"time"
 
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	NUM_SYSLOG_WORKERS = 8
+	CHANNEL_DEPTH = 8192
+)
+
 // SyslogHook to send logs via syslog.
 type SyslogHook struct {
-	Writer        *syslog.Writer
-	SyslogNetwork string
-	SyslogRaddr   string
+	Writer		*syslog.Writer
+	SyslogNetwork	string
+	SyslogRaddr	string
+	Tag		string
+	syschan		chan *SEntry
+}
+
+type SEntry struct {
+	line		string
+	level		logrus.Level
 }
 
 // Creates a hook to be added to an instance of logger. This is called with
 // `hook, err := NewSyslogHook("udp", "localhost:514", syslog.LOG_DEBUG, "")`
 // `if err == nil { log.Hooks.Add(hook) }`
 func NewSyslogHook(network, raddr string, priority syslog.Priority, tag string) (*SyslogHook, error) {
-	w, err := syslog.Dial(network, raddr, priority, tag)
-	return &SyslogHook{w, network, raddr}, err
+	w, err := syslog.Dial("ctcp" , raddr, priority, nil)
+	if err != nil {
+		return nil, err
+	}
+	hook := &SyslogHook{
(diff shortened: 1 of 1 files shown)


hooks/syslog/syslog.go
```diff
@@ -4,50 +4,96 @@ package syslog
 
 import (
 	"fmt"
-	"log/syslog"
+	//"log/syslog"
+	"github.com/GolangResources/syslog/syslog"
```
dgsb, line 8: What does this package brings to the standard library one ?

hooks/syslog/syslog.go
```diff
@@ -4,50 +4,96 @@ package syslog
...
 }
 
 // Creates a hook to be added to an instance of logger. This is called with
 // `hook, err := NewSyslogHook("udp", "localhost:514", syslog.LOG_DEBUG, "")`
 // `if err == nil { log.Hooks.Add(hook) }`
-func NewSyslogHook(network, raddr string, priority syslog.Priority, tag string) (*SyslogHook, error) {
-	w, err := syslog.Dial(network, raddr, priority, tag)
-	return &SyslogHook{w, network, raddr}, err
+func NewSyslogHook(network, raddr string, priority sg.Priority, tag string) (*SyslogHook, error) {
+	w, err := syslog.Dial("ctcp" , raddr, priority, nil)
```
dgsb, line 39: What is `ctcp` ?

hooks/syslog/syslog.go
```diff
@@ -4,50 +4,96 @@ package syslog
...
+	if err != nil {
+		return nil, err
+	}
+	hook := &SyslogHook{
+		w,
+		network,
+		raddr,
+		tag,
+		make(chan *SEntry, CHANNEL_DEPTH),
```
dgsb, line 48: Could the channel creation be configurable ?

We need to be able to keep a compatible behaviour with the former version.

hooks/syslog/syslog.go
```diff
@@ -4,50 +4,96 @@ package syslog
...
+	}
+	for i := 0; i <= NUM_SYSLOG_WORKERS; i++ {
+		go hook.worker(i)
```
dgsb, line 51: We need to be able to stop the started go routine and to wait for them.

hooks/syslog/syslog.go
```diff
@@ -4,50 +4,96 @@ package syslog
...
+		return nil, err
+	}
+	hook := &SyslogHook{
+		w,
+		network,
+		raddr,
+		tag,
+		make(chan *SEntry, CHANNEL_DEPTH),
+	}
+	for i := 0; i <= NUM_SYSLOG_WORKERS; i++ {
```
dgsb, line 50: Same here we need this hook to be configurable so that we can keep the former behaviour.

hooks/syslog/syslog.go
```diff
@@ -4,50 +4,96 @@ package syslog
...
+		go hook.worker(i)
+	}
+	return hook, err
+}
+
+func (hook *SyslogHook) worker(i int) {
+	for entry := range hook.syschan {
+		time.Sleep(1)
```
dgsb, line 58: Why is there a sleep here ?

hooks/syslog/syslog.go
```diff
@@ -4,50 +4,96 @@ package syslog
...
+		func() error {
```
dgsb, line 59: What is the purpose of the closure ?

dgsb:
hello @aluvare thanks for your contribution.
I've done some comments/questions on your code changes.
We also need passing tests.

Issue #1217, opened 2021-01-20
Add a way for hooks to mutate/write entry Data fields


epelc:
Hooks should have a way to safely(re goroutine safe) write/update/add fields to entries. There are several issues stemming from the data field not having a lock and hooks not having any way to mutate fields.

One option is to switch from using a hook to using a formatter but you cannot easily wrap formatters while safely modifying fields. ie you cannot directly modify Data due to race conditions, you cannot use `entry.WithFields` then pass to say json formatter because that screws up entry release/locks/buffer usage, etc.

# Use Case

Our use case is a simple hook which adds a hostname to every log entry to help differentiate which server a log came from.

# Potential solutions

All current solutions require forks/minor changes if you are using the JSON formatter due to internal fields. They do not all touch this though, only the use a formatter option below(2nd option).

## Improve Hooks API(best)

Adding a way for hooks to return a new entry to replace the current one would allow easily modifying/adding new fields. This would probably have the lowest api impact and be the easiest to implement. Hooks could optionally implement a `MutatbleHook` interface which would take precident over the existing `Fire`. ie something like `FireAndMutate(*Entry) (*Entry, error)` where the returned entry is passed onto the next hook and used if non-nil. If any hook returned a new entry then it could be used in place of the current one.

## Use a formatter instead(alternative, also very good if formatters could be wrapped easily while modifying entries)

The only real current option around is to modify the json formatter or your target formatter to directly include your new/modified/updated field logic. Even doing this hardly works and requires a fork of the repo due to the `Entry.err` field being private.

The main issue with this is that `Entry.WithField` does not create a full copy(ie the error, buffer, and potentially other fields are not copied). It also has issues with releasing entries when calling from within a formatter. I did not fully investigate this although a simple option is to modify the underlying formatter(ie JSON) itself to simply add the field before sending to `json.Encoder.Encode`. This side steps all the issues/perf with creating entry copies/modifications.

## Add Data locks

Another option is to add a mutex around the Data field so that hooks could safely modify them.

See the following related issues.
https://github.com/sirupsen/logrus/issues/1046

https://github.com/sirupsen/logrus/issues/953
https://github.com/vmware-tanzu/velero/issues/599

epelc:
I've added a PR for a solution which we are using. It happens to be the simplest to implement and also the most performant but requires updating each formatter to support it. Luckily I believe most people use the JSON formatter.

[...]

Issue #1193, opened 2020-10-29
fireHooks lock system and became very lag.(open too many file)


Detail0805:
![abcd](https://user-images.githubusercontent.com/29973842/97578512-ddadc180-1a2b-11eb-89bf-03d52e995290.jpg)

I am using ELK and make a tcp connection to my logstash.

Today my ELK is crash down and I found my server will slowly by this code.

When logrus try to do this:

               fmt.Fprintf(os.Stderr, "Failed to fire hook: %v\n", err)

My linux file is open too much.

I know I can rewrite fire and levels function make it return nil.

I just 4-5 log / second.

dgsb:
We need more context to understand what your problem is.

Detail0805:
> We need more context to understand what your problem is.
> Which hooks are you using, a code sample, etc ...

The code is in the entry.go 259 line (version tag v1.7.0).

My scenario is when elk is disconnected, this code will cause my operating system to open the file repeatedly.

Because ELK log is very huge. The file will open to many and let server crash down.

I think no one thought that the log server would be disconnected.

My suggestion is to use rate limit algorithm to solve.

dgsb:
We have changed the locking system in v1.8.1

Issue #953, opened 2019-04-17
race condition with mutating hooks & multiple goroutines


skriss:
It looks like there's still a race condition when you have hooks attached to your logger that add entries to the `Data` map, and you have multiple goroutines logging to the same `*Entry` concurrently.  I can intermittently reproduce this with a simple hook that adds a field to the entry, and 10 goroutines all logging to the same `*Entry`.

I think what's going on is -- the `log` method for `Entry` has a non-pointer receiver, with a comment indicating this is intended to prevent race conditions (https://github.com/sirupsen/logrus/blob/master/entry.go#L204). However, the `Entry` struct has a field, `Data`, which is a `map`. So when the `log` method is called multiple times concurrently, `Entry` is passed by value due to it being a non-pointer receiver, but all of the different copies of `Entry` still have a reference to the same `Data` map (since map is a reference type in go).

If this analysis is correct, one way this could be solved is by changing `Entry`'s `Log` function to make a copy of `Entry`, including a deep copy of the `Data` map, before calling `log`. I'm not familiar enough with the code to know if this would be the best approach or not, looking for input.

dgsb:
This should be fixed in v1.8.1

xieyuschen:
Note that the code line mentioned by @skriss could be here: https://github.com/sirupsen/logrus/blob/9b3cdde74fbe9443d704467498a7dcb61a79de9b/entry.go#L206. The link pointed by skriss is unuseful as the code commits increasing.

PR #1229, opened 2021-02-16
fix for entry data field race condition

dgsb:


dgsb:
Fixes #1217, #1046, #1193, #953

entry.go
```diff
@@ -212,68 +224,68 @@ func (entry Entry) HasCaller() (has bool) {
 
 // This function is not declared with a pointer value because otherwise
```
epelc, line 225: This comment should be updated since it's a pointer method now.

Is it safe to use via pointer?

entry.go
```diff
@@ -212,68 +224,68 @@ func (entry Entry) HasCaller() (has bool) {
...
-	defer entry.Logger.mu.Unlock()
 	serialized, err := entry.Logger.Formatter.Format(entry)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to obtain reader, %v\n", err)
 		return
 	}
-	if _, err = entry.Logger.Out.Write(serialized); err != nil {
-		fmt.Fprintf(os.Stderr, "Failed to write to log, %v\n", err)
-	}
+	func() {
```
epelc, line 282: This extra function wrapper isn't required here since the write does nothing else after writing. ie the lock doesn't require it.

entry.go
dgsb: Yes it is safe, the comment should indeed be updated.

entry.go
dgsb: Yes it's a leftover from temporary step, I will remove it

epelc:
@dgsb also noticed the `Entry.Dup()` method is public. I know you just did a minor release but noticed in a comment on other issue you said there were no api changes.

Not sure if this was intentionally left public or not. Might be soon enough to catch if it wasn't intentional.

dgsb:
Good catch we should have a bump on the minor bump, thanks for your thorough review 👍

entry.go
```diff
@@ -78,8 +78,20 @@ func NewEntry(logger *Logger) *Entry {
 	}
 }
 
+func (entry *Entry) Dup() *Entry {
+	data := make(Fields, len(entry.Data))
+	for k, v := range entry.Data {
+		data[k] = v
+	}
+	return &Entry{Logger: entry.Logger, Data: data, Time: entry.Time, Context: entry.Context, err: entry.err}
```
shallowclouds, line 86: Why not copy `caller` into new entry? There are some scenarios where I want to pass the caller by myself, for example, many library accept a logger interface and use it to output internal log(such as SQL exec log for [`Gorm`](https://github.com/go-gorm/gorm/blob/26dd4c980a62d47c990a05da9e5566bff3b2b00c/logger/logger.go#L52)), so many users wrap logrus to the needed interface, but the caller got by `logrus` is inaccuracy, and I get the caller and pass it into `Entry`, and this change breaks my wrapped logger for these library.

Can I send a PR for copying caller in `Entry.Dup` util? What do you think of it? Thanks! @dgsb @epelc

entry.go
epelc: Just wanted to point out since I was checking back through this that it's safe because of the `entry.Dup()` call in the new version.

Hooks are now safe to modify entry because any log call on entry duplicates the entire entry.

Issue #1233, opened 2021-02-17
Data Race in Hooks


Tiernan-Stapleton:
I work with a service that uses logrus extensively.
Our test suite has encountered a race condition in the latest version of logrus (v1.7.1)

The test output is as follows:

```
$ go test -race ./...
==================
WARNING: DATA RACE
Write at 0x00c000012240 by goroutine 11:
  runtime.mapassign_fast32()
      /usr/local/go/src/runtime/map_fast32.go:92 +0x0
  github.com/sirupsen/logrus.LevelHooks.Add()
      /home/tiernan/code/go-logrus-hook/vendor/github.com/sirupsen/logrus/hooks.go:20 +0x195
  github.com/sirupsen/logrus.(*Logger).AddHook()
      /home/tiernan/code/go-logrus-hook/vendor/github.com/sirupsen/logrus/logger.go:369 +0xbe
  github.com/sirupsen/logrus.AddHook()
      /home/tiernan/code/go-logrus-hook/vendor/github.com/sirupsen/logrus/exported.go:51 +0xaa
  github.com/sirupsen/logrus/hooks/test.NewGlobal()
      /home/tiernan/code/go-logrus-hook/vendor/github.com/sirupsen/logrus/hooks/test/test.go:25 +0x74
  github.com/hpe-hcss/go-logrus-hook/pkg/main_test.TestLog.func1()
      /home/tiernan/code/go-logrus-hook/pkg/main/log_output_test.go:19 +0x59
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1054 +0x1eb

Previous read at 0x00c000012240 by goroutine 10:
  runtime.mapaccess1_fast32()
      /usr/local/go/src/runtime/map_fast32.go:12 +0x0
  github.com/sirupsen/logrus.LevelHooks.Fire()
      /home/tiernan/code/go-logrus-hook/vendor/github.com/sirupsen/logrus/hooks.go:27 +0x50
  github.com/sirupsen/logrus.(*Entry).fireHooks()
      /home/tiernan/code/go-logrus-hook/vendor/github.com/sirupsen/logrus/entry.go:270 +0xb6
  github.com/sirupsen/logrus.(*Entry).log()
      /home/tiernan/code/go-logrus-hook/vendor/github.com/sirupsen/logrus/entry.go:247 +0x24a
  github.com/sirupsen/logrus.(*Entry).Log()
[...]

PR #1234, opened 2021-02-17
fix race condition AddHook and traces

dgsb:


dgsb:
Fixes #1233

entry.go
```diff
@@ -261,7 +261,15 @@ func (entry *Entry) log(level Level, msg string) {
...
 
 func (entry *Entry) fireHooks() {
-	err := entry.Logger.Hooks.Fire(entry.Level, entry)
+	var tmpHooks LevelHooks
+	entry.Logger.mu.Lock()
+	tmpHooks = make(LevelHooks, len(entry.Logger.Hooks))
+	for k, v := range entry.Logger.Hooks {
+		tmpHooks[k] = v
+	}
+	entry.Logger.mu.Unlock()
```
dolmen, line 270: This could be simplified:
```go
entry.Logger.mu.Lock()
tmpHooks := append(LevelHooks(nil), entry.Logger.Hooks...)
entry.Logger.mu.Unlock()
```

dgsb: it may lead to more reallocation and copy depending on the array size though.

dgsb: or not

dgsb: 🤷‍♂️

PR #1236, opened 2021-02-19
format http.Request field in JSONFormatter

maksim-paskal:
Fix json.Marshal error while printing http.Request in field, using JSONFormatter
```
Failed to obtain reader, Failed to marshal fields to JSON, json: unsupported type: func()
```
Common pattern of usage logrus in web application is log messages with request context - but JSONFormatter can't marshal http.Request - and usage http.Request type in JSONFormatter is imposible - for usage logging, application must implement some formatter before adding http.Request field in logger
```go
package main

import (
	"net/http"

)

func main() {
	log.SetFormatter(&log.JSONFormatter{})

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		log.WithField("request", r).Info("some info")
		// some logic
		w.Write([]byte("ok"))
	})
	http.ListenAndServe(":8080", nil)
}
```
I suggest use simple http.Request formatter before json marshaling - its will simplify usage logging in web apps

### DIFF
--- a/http_request_formater.go
+++ b/http_request_formater.go
@@ -1,0 +1,27 @@
+package logrus
+
+import (
+	"net/http"
+	"net/url"
+)
+
+type httpRequestType struct {
+	URL        *url.URL
+	Method     string
+	Header     http.Header
+	RemoteAddr string
+	RequestURI string
+	Host       string
+}
+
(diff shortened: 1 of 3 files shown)


maksim-paskal:
stale, rework

PR #1239, opened 2021-02-19
CI: use GitHub Actions

thaJeztah:
Use GitHub actions to run golang-ci-lint, cross, and test.

The "Test()" target in Mage was a plain "go test -v ./...", and should be
portable to other CI systems if needed; running it through the mage file
effectively resulted in "compile a go binary to run go".

The "Lint()" target in Mage relied on Travis CI setting up Golang-CI lint
before it was executed, which required bash. Moving it to GitHub actions
also allowed it to be run on Windows. Golang CI can still be run locally
either through the mage file (which is kept for now), or running
`golangci-lint run ./...` after installing golangci-lint.

The "CrossBuild() Mage target is still used to perform cross-compile, as it
contains code to generate the GOOS/GOARCH matrix, which makes it easier
to run locally.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,61 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+
+  lint:
+    name: Golang-CI Lint
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.platform }}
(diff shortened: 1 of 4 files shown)


thaJeztah:
I opened https://github.com/thaJeztah/logrus/pull/1 to test this (GH Actions don't run on a repository until merged).
Checks can be seen here; https://github.com/thaJeztah/logrus/pull/1/checks

dgsb:
@thaJeztah let's merge that and see how it fares versus travis

thaJeztah:
Thanks! Feel free to "ping" me in case there's issues 👍

Issue #1240, opened 2021-02-23
change type Fields map[string]interface{} into type Fielder interface{}


ethantsien:
when i wrap the logrus,i must implement all methods, because the Fields parameter,so i suggest if we can change the type Fields into type Fielder interface{}

dgsb:
Hello @ethantsien we cannot integrate your requested change, this is an API breaking change

PR #1241, opened 2021-02-23
add Fielder

ethantsien:
#1240


dgsb:
Hello @ethantsien, thank for your contribution, but we cannot merge because that this is not a backward compatible changes.

We avoid having, such breaking changes which would raise a major version bump which prevents users to automatically bump their dependencies versions.

PR #1249, opened 2021-03-18
Change godoc badge to pkg.go.dev badge

injustease:
Even tho [godoc](https://godoc.org/github.com/sirupsen/logrus) is already redirect to [pkg.go.dev](https://pkg.go.dev/github.com/sirupsen/logrus), it's still nice to have badge link direct to [pkg.go.dev](https://pkg.go.dev/github.com/sirupsen/logrus).


dgsb:
@injustease thanks for your contribution

PR #1250, opened 2021-03-18
refactor(): Make interfaces smaller and use composition

obalunenko:
Bigger interface - weaker abstarction.


dgsb:
@obalunenko thanks for your contribution, however we cannot merge that.
I really think exposing these public interfaces has been a mistake.
As interfaces are implicitely implemented in go, most of the time interfaces are declared at usage side.
I'd rather not add new interfaces we will have to support.

Issue #1247, opened 2021-03-16
Can we make BufferPool bind to Logger?


edoger:
Sometimes we need to specify different buffer pools for different loggers. The current version makes it impossible for us to do so!

like: 
```
Logger.SetBufferPool(pool BufferPool)
```

The advantage of this is that loggers for different purposes can use different buffers.
```
type LimitBufferPool struct {
    pool *sync.Pool
    cap int
}

func (p *LimitBufferPool) Put(buf *bytes.Buffer) {
    if buf.Cap() > p.cap {
        return
    }
    p.pool.Put(buf)
}

func (p *LimitBufferPool) Get() *bytes.Buffer {
    return p.pool.Get().(*bytes.Buffer)
}
```

dgsb:
That could be practical indeed. The trick though is too make the implementation so that it doesn't break application which does not set this field upon static declaration of a logger.
A pull request would be most welcome.

edoger:
I'll implement it by the end of the weekend.

edoger:
link #1253

edoger:
The feature has been implemented.

PR #1253, opened 2021-04-20
Add support for the logger private buffer pool.

edoger:
Add a private buffer pool for the logger. By default, the default global buffer pool will be used.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -232,6 +232,7 @@ func (entry *Entry) log(level Level, msg string) {
 
 	newEntry.Logger.mu.Lock()
 	reportCaller := newEntry.Logger.ReportCaller
+	bufPool := newEntry.getBufferPool()
 	newEntry.Logger.mu.Unlock()
 
 	if reportCaller {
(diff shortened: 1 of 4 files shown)


edoger:
link #1247

entry.go
```diff
@@ -260,6 +261,13 @@ func (entry *Entry) log(level Level, msg string) {
 	}
 }
 
+func (entry *Entry) getBufferPool() (pool BufferPool) {
+	if entry.Logger.BufferPool != nil {
```
dgsb, line 265: Can we have that protected by the logger mutex ?

dgsb (changes requested):
Thanks for your contribution @edoger

I just have a small comment regarding the missing lock when getting the buffer pool from the logger.

entry.go
edoger: The only place to call this method is already protected by a mutex.

dgsb: ok, I missed that 👍

dgsb (approved):

PR #1257, opened 2021-05-17
Add log level output support.

edoger:
This patch provides the function of writing different levels of logs to different outputs, and the goal is to solve the [#403](https://github.com/sirupsen/logrus/issues/403) issue. The patch remains backward compatible.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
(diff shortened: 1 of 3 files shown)


edoger:
@dgsb The part of the test failure does not seem to be caused by this patch. Can you help restart a failed test?

edoger:
The part of the test that failed seems to have found an existing bug.

PR #1271, opened 2021-08-04
Add in preprocessing hook to add additional key/value pairs

johnnyb:
We need to add additional key/value pairs to the log message for our logging flow.  Namely we want to auto-add the app name.  Anyway, this is a generic preprocessing hook that allows any modification to the data right before it is shipped to the JSON formatter.


Sonaing88661:
johnnyb:master

PR #1273, opened 2021-08-11
fix #1272

lwydyby:
https://github.com/sirupsen/logrus/issues/1272

Issue #1272, opened 2021-08-11
When modifying hook concurrently, there is an exception in map processing


lwydyby:
gist: https://gist.github.com/lwydyby/58700d6ea1082d25ecc46049633243c7
error log:
````
2021-08-11 10:13:18 [19 ] [info ] [1233] :0 test why 2
2021-08-11 10:13:18 [18 ] [info ] [123] :0 test why 
2021-08-11 10:13:18 [19 ] [info ] [1233] :0 test why 2
2021-08-11 10:13:18 [18 ] [info ] [123] :0 test why 
2021-08-11 10:13:18 [19 ] [info ] [1233] :0 test why 2
2021-08-11 10:13:18 [18 ] [info ] [123] :0 test why 
fatal error: concurrent map iteration and map write

goroutine 19 [running]:
runtime.throw(0x10e42bf, 0x26)
        /usr/local/go/src/runtime/panic.go:1117 +0x72 fp=0xc0001b9b80 sp=0xc0001b9b50 pc=0x10334b2
runtime.mapiternext(0xc0001b9cb0)
        /usr/local/go/src/runtime/map.go:858 +0x54c fp=0xc0001b9c00 sp=0xc0001b9b80 pc=0x100f64c
github.com/sirupsen/logrus.(*Entry).fireHooks(0xc00019e2a0)
        /Users/liwei/GolandProjects/pkg/mod/github.com/sirupsen/logrus@v1.8.1/entry.go:267 +0x148 fp=0xc0001b9e10 sp=0xc0001b9c00 pc=0x10ae268
github.com/sirupsen/logrus.(*Entry).log(0xc00019e070, 0x4, 0xc00019a100, 0xa)
        /Users/liwei/GolandProjects/pkg/mod/github.com/sirupsen/logrus@v1.8.1/entry.go:241 +0x119 fp=0xc0001b9e80 sp=0xc0001b9e10 pc=0x10adf39
github.com/sirupsen/logrus.(*Entry).Log(0xc00019e070, 0x4, 0xc0001b9ef0, 0x1, 0x1)
        /Users/liwei/GolandProjects/pkg/mod/github.com/sirupsen/logrus@v1.8.1/entry.go:293 +0x86 fp=0xc0001b9eb8 sp=0xc0001b9e80 pc=0x10ae726
github.com/sirupsen/logrus.(*Entry).Logf(0xc00019e070, 0xc000000004, 0x10deedb, 0xa, 0x0, 0x0, 0x0)
[...]

PR #1276, opened 2021-09-01
fix: runtime panic (via updating dependency)

butuzov:
This PR updates x/sys library so logrus can be used on Darwrin/go1.17
see: https://github.com/golang/go/issues/45702

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/sirupsen/logrus
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/stretchr/testify v1.7.0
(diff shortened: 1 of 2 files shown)


butuzov:
This is related to #1272  and has an alternative solution as #1273

thaJeztah:
Perhaps someone could add go 1.17 to CI


In general, I think this is fine to update; that said, if your project uses Logrus and runs into this issue, you should be able to specify a more recent version of golang.org/x/sys in your project's `go.mod` (if needed with `// indirect`). This will force go modules to get a more current version than the minimum version required in the logrus go.mod

butuzov:
That's what I did atm, however, isn't a better solution is to fix the issue at the root?

dgsb:
duplicate of #1280

PR #1309, opened 2022-01-26
Provide a concurrency safe way to read the logger's IO writer.

vtsao:


vtsao:
Can I get somebody to please take a look at this? Thanks.

PR #1321, opened 2022-03-22
gopherjs: fix build with js and solaris build tags

paralin:
Fix the solaris build tag, and add the gopherjs tag to the terminal_check file.

This fixes the build with "gopherjs build -v"


paralin:
@sirupsen These are a couple very minor changes but they fix the build on gopherjs - I've been testing these commits for the last couple years now.

paralin:
... nevermind, turns out these are no longer needed. gopherjs sets the "js" build tag properly now.

PR #1330, opened 2022-05-25
level as int

buddhamagnet:


dgsb:
We cannot merge that, this is an unexpected behavioral change

Issue #1336, opened 2022-06-06
New CVE was discovered CVE-2022-28948


izhakmo:
https://nvd.nist.gov/vuln/detail/CVE-2022-28948

https://github.com/go-yaml/yaml/releases/tag/v3.0.1

link to issue : https://github.com/go-yaml/yaml/issues/666

chkp-rigor:
Hi @dgsb @sirupsen,
It's a security issue (even if potential) that code scanners show. Can you please prioritize this PR?
Thanks in advance! And thank you for the logrus!

PR #1337, opened 2022-06-06
update gopkg.in/yaml.v3 to v3.0.1

izhakmo:
https://nvd.nist.gov/vuln/detail/CVE-2022-28948

https://github.com/go-yaml/yaml/releases/tag/v3.0.1

link to issue : https://github.com/go-yaml/yaml/issues/666

fixes https://github.com/sirupsen/logrus/issues/1336


thaJeztah:
Curious; how was this PR created, because it's updating `go.sum` without any changes in `go.mod`; from what I can see the is an indirect dependency, so not immediately sure how this would relate to logrus itself; perhaps another dependency that causes it to be in the dependency tree could be updated though

sio4:
https://github.com/sirupsen/logrus/pull/1344 includes this reverted change by updating testify.
I didn't check the CVE and the CVE may not affect logrus directly, but it could be better to update it.

thaJeztah:
Yeah, looks like it's only used in test-code; 

```bash
go mod graph | grep ' gopkg.in/yaml.v3'
github.com/stretchr/testify@v1.7.0 gopkg.in/yaml.v3@v3.0.0-20200313102051-9f266ea9e77c
```

Which only uses it in `assert.YAMLEq()` (which appears unused in the codebase); https://github.com/stretchr/testify/blob/v1.7.0/assert/assertions.go#L1543-L1559

But updating testify to 1.7.2 or above (with https://github.com/stretchr/testify/commit/41453c009af9a942261b7a25a88521d0d6804e7f) should update it.

PR #1343, opened 2022-07-19
update dependencies

dgsb:


sio4:
It seems like this PR reverted https://github.com/sirupsen/logrus/pull/1337 which is related to CVE. Is it an intentional change?

thaJeztah:
https://github.com/sirupsen/logrus/pull/1337 looks odd, as it didn't update `go.mod`, so shouldn't have affected `go.sum`. Given that it's an indirect dependency, I don't see what the immediate responsibility is of logrus to update it though 🤔

sio4:
Agree. will file a PR for it soon.

PR #1356, opened 2022-11-01
update comment

olegromanchuk:
A custom log formatter doesn't work with this info from [formatter.go](https://github.com/sirupsen/logrus/blob/master/formatter.go). Used entry.Time, entry.Level, entry.Message instead.
```
// * `entry.Data["msg"]`. The message passed from Info, Warn, Error ..
// * `entry.Data["time"]`. The timestamp.
// * `entry.Data["level"]. The level the entry was logged at.
```


thaJeztah:
- opened https://github.com/sirupsen/logrus/pull/1572 tot clarify the behavior

PR #1364, opened 2022-12-29
Support GNU/Hurd

sthibaul:


terminal_check_bsd.go
```diff
@@ -1,4 +1,4 @@
-// +build darwin dragonfly freebsd netbsd openbsd
+// +build darwin dragonfly freebsd netbsd openbsd hurd
```
AkihiroSuda, line 1: > `terminal_check_bsd.go`

Looks like the filename now should be changed?

sthibaul: Actually, no: the hurd libc API is largely very bsd-ish, it even defines BSD

PR #1372, opened 2023-02-17
Add instructions to use different log levels for local and syslog

### LINKED ISSUE #1369: logrus_syslog ignoring log level
Following the docs at https://github.com/sirupsen/logrus/#hooks I'm trying to use an external syslog server to send only WARNING (and above) logs there, while sending debug to local logs. Unfortunately it doesn't seem to work. The external syslog server receives whatever I setup with `log.SetLevel()` instead of the priority argument passed to `logrus_syslog.NewSyslogHook`.

Am I doing anything wrong here? Maybe I misinterpreted the docs?

```go
package main

import (
	logrus_syslog "github.com/sirupsen/logrus/hooks/syslog"
)

func main() {
	log.SetLevel(log.DebugLevel)

	hook, err := logrus_syslog.NewSyslogHook("tcp", "localhost:5140", syslog.LOG_WARNING, "test")
	if err != nil {
		panic(err)
	}
	log.AddHook(hook)

	// These 2 should go both to local and remote logs
	log.Warn("ciao with warn")
	log.Error("ciao with error")
	
	// These should only go to local logs and not to external syslog,
	// but instead they're shown there too :(
	log.Debug("ciao with debug")
	log.Info("ciao with info")
}
```

tommyblue:
This PR adds instructions to the syslog readme about how to send different log levels to local logging (`log.SetLevel`) and syslog hook.

The related issue #1369 explains the issue. I didn't change the hook behaviour to avoid disrupting users implementations. Adding instructions to the readme seems a fair solution to help people having the same issue I had.

fixes #1369

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ the last thing you want from your Logging library (again...).
 
 This does not mean Logrus is dead. Logrus will continue to be maintained for
 security, (backwards compatible) bug fixes, and performance (where we are
-limited by the interface). 
+limited by the interface).
 
 I believe Logrus' biggest contribution is to have played a part in today's
 widespread use of structured logging in Golang. There doesn't seem to be a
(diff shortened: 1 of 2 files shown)


tommyblue:
@sirupsen any chance you could take a look at the pr?

Issue #1370, opened 2023-02-02
Error while reading from Writer: bufio.Scanner: token too long


kke:
This happens if you try to log something that is longer than 64kb without newlines through something like `cmd.Stdout = logrus.Writer()`.

This can cause a denial of service in some cases.

cubxxw:
<img src="https://media2.giphy.com/media/d9RbxjZ8QXesiYoerE/giphy.gif"/>

ozfive:
The way to do this would be to modify func (entry *Entry) writerScanner(reader *io.PipeReader, printFunc func(args ...interface{})) in writer.go with a custom split function to split the input into chunks of up to 64kb

```go
// writerScanner scans the input from the given reader and writes it to the logger using the specified print function.
// It splits the input into chunks of up to 64KB to avoid buffer overflows.
func (entry *Entry) writerScanner(reader *io.PipeReader, printFunc func(args ...interface{})) {
	scanner := bufio.NewScanner(reader)

	// Set the buffer size to the maximum token size to avoid buffer overflows
	scanner.Buffer(make([]byte, bufio.MaxScanTokenSize), bufio.MaxScanTokenSize)

	// Define a split function to split the input into chunks of up to 64KB
	chunkSize := 64 * 1024 // 64KB
	splitFunc := func(data []byte, atEOF bool) (int, []byte, error) {
		if len(data) > chunkSize {
			return chunkSize, data[:chunkSize], nil
		}
		return 0, nil, nil
	}

	// Use the custom split function to split the input
	scanner.Split(splitFunc)

	// Scan the input and write it to the logger using the specified print function
	for scanner.Scan() {
		printFunc(strings.TrimRight(scanner.Text(), "\r\n"))
	}

	// If there was an error while scanning the input, log an error
	if err := scanner.Err(); err != nil {
		entry.Errorf("Error while reading from Writer: %s", err)
	}

	// Close the reader when we are done
	reader.Close()
}
```

kke:
Still relevant

brahads:
I see that there were a couple of suggestions for this issue. Do we have a timeline when this issue might be fixed in this package?

30787:
Can you confirm which version has the vulnerability remediated? v1.9.2 shows this vulnerability in twistlock scan results for sles12sp5 basesd container image.

kke:
Fixed in v1.9.3 via #1384

svetasmirnova:
[...]

PR #1376, opened 2023-03-10
This commit fixes a potential denial of service vulnerability in logrus.Writer() that could be triggered by logging text longer than 64kb without newlines.

ozfive:
Fix denial of service vulnerability in logrus.Writer()

The related issue [#1370](https://github.com/sirupsen/logrus/issues/1370) explains the problem.

This pull request fixes a potential denial of service vulnerability in logrus.Writer() that could be triggered by logging text longer than 64kb without newlines. Previously, the bufio.Scanner used by Writer() would hang indefinitely when reading such text without newlines, causing the application to become unresponsive.

To fix the issue, I've modified the writerScanner() function to split input into chunks of up to 64kb using a custom split function. I've also set the scanner buffer size to the maximum token size using bufio.MaxScanTokeSize.

With these changes, logrus.Writer() now properly handles long, newline-free log messages without causing a denial of service.

Note that the code has also been updated with more comments and that this fix is backwards-compatible. It does not affect existing applications using logrus.Writer()

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"io"
 	"runtime"
+	"strings"
 )
 
 // Writer at INFO level. See WriterLevel for details.
@@ -20,15 +21,18 @@ func (logger *Logger) WriterLevel(level Level) *io.PipeWriter {
 	return NewEntry(logger).WriterLevel(level)
 }
 
+// Writer returns an io.Writer that writes to the logger at the info log level
 func (entry *Entry) Writer() *io.PipeWriter {
 	return entry.WriterLevel(InfoLevel)
 }
 
+// WriterLevel returns an io.Writer that writes to the logger at the given log level
 func (entry *Entry) WriterLevel(level Level) *io.PipeWriter {
 	reader, writer := io.Pipe()
 
 	var printFunc func(args ...interface{})
 
+	// Determine which log function to use based on the specified log level
 	switch level {
 	case TraceLevel:
 		printFunc = entry.Trace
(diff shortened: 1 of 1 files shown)


ozfive:
@sirupsen Its not clear to me what is failing in the tests as I don't have any experience with appveyor. Can you take a look and see what the issue is?

dolmen:
https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/

ashmckenzie:
@ozfive I believe the following patch is required for this to pass CI:

```
diff --git a/writer.go b/writer.go
index 36032d0..7e7703c 100644
--- a/writer.go
+++ b/writer.go
@@ -75,7 +75,8 @@ func (entry *Entry) writerScanner(reader *io.PipeReader, printFunc func(args ...
                if len(data) > chunkSize {
                        return chunkSize, data[:chunkSize], nil
                }
-               return 0, nil, nil
+
+               return len(data), data, nil
        }

        //Use the custom split function to split the input
```

Also, I think what @dolmen is trying to say is that your commit message could be a little more concise on the first line and then move the more detailed content to the 'body' section.

I've created https://github.com/ozfive/logrus/pull/1 to demonstrate the fix as well as a suggested commit message. WDYT?

ozfive:
You are a god among men. I am still learning git so will make it my job to look up patching and try to understand what you did here.

ashmckenzie:
@sirupsen I believe this is ready for review, when you have time please 🙂 The git commits might need a squash though.

scarroll32:
@sirupsen is it possible to trigger the CI on this? Would love to see it merged.

voidspooks:
@sirupsen Hi, would it be possible to get this change merged?

ashmckenzie:
Thanks so much for merging @sirupsen 🙇‍♂️ May we please have a new release to include this fix? 🙏

sirupsen:
ah yes tagged v1.8.3

ashmckenzie:
> ah yes tagged v1.8.3

Thanks for creating a new tag @sirupsen 🙇‍♂️ The latest tag/release prior to the new v1.8.3 tag was v1.9.0 (which is what we use in our golang projects here at GitLab). Will there also be a v1.9.1 tag and release?

sirupsen:
🤦🏻 v1.9.1 released, I didn't pull down the most recent tags before deciding the new one

Luap99:
> Note that the code has also been updated with more comments and that this fix is backwards-compatible

This is no longer splitting at newlines at all so it is chaining the output for users. With this change there is no reason to use the Scanner at all, you could just Read from the reader and write that with printFunc().

Anyway the reason I am here is because the change is broken and causes panics: https://github.com/sirupsen/logrus/issues/1383

thaJeztah:
@kolyshkin any idea where these @openshift-merge-robot comments come from? Something misconfigured somewhere?

PR #1384, opened 2023-05-17
fix panic in Writer

### LINKED ISSUE #1383: v1.9.1 causes panic in writer
Found while updating to the lasted version:
```
panic: bufio.Scan: too many empty tokens without progressing

goroutine 27 [running]:
bufio.(*Scanner).Scan(0xc000653f28)
        /usr/lib/golang/src/bufio/scan.go:167 +0x82c
github.com/sirupsen/logrus.(*Entry).writerScanner(0x0?, 0xc000014058, 0xc000152140)
        /home/pholzing/go/src/github.com/containers/podman/vendor/github.com/sirupsen/logrus/writer.go:86 +0x11f
created by github.com/sirupsen/logrus.(*Entry).WriterLevel
        /home/pholzing/go/src/github.com/containers/podman/vendor/github.com/sirupsen/logrus/writer.go:57 +0x3d1
```

This seems to be caused by https://github.com/sirupsen/logrus/pull/1376 which does not look correct. the split function never tells the scanner to stop. Second it even changes the behaviour because it now no longer splits at newlines.

It easy to reproduce, just apply this test to the repo here:
```diff
diff --git a/writer_test.go b/writer_test.go
index 5c34927..a6f0b3c 100644
--- a/writer_test.go
+++ b/writer_test.go
@@ -1,10 +1,15 @@
 package logrus_test
 
 import (
+       "bytes"
        "log"
        "net/http"
+       "strings"
+       "testing"
+       "time"
 
        "github.com/sirupsen/logrus"
+       "github.com/stretchr/testify/assert"
 )
 
 func ExampleLogger_Writer_httpServer() {
@@ -32,3 +37,30 @@ func ExampleLogger_Writer_stdlib() {
        // Not logrus imported under the name `log`.
        log.SetOutput(logger.Writer())
 }
+
+func TestWriterSplitNewlines(t *testing.T) {
+       buf := bytes.NewBuffer(nil)
+       logger := logrus.New()
+       logger.Formatter = &logrus.TextFormatter{
+               DisableColors:    true,
+               DisableTimestamp: true,
+       }
+       logger.SetOutput(buf)
+       writer := logger.Writer()
+
+       const logNum = 10
+
+       for i := 0; i < logNum; i++ {
+               _, err := writer.Write([]byte("bar\nfoo\n"))
+               if err != nil {
+                       assert.NoError(t, err, "write.Write failed")
+               }
+       }
+       writer.Close()
+       // Test is flaky because it writes in another goroutine,
+       // we need to make sure to wait a bit so all write are done.
+       time.Sleep(500 * time.Millisecond)
+
+       lines := strings.Split(strings.TrimRight(buf.String(), "\n"), "\n")
+       assert.Len(t, lines, logNum*2, "logger printed incorrect number of lines")
+}
```

Luap99:
Commit 766cfece introduced this bug by defining an incorrect split function. First it breaks the old behavior because it never splits at newlines now. Second, it causes a panic because it never tells the scanner to stop. See the bufio.ScanLines function, something like:
```
if atEOF && len(data) == 0 {
	return 0, nil, nil
}
```
is needed to do that.

This commit fixes it by restoring the old behavior and calling bufio.ScanLines but also keep the 64KB check in place to avoid buffering for to long.

Two tests are added to ensure it is working as expected.

Fixes #1383

### DIFF
--- a/writer_test.go
+++ b/writer_test.go
@@ -1,10 +1,16 @@
 package logrus_test
 
 import (
+	"bufio"
+	"bytes"
 	"log"
 	"net/http"
+	"strings"
+	"testing"
+	"time"
 
 	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
 )
 
 func ExampleLogger_Writer_httpServer() {
@@ -31,4 +37,62 @@ func ExampleLogger_Writer_stdlib() {
 	// Note that `log` here references stdlib's log
 	// Not logrus imported under the name `log`.
 	log.SetOutput(logger.Writer())
+}
+
+func TestWriterSplitNewlines(t *testing.T) {
+	buf := bytes.NewBuffer(nil)
+	logger := logrus.New()
+	logger.Formatter = &logrus.TextFormatter{
+		DisableColors:    true,
+		DisableTimestamp: true,
+	}
+	logger.SetOutput(buf)
+	writer := logger.Writer()
+
+	const logNum = 10
+
+	for i := 0; i < logNum; i++ {
+		_, err := writer.Write([]byte("bar\nfoo\n"))
+		assert.NoError(t, err, "writer.Write failed")
+	}
+	writer.Close()
+	// Test is flaky because it writes in another goroutine,
+	// we need to make sure to wait a bit so all write are done.
+	time.Sleep(500 * time.Millisecond)
+
+	lines := strings.Split(strings.TrimRight(buf.String(), "\n"), "\n")
+	assert.Len(t, lines, logNum*2, "logger printed incorrect number of lines")
+}
+
+func TestWriterSplitsMax64KB(t *testing.T) {
+	buf := bytes.NewBuffer(nil)
+	logger := logrus.New()
+	logger.Formatter = &logrus.TextFormatter{
+		DisableColors:    true,
+		DisableTimestamp: true,
+	}
+	logger.SetOutput(buf)
+	writer := logger.Writer()
+
+	// write more than 64KB
+	const bigWriteLen = bufio.MaxScanTokenSize + 100
+	output := make([]byte, bigWriteLen)
+	// lets not write zero bytes
+	for i := 0; i < bigWriteLen; i++ {
+		output[i] = 'A'
+	}
+
+	for i := 0; i < 3; i++ {
+		len, err := writer.Write(output)
+		assert.NoError(t, err, "writer.Write failed")
+		assert.Equal(t, bigWriteLen, len, "bytes written")
+	}
+	writer.Close()
+	// Test is flaky because it writes in another goroutine,
+	// we need to make sure to wait a bit so all write are done.
+	time.Sleep(500 * time.Millisecond)
+
+	lines := strings.Split(strings.TrimRight(buf.String(), "\n"), "\n")
+	// we should have 4 lines because we wrote more than 64 KB each time
+	assert.Len(t, lines, 4, "logger printed incorrect number of lines")
 }
--- a/writer.go
+++ b/writer.go
@@ -70,16 +70,16 @@ func (entry *Entry) writerScanner(reader *io.PipeReader, printFunc func(args ...
 	scanner.Buffer(make([]byte, bufio.MaxScanTokenSize), bufio.MaxScanTokenSize)
 
 	// Define a split function to split the input into chunks of up to 64KB
-	chunkSize := 64 * 1024 // 64KB
+	chunkSize := bufio.MaxScanTokenSize // 64KB
 	splitFunc := func(data []byte, atEOF bool) (int, []byte, error) {
-		if len(data) > chunkSize {
+		if len(data) >= chunkSize {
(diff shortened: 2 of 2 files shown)


writer_test.go
```diff
@@ -32,3 +38,61 @@ func ExampleLogger_Writer_stdlib() {
...
+		assert.Equal(t, bigWriteLen, len, "bytes written")
+	}
+	writer.Close()
+	// Test is flaky because it writes in another goroutine,
+	// we need to make sure to wait a bit so all write are done.
+	time.Sleep(500 * time.Millisecond)
+
+	lines := strings.Split(strings.TrimRight(buf.String(), "\n"), "\n")
+	// we should have 4 lines because we wrote more than 64 KB each time
+	assert.Len(t, lines, 4, "logger printed incorrect number of lines")
```
sirupsen, line 97: Hey! Thank you for fixing this (I had to revert the other commit in master, because of this panic). Let's get this fixed properly.

I think there's two ways of dealing with this, and I'm interested in your take (and those on the original issue @kke @brahads @ozfive @ashmckenzie @scarroll32 @voidspooks).

Intuitively, I would expect this to yield 6 log lines:

```
# approach 1
line1: 1024 * 64 * A
line2: 100 * A
line3: 1024 * 64 * A
line4: 100 * A
line5: 1024 * 64 * A
line6: 100 * A
```

But, this implementation from @ozfive yields:

```
# approach 2 (implemented)
line1: 1024 * 64 * A
line2: 1024 * 64 * A 
line3: 1024 * 64 * A
line4: 300 * A
```

In other words, log lines bleed into each other.

I prefer approach 1, this seems way more intuitive to me when reading logs.

Luap99: I personally do not care much, the current behavior is to only split at newlines. So if you write something without a newline at the end it also buffered it like we do now.

I am fine with changing the behavior to approach 1 but at that point we might as well just do not use the Scanner at all and exec the print function for each Write() call the caller makes, in this case we do not need to worry about the scanner limitations at all and do not need extra buffers.

KarthikNayak: @sirupsen bumping this. I agree with @Luap99's explanation though.

hatricker: It would be great if we can put a fix for the panic soon so that we can upgrade to a version without the DoS vulnerability that's described in https://github.com/sirupsen/logrus/issues/1370

ashmckenzie: Sorry for the delay @sirupsen (was on vacation 😉), using the following test program:

```golang
package main

import (
	"bufio"
	"strings"
	"time"

	logrus "github.com/sirupsen/logrus"
)

func main() {
	output1 := strings.Repeat("A", bufio.MaxScanTokenSize) + "Z"
	output2 := strings.Repeat("B", bufio.MaxScanTokenSize) + "Z"

	logger := logrus.New()
	writer := logger.Writer()

	writer.Write([]byte("start"))
	writer.Write([]byte(output1))
	writer.Write([]byte(output2))
	writer.Write([]byte("end"))

	writer.Close()

	time.Sleep(500 * time.Millisecond)
}
```

as well as an updated `writerScanner()` function:

```golang
func (entry *Entry) writerScanner(reader *io.PipeReader, printFunc func(args ...interface{})) {
	for {
		data := make([]byte, bufio.MaxScanTokenSize)
		c, err := reader.Read(data)
		if err != nil && err != io.EOF {
			entry.Errorf("Error while reading from Writer: %s", err)
			break
		}

		if c > 0 {
			printFunc(string(bytes.Trim(data, "\x00")))
		}
	}

	// Close the reader when we are done
	reader.Close()
}
```

I believe this achieves the preferred `# approach 1`:

```sh
$ go run play.go 2>&1
time="2023-06-02T22:06:34+10:00" level=info msg=start
time="2023-06-02T22:06:34+10:00" level=info msg=A<lots of As>
time="2023-06-02T22:06:34+10:00" level=info msg=Z
time="2023-06-02T22:06:34+10:00" level=info msg=B<lots of Bs>
time="2023-06-02T22:06:34+10:00" level=info msg=Z
time="2023-06-02T22:06:34+10:00" level=info msg=end
```

Am just updating the `TestWriterSplitsMax64KB()` test in a new branch to demonstrate. WDYT?

sirupsen:
@ashmckenzie @Luap99 I've decided to merge this behaviour and it has been released as v1.9.3. Thank you both for your efforts!

It is the least invasive change, and I'm committed to maintaining as strong backwards compatibility as possible due to the # of users of logrus. Let me know if you find any behaviours with this version

ashmckenzie:
Thanks @sirupsen and @Luap99, am very pleased to see this merged 🙂:+1::bow:

kke:
Great! But as noted here https://github.com/k0sproject/k0s/pull/3182#issuecomment-1578691745 - the downside is it's not unicode safe

kke:
Yeah, for us the problem comes from an external program which is launched through os/exec and is given logrus a writer as `cmd.Stdout`, one of the external tools dumped a certificate or whatever in debug mode and made the writer crash.

PR #1387, opened 2023-05-28
support using with tinygo

rajatjindal:
hello

I am using `logrus` in a `golang wasm project` ([code here](https://github.com/rajatjindal/goodfirstissue/tree/wasm6)) and when trying to build it using `tinygo`, I am running into following issue:

```
$) tinygo build -target=wasi -gc=leaking -no-debug --tags 'tinygo' -o main.wasm main.go
wasm-ld: error: /var/folders/5r/6y6fr12x1dq6qnt74zn4_1wm0000gn/T/tinygo159344941/main.o: undefined symbol: golang.org/x/sys/unix.Syscall
error: failed to link /var/folders/5r/6y6fr12x1dq6qnt74zn4_1wm0000gn/T/tinygo159344941/main: exit status 1
Error: Build command for component goodfirstissue failed with status Exited(1)
```

I ran into a similar problem using [mattn/go-isatty](https://github.com/mattn/go-isatty), which is apparently fixed by a [similar PR](https://github.com/mattn/go-isatty/pull/74) 

therefore I am submitting this PR for `logurs` to allow us to use it with `tinygo` in a similar fashion.

Kindly let me know if this makes sense. I am happy to answer or implement any feedback that you might have.

Regards
Rajat Jindal

### DIFF
--- a/terminal_check_bsd.go
+++ b/terminal_check_bsd.go
@@ -1,5 +1,6 @@
+//go:build (darwin || dragonfly || freebsd || netbsd || openbsd) && (!js || !tinygo)
 // +build darwin dragonfly freebsd netbsd openbsd
-// +build !js
+// +build !js !tinygo
 
 package logrus
 
--- a/terminal_check_js.go
+++ b/terminal_check_js.go
@@ -1,4 +1,5 @@
-// +build js
+//go:build js || tinygo
+// +build js tinygo
 
 package logrus
 
--- a/terminal_check_no_terminal.go
+++ b/terminal_check_no_terminal.go
@@ -1,4 +1,5 @@
-// +build js nacl plan9
+//go:build js || nacl || plan9 || tinygo
+// +build js nacl plan9 tinygo
 
 package logrus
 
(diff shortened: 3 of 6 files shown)


rajatjindal:
requesting review from @sirupsen

sirupsen:
Can you explain https://github.com/sirupsen/logrus/pull/1388/files vs this PR? @flavio

flavio:
They are similar. TinyGo and the official Go compiler use a different identifier for the WASI target. TinyGo uses `wasi`, while the official Go compiler will use `wasip1`.

I think we could make the two PR conflate into a single one and make it target both GOOS values.

BTW, the usage of the `tinygo` identifier seems a bit too wide to me. This would identify also a TinyGo binary built for a regular linux/osx/windows system.

rajatjindal:
I am happy to have my PR closed in favor of #1388 if we want to add support of `tinygo` there as well.  thank you.

flavio:
I've updated my PR to include tinygo + wasi support

sirupsen:
merged the other one

PR #1389, opened 2023-06-02
Improve how data > 64KB is logged

ashmckenzie:
This PR extends https://github.com/sirupsen/logrus/pull/1384 to support the desired approach of create new log lines when the log data is greater than 64KB.

The following is a decent test program that demonstrates this working:

```golang
package main

import (
	"bufio"
	"strings"
	"time"

	logrus "github.com/sirupsen/logrus"
)

func main() {
	output1 := strings.Repeat("A", bufio.MaxScanTokenSize) + "Z"
	output2 := strings.Repeat("B", bufio.MaxScanTokenSize) + "Z"

	logger := logrus.New()
	writer := logger.Writer()

	writer.Write([]byte("start"))
	writer.Write([]byte(output1))
	writer.Write([]byte(output2))
	writer.Write([]byte("end"))

	writer.Close()

	time.Sleep(500 * time.Millisecond)
}
```

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -55,37 +55,23 @@ func (entry *Entry) WriterLevel(level Level) *io.PipeWriter {
 	return writer
 }
 
-// writerScanner scans the input from the reader and writes it to the logger
 func (entry *Entry) writerScanner(reader *io.PipeReader, printFunc func(args ...interface{})) {
-	scanner := bufio.NewScanner(reader)
-
-	// Set the buffer size to the maximum token size to avoid buffer overflows
-	scanner.Buffer(make([]byte, bufio.MaxScanTokenSize), bufio.MaxScanTokenSize)
-
-	// Define a split function to split the input into chunks of up to 64KB
-	chunkSize := bufio.MaxScanTokenSize // 64KB
-	splitFunc := func(data []byte, atEOF bool) (int, []byte, error) {
-		if len(data) >= chunkSize {
-			return chunkSize, data[:chunkSize], nil
+	for {
+		data := make([]byte, bufio.MaxScanTokenSize)
+		c, err := reader.Read(data)
+		if err != nil && err != io.EOF {
+			entry.Errorf("Error while reading from Writer: %s", err)
+			break
 		}
 
-		return bufio.ScanLines(data, atEOF)
+		if c > 0 {
+			chunks := strings.Split(strings.TrimRight(string(data[:c]), "\n"), "\n")
+			for _, section := range chunks {
+				printFunc(section)
+			}
+		}
 	}
 
-	// Use the custom split function to split the input
-	scanner.Split(splitFunc)
-
-	// Scan the input and write it to the logger using the specified print function
-	for scanner.Scan() {
-		printFunc(strings.TrimRight(scanner.Text(), "\r\n"))
-	}
-
-	// If there was an error while scanning the input, log an error
-	if err := scanner.Err(); err != nil {
-		entry.Errorf("Error while reading from Writer: %s", err)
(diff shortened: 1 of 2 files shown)


ashmckenzie:
WDYT @Luap99, @sirupsen ?

writer.go
```diff
@@ -54,14 +56,25 @@ func (entry *Entry) WriterLevel(level Level) *io.PipeWriter {
...
+	for {
+		data := make([]byte, bufio.MaxScanTokenSize)
+		c, err := reader.Read(data)
+		if err != nil && err != io.EOF {
+			entry.Errorf("Error while reading from Writer: %s", err)
+			break
+		}
+
+		if c > 0 {
+			chunks := strings.Split(strings.TrimRight(string(bytes.Trim(data, "\x00")), "\n"), "\n")
```
Luap99, line 70: why is the zero byte trimmed, this feels arbitrary?

writer_test.go
```diff
@@ -32,3 +38,61 @@ func ExampleLogger_Writer_stdlib() {
...
+		assert.NoError(t, err, "writer.Write failed")
+		assert.Equal(t, bigWriteLen, len, "bytes written")
+	}
+	writer.Close()
+	// Test is flaky because it writes in another goroutine,
+	// we need to make sure to wait a bit so all write are done.
+	time.Sleep(500 * time.Millisecond)
+
+	lines := strings.Split(strings.TrimRight(buf.String(), "\n"), "\n")
+	// we should have 4 lines because we wrote more than 64 KB each time
```
Luap99, line 96: should change the comment 6

Luap99 (commented):
I don't know how serious backwards compatibility is taken but we never know if users of logrus depend on the current newline split behaviour which is buffer the write until a newline is found.
Note that I personally do not care about the exact behaviour so I can live with either one but this project is used by so many we can never be sure if there are people who depend on the current behaviour.

I personally would prefer an implementation were we would just get rid of the extra goroutine and pipe implementation and just return a writer that does exactly what you are doing here. This would also removes the race condition that you can see in the test as everything would be single threaded. However because `io.PipeWriter` is a struct we cannot change the return type it without breaking the logrus API which is even worse.

In any case these are just my thoughts, I wouldn't be opposed to merge this.

writer_test.go
ashmckenzie: Fixed.

writer.go
ashmckenzie: During testing, for data that was < 64KB, the remaining bytes were `\x00` which causes issues, so `bytes.Trim(data, "\x00")` was there to clean this up. I've since improved the logic to only process the number of read bytes returned by `reader.Read()` which improves things, WDYT?

ashmckenzie:
> I don't know how serious backwards compatibility is taken but we never know if users of logrus depend on the current newline split behaviour which is buffer the write until a newline is found.
> Note that I personally do not care about the exact behaviour so I can live with either one but this project is used by so many we can never be sure if there are people who depend on the current behaviour.

This is a good point and one I think @sirupsen and the other maintainers need to decide.

> I personally would prefer an implementation were we would just get rid of the extra goroutine and pipe implementation and just return a writer that does exactly what you are doing here. This would also removes the race condition that you can see in the test as everything would be single threaded. However because io.PipeWriter is a struct we cannot change the return type it without breaking the logrus API which is even worse.

This sounds like an excellent follow-up refactor issue. Want to pair up on it? 🙂

> In any case these are just my thoughts, I wouldn't be opposed to merge this.

Thank-you, we really appreciate your time and input 🙇

PR #1390, opened 2023-06-03
Fix staticcheck issues

ashmckenzie:
This PR fixes the following go-staticcheck errors:

```sh
$ staticcheck
entry_test.go:39:49: should not use built-in type string as key for value; define your own type to avoid collisions (SA1029)
entry_test.go:59:50: should not use built-in type string as key for value; define your own type to avoid collisions (SA1029)
entry_test.go:63:50: should not use built-in type string as key for value; define your own type to avoid collisions (SA1029)
logrus_test.go:547:2: should merge variable declaration with assignment on next line (S1021)
```

### DIFF
--- a/entry_test.go
+++ b/entry_test.go
@@ -10,6 +10,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+type contextKeyType string
+
 func TestEntryWithError(t *testing.T) {
 
 	assert := assert.New(t)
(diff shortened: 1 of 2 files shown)


ashmckenzie:
@sirupsen here's a very small PR to fix some go-staticcheck issues in a few tests 🙂

sirupsen:
Cheers

Issue #1401, opened 2023-09-01
Feature: Add slog compatibility layer


flimzy:
I know this project is in maintenence mode, but I think it would be worth making an exception to the "no new features" rule to add a simple `slog` compatibility layer, as the alternative is some minor chaos when others create competing implementations, and inevitably abandon them.

This could take one, or even two, forms:

- A `slog` hook for Logrus
- A Logrus handler for `slog`

I would be happy to take this on and write the code (I have a partial implementation already completed for a client I'm working with).

ripienaar:
Would love to see this accepted in core logrus also.

flimzy:
I've begun work on this, and intend to create a PR before too long.

PR #1407, opened 2023-11-09
Add slog hook

flimzy:
This addresses half of #1401. The easy half, honestly.

I have an implementation of the other half we've been using in our own project, but it will need some polish before making it public, so I'm submitting this PR now to gauge interest of the project maintainers.  If we like this, I'll work on polishing the other half in a later PR.

### DIFF
--- a/hooks/slog/slog.go
+++ b/hooks/slog/slog.go
@@ -9,9 +9,16 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// LevelMapper maps a [github.com/sirupsen/logrus.Level] value to a
+// [slog.Leveler] value. To change the default level mapping, for instance
+// to allow mapping to custom or dynamic slog levels in your application, set
+// [SlogHook.LevelMapper] to your own implementation of this function.
+type LevelMapper func(logrus.Level) slog.Leveler
+
 // SlogHook sends logs to slog.
 type SlogHook struct {
-	logger *slog.Logger
+	logger      *slog.Logger
+	LevelMapper LevelMapper
 }
 
 var _ logrus.Hook = (*SlogHook)(nil)
@@ -31,7 +38,10 @@ func NewSlogHook(logger *slog.Logger) *SlogHook {
 	}
 }
 
-func (*SlogHook) toSlogLevel(level logrus.Level) slog.Level {
+func (h *SlogHook) toSlogLevel(level logrus.Level) slog.Leveler {
+	if h.LevelMapper != nil {
+		return h.LevelMapper(level)
+	}
 	switch level {
 	case logrus.PanicLevel, logrus.FatalLevel, logrus.ErrorLevel:
 		return slog.LevelError
@@ -62,7 +72,7 @@ func (h *SlogHook) Fire(entry *logrus.Entry) error {
 	}
 	h.logger.Log(
 		entry.Context,
-		h.toSlogLevel(entry.Level),
+		h.toSlogLevel(entry.Level).Level(),
 		entry.Message,
 		attrs...,
 	)
--- a/hooks/slog/slog.go
+++ b/hooks/slog/slog.go
@@ -1,0 +1,70 @@
+//go:build go1.21
+// +build go1.21
+
+package slog
+
+import (
+	"log/slog"
+
+	"github.com/sirupsen/logrus"
+)
+
+// SlogHook sends logs to slog.
+type SlogHook struct {
+	logger *slog.Logger
+}
+
+var _ logrus.Hook = (*SlogHook)(nil)
+
+// NewSlogHook creates a hook that sends logs to an existing slog Logger.
+// This hook is intended to be used during transition from Logrus to slog,
+// or as a shim between different parts of your application or different
+// libraries that depend on different loggers.
+//
+// Example usage:
+//
+//	logger := slog.New(slog.NewJSONHandler(os.Stderr, nil))
+//	hook := NewSlogHook(logger)
+func NewSlogHook(logger *slog.Logger) *SlogHook {
+	return &SlogHook{
+		logger: logger,
+	}
+}
+
+func (*SlogHook) toSlogLevel(level logrus.Level) slog.Level {
+	switch level {
+	case logrus.PanicLevel, logrus.FatalLevel, logrus.ErrorLevel:
+		return slog.LevelError
+	case logrus.WarnLevel:
+		return slog.LevelWarn
+	case logrus.InfoLevel:
+		return slog.LevelInfo
+	case logrus.DebugLevel, logrus.TraceLevel:
+		return slog.LevelDebug
+	default:
+		// Treat all unknown levels as errors
+		return slog.LevelError
+	}
+}
+
+// Levels always returns all levels, since slog allows controlling level
+// enabling based on context.
+func (h *SlogHook) Levels() []logrus.Level {
+	return logrus.AllLevels
+}
+
+// Fire sends entry to the underlying slog logger. The Time and Caller fields
+// of entry are ignored.
+func (h *SlogHook) Fire(entry *logrus.Entry) error {
+	attrs := make([]interface{}, 0, len(entry.Data))
+	for k, v := range entry.Data {
+		attrs = append(attrs, slog.Any(k, v))
+	}
+	h.logger.Log(
+		entry.Context,
+		h.toSlogLevel(entry.Level),
+		entry.Message,
+		attrs...,
+	)
+	return nil
+}
--- a/hooks/slog/slog_test.go
+++ b/hooks/slog/slog_test.go
@@ -1,0 +1,72 @@
+//go:build go1.21
+// +build go1.21
+
+package slog
+
+import (
+	"bytes"
+	"io"
+	"log/slog"
+	"strings"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+)
+
+func TestSlogHook(t *testing.T) {
+	tests := []struct {
+		name string
+		fn   func(*logrus.Logger)
+		want []string
+	}{
+		{
+			name: "defaults",
+			fn: func(log *logrus.Logger) {
+				log.Info("info")
+			},
+			want: []string{
+				"level=INFO msg=info",
+			},
+		},
+		{
+			name: "with fields",
+			fn: func(log *logrus.Logger) {
+				log.WithFields(logrus.Fields{
+					"chicken": "cluck",
+				}).Error("error")
+			},
+			want: []string{
+				"level=ERROR msg=error chicken=cluck",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			buf := &bytes.Buffer{}
+			slogLogger := slog.New(slog.NewTextHandler(buf, &slog.HandlerOptions{
+				// Remove timestamps from logs, for easier comparison
+				ReplaceAttr: func(_ []string, a slog.Attr) slog.Attr {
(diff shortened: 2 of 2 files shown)


flimzy:
I see that CI is failing, and probably needs to be updated.

For the tests in this PR to be meaningful, we'll alsoneed to run tests against Go 1.21 (since the new code is behind a build tag).

If you'd like me to create a PR to address any of these, please let me know.  I'm quite happy to contribute, but don't want to waste my effort if the package maintainers aren't interested :wink:

hooks/slog/slog.go
```diff
@@ -0,0 +1,80 @@
...
+// enabling based on context.
+func (h *SlogHook) Levels() []logrus.Level {
+	return logrus.AllLevels
+}
+
+// Fire sends entry to the underlying slog logger. The Time and Caller fields
+// of entry are ignored.
+func (h *SlogHook) Fire(entry *logrus.Entry) error {
+	attrs := make([]interface{}, 0, len(entry.Data))
+	for k, v := range entry.Data {
```
mdelapenya: question: would be of interest to sort the fields to have a deterministic list of fields? I understand that it could come with some performance cost, although we could decide to sort them only when the count is low, i.e., 10.

hooks/slog/slog.go
```diff
@@ -0,0 +1,80 @@
...
+		attrs = append(attrs, slog.Any(k, v))
+	}
+	h.logger.Log(
+		entry.Context,
+		h.toSlogLevel(entry.Level).Level(),
+		entry.Message,
+		attrs...,
+	)
```
mdelapenya: suggestion: Logger.Log() does not expose handler errors to the caller, and there are some situations where the handler can error (file permissions, disk full...), so I think this is silently discarding errors, even if the logging failed.

If you use a slog.Handler in the SlogHook instead of a logger, and build a record here, you can handle those errors here. Please noticed I added an optional errorHandler, that could be configured in the constructor or via functional options.... or simply completely skip it and always print 🤷 

```go
func (h *SlogHook) Fire(entry *logrus.Entry) error {
	attrs := make([]interface{}, 0, len(entry.Data))
	for k, v := range entry.Data {
		attrs = append(attrs, slog.Any(k, v))
	}

	record := slog.NewRecord(entry.Time, h.toSlogLevel(entry.Level).Level(),
		entry.Message, 0)
	for _, attr := range attrs {
		record.AddAttrs(attr.(slog.Attr))
	}

	err := h.handler.Handle(entry.Context, record)
	if err != nil {
		if h.errorHandler != nil {
			h.errorHandler(err, entry)
		} else {
			// print to stderr if no custom error handler is set
			fmt.Fprintf(os.Stderr, "slog handler error: %v\n", err)
		}
		return err
	}
	return nil
}
```

If this is of your interest, we should also extend the tests, and include a README.

Thoughts?

hooks/slog/slog.go
```diff
@@ -0,0 +1,80 @@
...
+// LevelMapper maps a [github.com/sirupsen/logrus.Level] value to a
+// [slog.Leveler] value. To change the default level mapping, for instance
+// to allow mapping to custom or dynamic slog levels in your application, set
+// [SlogHook.LevelMapper] to your own implementation of this function.
+type LevelMapper func(logrus.Level) slog.Leveler
+
+// SlogHook sends logs to slog.
+type SlogHook struct {
+	logger      *slog.Logger
+	LevelMapper LevelMapper
```
mdelapenya, line 20: question: is it possible to change the levelMapper? Else, I'd keep it private
```suggestion
	levelMapper LevelMapper
```

flimzy: Yes, the idea is that it can be specified by the user. See the GoDoc blurb on the `LevelMapper` type:

```
// LevelMapper maps a [github.com/sirupsen/logrus.Level] value to a
// [slog.Leveler] value. To change the default level mapping, for instance
// to allow mapping to custom or dynamic slog levels in your application, set
// [SlogHook.LevelMapper] to your own implementation of this function.
type LevelMapper func(logrus.Level) slog.Leveler
```

hooks/slog/slog.go
flimzy: It's not a difficult change, but of limited utility, IMO.  Any error from Fire() is just written to STDERR.

I think a better approach would be to implement the other half of #1401 and make Logrus provide an slog handler.  Then the error would propagate the other direction, and slog would be responsible for handling any error from Fire(), rather than the other way around.

That said, I have implemented your suggestion--logging an error to STDERR is still better than ignoring it.

hooks/slog/slog.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Levels always returns all levels, since slog allows controlling level
+// enabling based on context.
+func (h *SlogHook) Levels() []logrus.Level {
+	return logrus.AllLevels
+}
+
+// Fire sends entry to the underlying slog logger. The Time and Caller fields
+// of entry are ignored.
+func (h *SlogHook) Fire(entry *logrus.Entry) error {
```
thaJeztah, line 74: Not 100% sure, but should we use the context that's set on the entry? https://github.com/sirupsen/logrus/blob/b61f268f75b6ff134a62cd62aee1095fa12e8d2e/entry.go#L68-L69

Something like;

```go
	ctx := entry.Context
	if ctx == nil {
		ctx = context.Background()
	}
```

thaJeztah: We could probably add an early return in case it shouldn't fire;

```go
	handler := h.logger.Handler()

	lvl := h.toSlogLevel(entry.Level).Level()
	if !handler.Enabled(ctx, lvl) {
		return nil
	}
```

flimzy:
I've added a test to validate that we report the proper caller, and updated the skip argument accordingly.

hooks/slog/slog.go
```diff
@@ -0,0 +1,82 @@
...
+	attrs := make([]any, 0, len(entry.Data))
+	for k, v := range entry.Data {
+		attrs = append(attrs, slog.Any(k, v))
+	}
+	var pcs [1]uintptr
+	// skip 8 callers to get to the original logrus caller
+	runtime.Callers(8, pcs[:])
+	r := slog.NewRecord(time.Now(), h.toSlogLevel(entry.Level).Level(), entry.Message, pcs[0])
```
thaJeztah: Should this use `entry.Time` (instead of `time.Now()`?

hooks/slog/slog.go
```diff
@@ -0,0 +1,82 @@
...
+// Fire sends entry to the underlying slog logger. The Time and Caller fields
+// of entry are ignored.
+func (h *SlogHook) Fire(entry *logrus.Entry) error {
+	attrs := make([]any, 0, len(entry.Data))
+	for k, v := range entry.Data {
+		attrs = append(attrs, slog.Any(k, v))
+	}
+	var pcs [1]uintptr
+	// skip 8 callers to get to the original logrus caller
+	runtime.Callers(8, pcs[:])
```
thaJeztah: Looking if there's some optimisations to be made here; there's also `entry.Caller` - wondering if that should be used?

thaJeztah: Heh - I was lazy; this is what ChatGPT suggested for this; not entirely sure if the `runtime.Callers` suggestion is good;


```go
	var pc uintptr
	if entry.Caller != nil {
		pc = entry.Caller.PC
	} else {
		var pcs [1]uintptr
		// Skip runtime.Callers + this function; adjust as needed or make configurable.
		runtime.Callers(2, pcs[:])
		pc = pcs[0]
	}

	r := slog.NewRecord(entry.Time, lvl, entry.Message, pc)

	if len(entry.Data) != 0 {
		attrs := make([]slog.Attr, 0, len(entry.Data))
		for k, v := range entry.Data {
			attrs = append(attrs, slog.Any(k, v))
		}
		r.AddAttrs(attrs...)
	}

	return handler.Handle(ctx, r)
```

hooks/slog/slog.go
flimzy: yeah, we should probably do both of those.  I wasn't aware (or had forgotten) that there's a context in the Entry struct.

hooks/slog/slog.go
flimzy: Yes!

hooks/slog/slog.go
flimzy: We definitely can do that, but it subtly changes behavior.  Perhaps in a good way?

If we use entry.Caller.PC, then to get `source=` in the final output,  both Logrus and the slog handler need to be configured to show sources.  With the current implementation,  the behavior depends only on slog's configuration.

I'm not sure which is "more correct".  I suppose in practice, they'll likely both be enabled or both disabled.  So I'll make the change to use `entry.Caller.PC` unless you feel that's not appropriate :)

hooks/slog/slog.go
```diff
@@ -0,0 +1,84 @@
...
+// of entry are ignored.
+func (h *SlogHook) Fire(entry *logrus.Entry) error {
+	lvl := h.toSlogLevel(entry.Level).Level()
+	if !h.logger.Handler().Enabled(entry.Context, lvl) {
+		return nil
+	}
+	attrs := make([]any, 0, len(entry.Data))
+	for k, v := range entry.Data {
+		attrs = append(attrs, slog.Any(k, v))
+	}
```
oreissig: One thing I noticed: logrus used to sort its values when printing the log line.
However here the iteration order of the `entry.Data` hashmap is not deterministic. The resulting values are put into a slice, which slog keeps ordered as is. This means bridged key-value-pairs will always be shuffled, making the log lines hard to read as a human.

flimzy: I guess that could be done. Though I'd push back on it a bit, as it adds computation to something that should be as lightweight as possible. Though I know that's a difference of philosophy between logrus and log/slog.  I guess logrus likes to make logs "nice" at the expense of performance.  log/slog aims to make logs very efficient, even if the ergonomics are a bit less than ideal.  As an example, when logging JSON with log/slog, it writes the key/value pairs as they are received, rather  than buffering them, to reduce allocations and other overhead. One of the results of this is that it means you can actually produce logs with duplicate JSON keys.

Now which philosophy is more appropriate for a feature bridging two worlds?  Hard to say.  It's a coin toss I guess.  But my feeling is that if you want this sort of niceness-over-performance, you're probably not going to be using slog in the first place, so you wouldn't be using this hook.  :shrug:

oreissig: > Though I'd push back on it a bit, as it adds computation to something that should be as lightweight as possible. Though I know that's a difference of philosophy between logrus and log/slog. I guess logrus likes to make logs "nice" at the expense of performance. log/slog aims to make logs very efficient, even if the ergonomics are a bit less than ideal.

I once wrote a similar bridge myself and also did some benchmarks. Interestingly, logrus bridged to slog (with sorting) is still significantly faster than plain logrus.
(at least when I used logrus as a mere facade like slf4j in Java world, which means I have to call `SetOutput` and `SetFormatter` and provide no-op implementations to disable logrus' own processing)

Another thing I think is worth considering:
slog has consistent value ordering. It's just that the values are not lexicographically sorted, but by insertion order. Hashmaps, on the other hand, are totally unpredictable in their order. IMHO that's worse than insertion order (but feel free to disagree, just wanted to give some aspects to think about 🙂)

flimzy: To be clear: I don't have a strong feeling on this. I won't likely be using this feature any longer any way :wink: I had forgotten that this PR even existed until I was pinged to make some changes.  I've long since retired logrus from any active projects I'm on.

So whatever direction the maintainers want to go I'm happy with.

PR #1410, opened 2023-11-18
fix text_formatter.go's bug

amone-bit:
There is a BUG. when TextFormatter's ForceColors field is true and if log level letter length is five, it will only display the first four letter.EX. ERROR will display ERRO.

image


amone-bit:
I'm so sorry about that I'm unfamiliar to pr. So as not to cause you any more trouble, I'm turning off the pr

PR #1413, opened 2023-12-29
Support error array in field when using json formatter

same-id:
In addition to:
  https://github.com/sirupsen/logrus/issues/137

### DIFF
--- a/json_formatter.go
+++ b/json_formatter.go
@@ -59,6 +59,14 @@ type JSONFormatter struct {
 	PrettyPrint bool
 }
 
+func errorsArrayToStrings(errs []error) []string {
+	strs := make([]string, len(errs))
+	for i, err := range errs {
+		strs[i] = err.Error()
+	}
+	return strs
+}
+
 // Format renders a single log entry
 func (f *JSONFormatter) Format(entry *Entry) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


same-id:
Bump, I think this can count as a bug fix

json_formatter.go
```diff
@@ -68,6 +76,8 @@ func (f *JSONFormatter) Format(entry *Entry) ([]byte, error) {
 			// Otherwise errors are ignored by `encoding/json`
 			// https://github.com/sirupsen/logrus/issues/137
 			data[k] = v.Error()
+		case []error:
+			data[k] = errorsArrayToStrings(v)
```
thaJeztah, line 80: Slightly thinking out loud here; now that go stdlib provides native support for "multi-errors" through `errors.Join`, I wonder if this is something that would be better fixed in the code using logrus.


i.e.; instead of `WithField("errors", []error{...})`, to use `WithError(errors.Join(errs....))` (or similar)

thaJeztah: The tricky bit is that logrus will (probably) never be able to fully cover all scenarios; if such a `[]error` (or a regular `error`) would be part of a `struct` that's set as field; it would probably still produce a non-string output for the error(s).

PR #1429, opened 2024-03-12
README: fix godoc.org -> pkg.go.dev

dolmen:
Fix references to godoc.org to point instead to pkg.go.dev.


thaJeztah (approved):
LGTM, thanks!

PR #1433, opened 2024-05-15
Add ExcludeKey option to TextFormatter

ningyougang:
- [x] Add ExcludeKey option to TextFormatter

* AS-IS (`ExcludeKey : false`, this is default)

  ```
  time="2024-05-14T16:47:03+09:00" level=warning msg="message"
  ```
* TO-BE (after applied `ExcludeKey : true`)
  ```
  2024-05-14T16:47:03+09:00 warning message
  ```

I think this is a common case, when ·some log collector agent` collect the logs from file or standard output, has no need the startingKey(e.g. `time=`)


text_formatter.go
```diff
@@ -331,9 +337,14 @@ func (f *TextFormatter) appendValue(b *bytes.Buffer, value interface{}) {
 		stringVal = fmt.Sprint(value)
 	}
 
-	if !f.needsQuoting(stringVal) {
+	// write value without quoting directly when ExcludeKey is true
+	if f.ExcludeKey {
```
ningyougang, line 341: If configure ExcludeKey  as true, just write value without quoting directly, otherwise, the logs will be as below
```
"0001-01-01T00:00:00Z" warning "message"
```

text_formatter.go
```diff
@@ -331,9 +337,14 @@ func (f *TextFormatter) appendValue(b *bytes.Buffer, value interface{}) {
...
 		b.WriteString(stringVal)
 	} else {
-		b.WriteString(fmt.Sprintf("%q", stringVal))
+		if !f.needsQuoting(stringVal) {
+			b.WriteString(stringVal)
+		} else {
+			b.WriteString(fmt.Sprintf("%q", stringVal))
+		}
```
ningyougang, line 348: Keep the same logic as before when ExcludeKey is false(this is default)

ningyougang:
Can someone review this PR?

Ilovemesomehim1 (approved):

PR #1449, opened 2024-11-28
Exposed Lock functionality/controls

UnaffiliatedCode:
When implementers are managing multiple applications within a single service, multiple logrus instances are used. These services can have different subset configurations such as Minimum level assignment.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -68,6 +68,10 @@ func (mw *MutexWrap) Unlock() {
 	}
 }
 
+func (mw *MutexWrap) Enable() {
+	mw.disabled = false
+}
+
 func (mw *MutexWrap) Disable() {
 	mw.disabled = true
 }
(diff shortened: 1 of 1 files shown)


BoomchainLabs (approved):

logger.go
```diff
@@ -346,6 +350,18 @@ func (logger *Logger) Exit(code int) {
...
+func (logger *Logger) Lock() {
+	logger.mu.Lock()
+}
+func (logger *Logger) Unlock() {
+	logger.mu.Unlock()
+}
+
+// When an implementer is manipulating the logger concurrently, the
+// implementer should call `SetYesLock` to enable the locking mechanism
+func (logger *Logger) SetYesLock() {
```
AkihiroSuda, line 362: What is "YesLock"?

UnaffiliatedCode: If you read line 369, you will see a paired method:
```go
//When file is opened with appending mode, it's safe to
//write concurrently to a file (within 4k message on Linux).
//In these cases user can choose to disable the lock.
func (logger *Logger) SetNoLock() {
	logger.mu.Disable()
}
```

This is a continuation of the pattern used by the authors.

kolyshkin: @UnaffiliatedCode afaik this should be called `SetLock`. Also, `SetNoLock` documentation should be updated by telling something like 
```go
// You can re-enable the lock later by calling [Logger.SetLock].
```

PR #1459, opened 2025-06-19
CI: update actions

AkihiroSuda:


thaJeztah:
oh! looks like the golangci-lint config needs updating as well (probably gonna be a bit tedious);

```
  Running [/home/runner/golangci-lint-2.1.6-linux-amd64/golangci-lint config path] in [/home/runner/work/logrus/logrus] ...
  Running [/home/runner/golangci-lint-2.1.6-linux-amd64/golangci-lint run] in [/home/runner/work/logrus/logrus] ...
  Error: can't load config: unsupported version of the configuration: "" See https://golangci-lint.run/product/migration-guide for migration instructions
  Failed executing command with error: can't load config: unsupported version of the configuration: "" See https://golangci-lint.run/product/migration-guide for migration instructions
```

.github/workflows/ci.yaml
```diff
@@ -47,15 +48,15 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        go-version: [1.17.x]
+        go-version: [stable, oldstable]
```
thaJeztah: should we keep a `oldest` version in the matrix you think? (could be the `1.17.x` or perhaps slightly newer if we want the updated `go.mod` format (which I think was in go1.18, not 100% sure)

AkihiroSuda: done

thaJeztah:
> Also seeing data race

Saw that one, yes! Happy you fixed that one as well; thank you!

PR #1460, opened 2025-06-20
go.mod: bump up deps; CI: remove appveyor, add macOS

AkihiroSuda:
The minimum Go version is now bumped up from 1.13 to 1.17. (Corresponds to the Go version tested on GHA)


go.mod
```diff
@@ -1,9 +1,14 @@
...
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/stretchr/testify v1.7.0
-	golang.org/x/sys v0.0.0-20220715151400-c0bba94af5f8
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-go 1.13
+go 1.17
```
thaJeztah: nit: perhaps a good opportunity to move it to the top (at least I'm used to go version being at the start of go.mod) WDYT?

thaJeztah:
> But I guess we can just remove Appveyor, as we now have Windows on GHA

Yes, makes sense to me. Less to maintain!

go.mod
AkihiroSuda: done

PR #1472, opened 2025-09-30
refactor: replace the deprecated function in the ioutil package

goldlinker:
Starting from Go 1.16, some functions in the ioutil package have been deprecated. The Go team recommends quickly switching to methods from the os or io packages instead. 

More info can see: https://github.com/golang/go/issues/45557


thaJeztah (approved):
yup, LGTM, thanks!

PR #1481, opened 2026-02-09
chore: set debug level log color as green

hushicn:
`gray = 37` is actually white. Change the color to green to make it easier to read debug level logs.


thaJeztah:
closing per the above.

PR #717, opened 2018-02-23
add gopherjs build tag

gracenoah:
it behaves the same way as the appengine tag

fix #716

PR #1483, opened 2026-02-24
fix: solaris build tags, go.mod go version, go mod tidy, go fix

paralin:
This is a series of commits that makes minor changes to bring logrus up to date and fixes GopherJS compatibility.

Changes:

- Bump the `go` directive in `go.mod` from 1.13 to 1.23 and run `go mod tidy` to drop unused hashes from `go.sum`. Go 1.24 is the oldest release still receiving security updates per the [Go release policy](https://go.dev/doc/devel/release#policy), and Go 1.17 (the previous effective floor) was released in 2021 and is long unsupported.
- Run `go fix ./...` to convert legacy `// +build`directives to `//go:build` syntax, and rename the deprecated `reflect.Ptr` to `reflect.Pointer` (deprecated since Go 1.18).
- Add missing `//go:build` constraint to `terminal_check_solaris.go` (fixes GopherJS trying to build it)
- Add `gopherjs` to the no-terminal build constraint in `terminal_check_js.go` so that GopherJS builds correctly report no terminal support, fixing `gopherjs build ./...`.

Currently go1.26 is the latest release and go1.25 is the oldest supported release (with security updates). Bring the oldest version we test on GitHub up to 1.23.x (oldest supported version minus one). It doesn't make sense to keep testing against 1.17 which was released in 2021 and does not receive security updates anymore. The most recent update to 1.17 was 2021-09-09.

cc @thaJeztah these are minor changes to upstream so I can get rid of my fork, thanks!


terminal_check_solaris.go
```diff
@@ -1,3 +1,5 @@
+//go:build solaris
```
thaJeztah, line 1: It's probably fine to add, but the file has a `_solaris` suffix, which should already handle this.

terminal_check_no_terminal.go
```diff
@@ -1,4 +1,4 @@
-//go:build js || nacl || plan9
+//go:build js || nacl || plan9 || gopherjs
```
thaJeztah: Slightly on the fence for this, as `gopherjs` is not known by Go itself; shouldn't `gopherjs` set the `js` (or otherwise) build-tag?

terminal_check_notappengine.go
```diff
@@ -10,7 +11,7 @@ import (
 func checkIfTerminal(w io.Writer) bool {
 	switch v := w.(type) {
 	case *os.File:
-		return isTerminal(int(v.Fd()))
+		return isTerminal(int(v.Fd())) //nolint:gosec
```
thaJeztah: It's good to use a more specific "ignore" comment to prevent any other issues detected by gosec from being ignored; also to add a short comment describing why it's OK to ignore, e.g.;

```go
return isTerminal(int(v.Fd())) // #nosec G115 -- Ignore "integer overflow conversion"; Fd() is OS-provided, not user-controlled
```

thaJeztah: That said; perhaps we should switch to use `golang.org/x/term` instead of maintaining a custom implementation; I gave that a quick go; 

- https://github.com/sirupsen/logrus/pull/1484

thaJeztah (commented):
Thanks; yeah, overall it's probably fine to start dropping 1.17.

I left some quick comments 😅

terminal_check_no_terminal.go
thaJeztah: I see `gopherjs` build-tags were added in https://github.com/sirupsen/logrus/commit/c840e59446e71aa3fec56f9317fd2570cc8b3f5f

- https://github.com/sirupsen/logrus/pull/717


But changed to `js` tags in https://github.com/sirupsen/logrus/commit/eed1c0f832603c63deae9d5ffa2c4a6e611ed31f

> Fix GopherJS build tags
>
> The GopherJS build tag is "js" not "gopherjs"


- https://github.com/sirupsen/logrus/pull/798

terminal_check_notappengine.go
thaJeztah: Hm... so not sure yet if we can switch; CoPilot spotted that `golang.org/x/term` may still be limited.

We can properly handle the gosec warning here though, so that we don't have to suppress the linter;

```go
fd := f.Fd()
maxInt := uintptr(^uint(0) >> 1)
if fd > maxInt {
	return false
}
return isTerminal(int(v.Fd()))
```

terminal_check_solaris.go
paralin: I recall adding it because tinygo doesn't handle solaris internally and was failing the check, or something like that. These changes are from 2+ years of work, heh.

terminal_check_no_terminal.go
paralin: I'll drop this as I'm not even using gopherjs at the moment, anyway.

terminal_check_notappengine.go
paralin: Fixed!

paralin:
@thaJeztah fixed your comments!

entry.go
thaJeztah, line 290: Looks like we can use `maps.Clone` here so that we don't have to manually construct the `tmpHooks`. While at it perhaps use an early return and `fmt.Fprintln` to avoid the `Fprintf` formatting;

```go
	entry.Logger.mu.Lock()
	tmpHooks := maps.Clone(entry.Logger.Hooks)
	entry.Logger.mu.Unlock()
	if len(tmpHooks) == 0 {
		return
	}

	if err := tmpHooks.Fire(entry.Level, entry); err != nil {
		_, _ = fmt.Fprintln(os.Stderr, "Failed to fire hook:", err)
	}
}
```

entry.go
```diff
@@ -82,9 +83,7 @@ func NewEntry(logger *Logger) *Entry {
 
 func (entry *Entry) Dup() *Entry {
 	data := make(Fields, len(entry.Data))
-	for k, v := range entry.Data {
-		data[k] = v
-	}
+	maps.Copy(data, entry.Data)
 	return &Entry{Logger: entry.Logger, Data: data, Time: entry.Time, Context: entry.Context, err: entry.err}
 }
```
thaJeztah, line 92: We can use `maps.Clone` here to skip manually constructing the `data`; in that case we can make this a struct-literal, and don't create a new Fields struct if the original was `nil`;

```go
func (entry *Entry) Dup() *Entry {
	return &Entry{
		Logger:  entry.Logger,
		Data:    maps.Clone(entry.Data),
		Time:    entry.Time,
		Context: entry.Context,
		err:     entry.err,
	}
}
```

entry.go
```diff
@@ -113,29 +112,25 @@ func (entry *Entry) WithError(err error) *Entry {
 // WithContext adds a context to the Entry.
 func (entry *Entry) WithContext(ctx context.Context) *Entry {
 	dataCopy := make(Fields, len(entry.Data))
-	for k, v := range entry.Data {
-		dataCopy[k] = v
-	}
+	maps.Copy(dataCopy, entry.Data)
 	return &Entry{Logger: entry.Logger, Data: dataCopy, Time: entry.Time, err: entry.err, Context: ctx}
 }
```
thaJeztah, line 125: Same here;

```go
func (entry *Entry) WithContext(ctx context.Context) *Entry {
	return &Entry{
		Logger:  entry.Logger,
		Data:    maps.Clone(entry.Data),
		Time:    entry.Time,
		Context: ctx,
		err:     entry.err,
	}
}
```

entry.go
```diff
@@ -156,9 +151,7 @@ func (entry *Entry) WithFields(fields Fields) *Entry {
 // WithTime overrides the time of the Entry.
 func (entry *Entry) WithTime(t time.Time) *Entry {
 	dataCopy := make(Fields, len(entry.Data))
-	for k, v := range entry.Data {
-		dataCopy[k] = v
-	}
+	maps.Copy(dataCopy, entry.Data)
 	return &Entry{Logger: entry.Logger, Data: dataCopy, Time: t, err: entry.err, Context: entry.Context}
 }
```
thaJeztah, line 168: And here;

```go
func (entry *Entry) WithTime(t time.Time) *Entry {
	return &Entry{
		Logger:  entry.Logger,
		Data:    maps.Clone(entry.Data),
		Time:    t,
		Context: entry.Context,
		err:     entry.err,
	}
}
```

thaJeztah (changes requested):
Thanks! I looked slightly more in-depth at the changes; left some small suggestions, but LGTM after those are addressed.

I also did a quick check what minimum version of Go is required for changes made, to see if it's worth being more conservative;

- go1.17: `// +build` -> `//go:build`
- go1.18: `bytes.Cut`
- go1.18: `for i := range someInt { ... }`
- go1.18: `any`
- go1.18: `reflect.Pointer`
- go1.23: `maps.Copy`, `maps.Clone`

So from the above, all changes, except for the `maps.Copy` and `maps.Clone` could be done while keeping the minimum version at go1.18, but the `maps` improvements are definitely cleaner, and I personally think it's OK to raise to go1.23

entry.go
paralin: done!

entry.go
paralin: done!

entry.go
paralin: done!

entry.go
paralin: done!

thaJeztah (approved):
LGTM, thanks!

PR #1523, opened 2026-04-20
fix: preserve metadata in Entry.Dup

Bojun-Vvibe:
**Repo:** sirupsen/logrus (⭐ 24000)
**Type:** bugfix
**Files changed:** 2
**Lines:** +38/-0

## What
This change fixes `Entry.Dup()` so it preserves the full entry metadata when cloning an entry. In addition to cloning `Data`, `Logger`, `Time`, `Context`, and internal error state, the duplicate now also keeps `Level`, `Caller`, `Message`, and `Buffer`. A regression test was added to verify that those fields are copied while the `Data` map remains independently cloned.

## Why
`Entry.Dup()` is a public API intended to produce a reusable copy of an entry for further modification. Its implementation and documentation had drifted: the method claimed to copy the entry's other metadata by value, but it silently dropped key fields that affect formatting and log output. That can lead to incomplete or surprising duplicated entries for callers that rely on `Dup()` outside the narrow internal path.

## Testing
Formatted the touched Go files with `gofmt -w entry.go entry_test.go`.
Attempted `GOCACHE=/tmp/sirupsen-logrus-gocache GOMODCACHE=/tmp/sirupsen-logrus-gomodcache go test -run '^TestEntryDupCopiesMetadata$' ./...`, but the environment is offline and the repo does not vendor test dependencies, so Go could not fetch `github.com/stretchr/testify` and `golang.org/x/sys`.

## Risk
Low - the behavior now matches the method's documented contract, and the change only affects the `Entry.Dup()` copy path with a targeted regression test.

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -108,6 +108,10 @@ func (entry *Entry) Dup() *Entry {
 		Logger:  entry.Logger,
 		Data:    maps.Clone(entry.Data),
 		Time:    entry.Time,
+		Level:   entry.Level,
+		Caller:  entry.Caller,
+		Message: entry.Message,
+		Buffer:  entry.Buffer,
 		Context: entry.Context,
 		err:     entry.err,
 	}
--- a/entry_test.go
+++ b/entry_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"runtime"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


thaJeztah:
Can you describe the problem this is fixing? AFAICS, this function is intentionally not copying these fields, as they are propagated when logging; https://github.com/sirupsen/logrus/blob/e9e33f0093d019966ffa544903196f4a25aafafc/entry.go#L279-L297

thaJeztah (changes requested):
see my comment

PR #1527, opened 2026-04-25
expose interfaces for Logger and Entry

thaJeztah:
Export small interfaces covering the common logging methods.

These interfaces allow consumers to depend on behavior instead of
concrete Logrus types, making it easier to write reusable code and
to substitute alternative logging implementations.

This patch splits up the FieldLogger interface to smaller interfaces
that together compose a FieldLogger.Chunk up the extended parts
to smaller interfaces that together compose a FieldLogger.

### DIFF
--- a/logrus.go
+++ b/logrus.go
@@ -118,11 +118,26 @@ const (
 	TraceLevel
 )
 
-// Won't compile if StdLogger can't be realized by a log.Logger
+// Compile-time interface assertions.
 var (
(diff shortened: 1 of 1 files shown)


logrus.go
```diff
@@ -150,34 +150,56 @@ type FieldLogger interface {
 	WithError(err error) *Entry
 
 	StdLogger
+	StdLogger
```
thaJeztah, line 167: oh; duplicated this one

PR #1528, opened 2026-04-25
fix: build against tinygo

paralin:
Fixes the build against tinygo by removing the import to the unix syscalls package conditional on the tinygo build tag.

Fixes tinygo build ./


paralin:
cc @thaJeztah

thaJeztah:
Ah, and there I thought we had all cross builds covered; curious: is there an easy way to check this in CI? (I never really played with tinygo)

thaJeztah (approved):
LGTM

paralin:
@thaJeztah Thanks! I think we could add a tinygo target - https://tinygo.org/getting-started/install/linux/

PR #1547, opened 2026-07-23
fix: nil Entry.WithField doesn't panic

sonnemusk:
```go
var e *logrus.Entry
e.WithField("k", 1) // panic
```

Return nil for WithField / WithFields / WithTime when entry is nil.


sonnemusk:
Closing this — it was a defensive nil guard, not something I hit in real use. Callers should not pass nil here; panicking (or failing at the call site) is fine.

PR #1548, opened 2026-07-23
fix: return error when Format is called with a nil Entry

sonnemusk:
## Summary
`TextFormatter.Format(nil)` and `JSONFormatter.Format(nil)` paniced.

Return `logrus: Format called with nil Entry` instead.

## Test plan
- [x] `go test .`
- [x] Unit tests for both formatters

### DIFF
--- a/formatter_nil_test.go
+++ b/formatter_nil_test.go
@@ -1,0 +1,22 @@
+package logrus
+
+import (
+	"strings"
+	"testing"
+)
+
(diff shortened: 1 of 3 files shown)


sonnemusk:
Closing this — it was a defensive nil guard, not something I hit in real use. Callers should not pass nil here; panicking (or failing at the call site) is fine.

PR #1549, opened 2026-07-23
fix: trim surrounding space in ParseLevel

sonnemusk:
## What

`ParseLevel` trims surrounding whitespace before matching the level name.

## Why

Levels often come from env vars or config (`LOG_LEVEL= info `). Extra space should not make a valid name fail.

## Test

- `TestParseLevelTrimSpace`

### DIFF
--- a/logrus.go
+++ b/logrus.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"log"
+	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


sonnemusk:
Closing this PR. After review, this is too low-value / looks like bulk automated noise rather than a carefully chosen change. Sorry for the noise — will focus on real, high-signal fixes only.

PR #1551, opened 2026-07-24
Add slog.Handler writing to logrus.Logger

### LINKED ISSUE #1401: Feature: Add slog compatibility layer

sonnemusk:
## Summary

Implements the remaining half of #1401: a `slog.Handler` that writes records into a `logrus.Logger`. The Logrus→slog hook already landed in #1407 (`hooks/slog`); this PR adds the reverse bridge in the same package.

```go
logger := logrus.New()
slog.SetDefault(slog.New(sloghook.NewHandler(logger)))
slog.Info("hello", "key", "value")
```

### Behavior

- **Levels**: default mapping Debug/Info/Warn/Error (and below-Debug → Trace); override with `Handler.LevelMapper`
- **Attributes**: attached as logrus fields; groups flattened with `.` key prefixes (same style as `slog.TextHandler`)
- **Context / time**: preserved via `WithContext` / `WithTime`
- **Enabled**: consults the underlying logger after level mapping
- **Go version**: `go.mod` is already 1.23, so no build tag / separate module is needed

### Tests

Coverage for level mapping, custom mapper, level filtering, `WithAttrs`/`WithGroup`, inline groups, context/time, and panic on nil logger.

Closes #1401

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Fixes:
 Features:
 
   * Basic `slog` hook for bridging Logrus entries to `log/slog`.
+  * Add `slog.Handler` (`hooks/slog.NewHandler`) for bridging `log/slog` records
+    into a Logrus logger (levels, fields, groups, context, and time).
   * Add minimal, composable logging interfaces for each log level. This enables
     consumers to depend on narrower interfaces, making it easier to substitute
     or adapt logging implementations.
--- a/hooks/slog/handler.go
+++ b/hooks/slog/handler.go
@@ -1,0 +1,186 @@
+package slog
+
+import (
+	"context"
+	"log/slog"
+	"maps"
+	"slices"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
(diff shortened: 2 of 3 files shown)


thaJeztah:
Sorry, I'm a bit behind on reviewing; was there a reason to close this? (I haven't looked at the code yet); was the implementation not correct?

sonnemusk:
closed by accident while cleaning up forks, nothing wrong with the code.

same commit: https://github.com/sirupsen/logrus/pull/1553

thaJeztah:
Thanks! I'll try to have a look at that one soon 🤞

PR #1553, opened 2026-08-10
Add slog.Handler writing to logrus.Logger

### LINKED ISSUE #1401: Feature: Add slog compatibility layer

sonnemusk:
Other half of #1401: slog.Handler that writes into a logrus.Logger. Reverse of the hook from #1407.

```go
logger := logrus.New()
slog.SetDefault(slog.New(sloghook.NewHandler(logger)))
slog.Info("hello", "key", "value")
```

- levels: Debug/Info/Warn/Error by default, below Debug → Trace (override via `Handler.LevelMapper`)
- attrs → logrus fields; groups flattened with `.` (same idea as slog.TextHandler)
- keeps context + record time
- Enabled() follows the underlying logger level

tests for mapping, custom mapper, level filter, WithAttrs/WithGroup, inline groups, ctx/time, nil logger panic.

Closes #1401

(same patch as #1551 — closed that one by mistake, reopening here)

### DIFF
--- a/hooks/slog/handler.go
+++ b/hooks/slog/handler.go
@@ -1,0 +1,209 @@
+package slog
+
+import (
+	"context"
+	"log/slog"
+	"maps"
+	"slices"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Handler is a [slog.Handler] that writes records to a [logrus.Logger].
+//
+// It is intended for bridging libraries or application code that log via slog
+// into an existing Logrus logger, for example during a gradual migration.
+//
+// By default, slog levels map to their corresponding Logrus levels. Trace,
+// Fatal, and Panic use custom slog levels below Debug and above Error,
+// respectively, preserving their relative severity:
+//
+//   - [slog.LevelDebug] - 4 -> [logrus.TraceLevel]
+//   - [slog.LevelDebug] -> [logrus.DebugLevel]
+//   - [slog.LevelInfo] -> [logrus.InfoLevel]
+//   - [slog.LevelWarn] -> [logrus.WarnLevel]
+//   - [slog.LevelError] -> [logrus.ErrorLevel]
+//   - [slog.LevelError] + 2 -> [logrus.FatalLevel]
+//   - [slog.LevelError] + 4 -> [logrus.PanicLevel]
+//
+// Levels between these boundaries map to the next lower Logrus severity.
+// Levels below Debug map to [logrus.TraceLevel], and levels above Panic map
+// to [logrus.PanicLevel].
+//
+// Mapping to [logrus.FatalLevel] or [logrus.PanicLevel] preserves the level
+// only; handling a record does not exit or panic.
+//
+// Example usage:
+//
+//	logger := logrus.New()
+//	slog.SetDefault(slog.New(NewHandler(logger)))
+//	slog.Info("hello", "key", "value")
+type Handler struct {
+	logger *logrus.Logger
+
+	// LevelMapper maps slog levels to Logrus levels. If nil, the default
+	// mapping is used. Set it to customize level mapping, for example to map
+	// custom slog levels to specific Logrus levels.
+	LevelMapper func(slog.Level) logrus.Level
+
+	// fields holds attributes from prior WithAttrs calls, already resolved
+	// under the group prefix that applied when they were added.
+	fields logrus.Fields
+
+	// groups is the current group name stack for attributes added later.
+	groups []string
+}
+
+var _ slog.Handler = (*Handler)(nil)
+
+// NewHandler creates a [slog.Handler] that writes to the provided
+// [logrus.Logger].
+//
+// The provided logger must not be nil. NewHandler panics if logger is nil.
+func NewHandler(logger *logrus.Logger) *Handler {
+	if logger == nil {
+		panic("cannot create handler from nil logger")
+	}
+	return &Handler{
+		logger: logger,
+	}
+}
+
+// toLogrusLevel maps a slog level using LevelMapper or the default mapping.
+func (h *Handler) toLogrusLevel(level slog.Level) logrus.Level {
+	if h.LevelMapper != nil {
+		return h.LevelMapper(level)
+	}
+	switch {
+	case level >= slogLevelPanic:
+		return logrus.PanicLevel
+	case level >= slogLevelFatal:
+		return logrus.FatalLevel
+	case level >= slogLevelError:
+		return logrus.ErrorLevel
+	case level >= slogLevelWarn:
+		return logrus.WarnLevel
+	case level >= slogLevelInfo:
+		return logrus.InfoLevel
+	case level >= slogLevelDebug:
+		return logrus.DebugLevel
+	case level >= slogLevelTrace:
+		return logrus.TraceLevel
+	default:
+		return logrus.TraceLevel
+	}
+}
+
+// Enabled reports whether the handler handles records at the given level.
+// It maps the slog level to a logrus level and consults the underlying logger.
+func (h *Handler) Enabled(_ context.Context, level slog.Level) bool {
+	return h.logger.IsLevelEnabled(h.toLogrusLevel(level))
+}
+
+// Handle converts the slog record into a logrus entry and logs it.
+// Record time, context, message, and attributes (including those from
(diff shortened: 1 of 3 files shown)


hooks/slog/handler.go
```diff
@@ -0,0 +1,186 @@
...
+		return logrus.TraceLevel
+	}
+}
+
+func (h *Handler) toLogrusLevel(level slog.Level) logrus.Level {
+	if h.LevelMapper != nil {
+		return h.LevelMapper(level)
+	}
+	return DefaultSlogLevelMapper(level)
+}
```
thaJeztah: I want to reduce the public API; `DefaultSlogLevelMapper` is only used to set the defaults; external callers wouldn't need it (they can leave the `LevelMapper` nil, which means that the defaults are used).

I was also having another look at the Hook implementation and realised we don't have to drop the extended levels provided by Logrus (similar to https://github.com/containerd/log/pull/12) so I updated the Hook's mapping, and added consts that we can use for clarity;

- https://github.com/sirupsen/logrus/pull/1561

We can take the same approach here as is used for the Hook; https://github.com/sirupsen/logrus/blob/ed957d512a86b08ac978d64f65ae256a1b56263f/hooks/slog/slog.go#L67-L91


Then document the mapping on the `Handler` struct; https://github.com/sirupsen/logrus/blob/ed957d512a86b08ac978d64f65ae256a1b56263f/hooks/slog/slog.go#L21-L36

And put relevant docs that are currently on `DefaultSlogLevelMapper` on the map's field; https://github.com/sirupsen/logrus/blob/ed957d512a86b08ac978d64f65ae256a1b56263f/hooks/slog/slog.go#L39-L42



HOWEVER; when looking at that; this would also add `Panic` and `Fatal` levels; logrus may make those run `os.Exit(1)` or `panic` - that's probably unexpected for the slog integration; there was a bug in the code that didn't work as advertised; opened a PR for that so that we can integrate those levels;

- https://github.com/sirupsen/logrus/pull/1564

hooks/slog/handler.go
```diff
@@ -0,0 +1,186 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+// SlogLevelMapper maps a [slog.Level] to a [logrus.Level].
+//
+// To change the default level mapping, for instance to map custom slog levels
+// or to send [slog.LevelError] to [logrus.FatalLevel], set
+// [Handler.LevelMapper] to your own implementation of this function.
+type SlogLevelMapper func(slog.Level) logrus.Level
```
thaJeztah: We can drop this type; I noticed we did the same for the Hook, but then realised it doesn't bring us much, and it's better to just put the function signature in the `Handler` struct directly, and document it there; see

- https://github.com/sirupsen/logrus/pull/1559

hooks/slog/handler.go
```diff
@@ -0,0 +1,186 @@
...
+	return h.logger.IsLevelEnabled(h.toLogrusLevel(level))
+}
+
+// Handle converts the slog record into a logrus entry and logs it.
+// Record time, context, message, and attributes (including those from
+// WithAttrs/WithGroup) are preserved. Attributes are attached as logrus fields;
+// group names are joined with "." as a key prefix (similar to [slog.TextHandler]).
+func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
+	level := h.toLogrusLevel(record.Level)
+
```
thaJeztah, line 114: Given the extra processing needed to do the conversion, perhaps it's worth taking advantage of the level here and return early if the log-level is not enabled;

```go
	level := h.toLogrusLevel(record.Level)
	if !h.logger.IsLevelEnabled(level) {
		return nil
	}
```

hooks/slog/handler.go
```diff
@@ -0,0 +1,186 @@
...
+	record.Attrs(func(a slog.Attr) bool {
+		appendAttr(fields, h.groups, a)
+		return true
+	})
+
+	entry := logrus.NewEntry(h.logger).WithContext(ctx).WithTime(record.Time)
+	if len(fields) > 0 {
+		entry = entry.WithFields(fields)
+	}
+	entry.Log(level, record.Message)
```
thaJeztah, line 134: We'll probably pay the price of duplicating the entry 2 - 3 times here (each `WithXXX` creates a copy of the entry, and clones the `Field.Data` map); as all the fields we need to set on the entry are exported, we can use a struct-literal to create the entry;

 ```go
	entry := &logrus.Entry{
		Logger:  h.logger,
		Data:    fields,
		Time:    record.Time,
		Context: ctx,
	}
	entry.Log(level, record.Message)
```

**Review by thaJeztah — CHANGES_REQUESTED** [review]
thanks! overall this looks good; I left some comments, and was already drafting some of those changes, so will do a quick rebase of your PR and push commits with those changes (to be squashed)

File: hooks/slog/handler.go [hunk 5]
```diff
@@ -0,0 +1,206 @@
...
+
+	if n := record.NumAttrs(); n > 0 {
+		if entry.Data == nil {
+			entry.Data = make(logrus.Fields, n)
+		}
+		record.Attrs(func(a slog.Attr) bool {
+			appendAttr(entry.Data, h.groups, a)
+			return true
+		})
+	}
```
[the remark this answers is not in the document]
thaJeztah: Yeah, good point; will update.

thaJeztah:
> * Valid non-group slog attributes may have an empty key, but this branch drops them even though the zero Attr was already filtered above. Built-in slog handlers retain such attributes, and `logrus.Fields` can represent the empty key, so records containing `slog.Attr{Key: "", Value: ...}` silently lose data. Only an empty key on a group should trigger inline-group behavior.
> 
> ```
> 	if attr.Key == "" {
> 		return
> 	}
> ```


Hm; yeah possibly we should do that.
 
> **hooks/slog/handler.go:117**
> 
> * When `Logger.ReportCaller` is enabled, leaving `Entry.Caller` unset makes `Entry.Log` inspect the current stack and report `Handler.Handle` as the caller rather than the original slog call site. `slog.Record.PC` carries that original source location and should be converted to `runtime.Frame` when caller reporting is enabled. A race-safe fix likely needs a logger accessor for the protected `ReportCaller` setting, plus a caller-preservation test analogous to the reverse slog hook.
> 
> ```
> 	entry := &logrus.Entry{
> 		Logger:  h.logger,
> 		Data:    h.fields,
> ```


Yup, already noticed that one, and made preparations to make that work; will do it in a follow-up;

- https://github.com/sirupsen/logrus/pull/1558

 
> **hooks/slog/handler.go:118**
> 
> * The claimed record-time preservation is not tested with a known record timestamp. `TestHandler_contextAndTime` sends `ts` only as an attribute and merely checks that the logger-generated entry time is nonzero, so replacing `record.Time` with the current time would still pass. Add a direct `Handle` test using `slog.NewRecord(ts, ...)` and assert `entry.Time.Equal(ts)`.
> 
> ```
> 		Time:    record.Time,
> ```


Good one; added a test

thaJeztah:
> **hooks/slog/handler.go:208**
> 
> * Assigning `attr.Value.Any()` directly into `Entry.Data` bypasses Logrus's unsupported-field handling in `Entry.WithField(s)`. A slog attribute containing a function or pointer-to-function therefore reaches `JSONFormatter`, whose marshal fails and causes the entire log record (including otherwise valid fields) to be dropped; normal Logrus field APIs instead omit that value and emit `logrus_error`. Route the flattened fields through `Entry.WithFields` (or provide equivalent validation/error propagation) before logging.
> 
> ```
> 	fields[key] = attr.Value.Any()
> ```


Yup, considered that one; for now I think it's too much of a corner case (and reason for complexity in logrus itself); for now I don't think we need to bother with that.

thaJeztah:
I'm gonna squash the commits; changes look good now 👍  (will do a follow-up for the `Entry.Caller` and some other minor changes.

thaJeztah (approved):
LGTM, thanks!

PR #1581, opened 2026-08-19
test: add TestLevelStringUnknown and TestParseLevelVariations in level_test.go

aoright:
### Summary
- In `level_test.go`, add unit test `TestLevelStringUnknown` to verify `Level.String()` for unknown level values.
- Add `TestParseLevelVariations` to verify case-insensitive matching and alias handling (e.g., `warn`/`warning`, uppercase, invalid input).

### DIFF
--- a/level_test.go
+++ b/level_test.go
@@ -60,3 +60,43 @@ func TestLevelMarshalText(t *testing.T) {
 		})
 	}
 }
+
+func TestLevelStringUnknown(t *testing.T) {
+	unknownLevel := logrus.Level(99)
+	require.Equal(t, "unknown", unknownLevel.String())
+}
+
+func TestParseLevelVariations(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected logrus.Level
+		hasErr   bool
+	}{
+		{"panic", logrus.PanicLevel, false},
+		{"PANIC", logrus.PanicLevel, false},
+		{"fatal", logrus.FatalLevel, false},
+		{"error", logrus.ErrorLevel, false},
+		{"warn", logrus.WarnLevel, false},
+		{"WARN", logrus.WarnLevel, false},
+		{"warning", logrus.WarnLevel, false},
+		{"WARNING", logrus.WarnLevel, false},
+		{"info", logrus.InfoLevel, false},
+		{"INFO", logrus.InfoLevel, false},
+		{"debug", logrus.DebugLevel, false},
+		{"trace", logrus.TraceLevel, false},
+		{"TRACE", logrus.TraceLevel, false},
+		{"invalid", logrus.Level(0), true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.input, func(t *testing.T) {
+			lvl, err := logrus.ParseLevel(tt.input)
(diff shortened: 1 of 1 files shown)


level_test.go
```diff
@@ -60,3 +60,43 @@ func TestLevelMarshalText(t *testing.T) {
...
+		expected logrus.Level
+		hasErr   bool
+	}{
+		{"panic", logrus.PanicLevel, false},
+		{"PANIC", logrus.PanicLevel, false},
+		{"fatal", logrus.FatalLevel, false},
+		{"error", logrus.ErrorLevel, false},
+		{"warn", logrus.WarnLevel, false},
+		{"WARN", logrus.WarnLevel, false},
+		{"warning", logrus.WarnLevel, false},
```
thaJeztah, line 81: I like the table-test,  but we should instead make this replace the existing test(s) (and move it to that file, because `ParseLevel` is defined in `logrus.go`; I actually wanted to update those to use a table-test, so perhaps you can update those so that we don't duplicate the exact tests?


Possibly the same test could cover `text -> level` and `level text` within the same table so that we don't end up with 2 table tests, but I haven't tried.


https://github.com/sirupsen/logrus/blob/134c80f81c9667a2d8c306f159d474857a20b735/logrus_test.go#L530-L613


Oh! One thing worth mentioning; that `logrus_test.go` uses dot-imports for logrus; that's not consistent with other tests-files we have, but changing it would cause too much code-churn, so I left that for now (and we probably should .. for now);
https://github.com/sirupsen/logrus/blob/134c80f81c9667a2d8c306f159d474857a20b735/logrus_test.go#L18-L19

level_test.go
```diff
@@ -60,3 +60,43 @@ func TestLevelMarshalText(t *testing.T) {
...
+		{"WARNING", logrus.WarnLevel, false},
+		{"info", logrus.InfoLevel, false},
+		{"INFO", logrus.InfoLevel, false},
+		{"debug", logrus.DebugLevel, false},
+		{"trace", logrus.TraceLevel, false},
+		{"TRACE", logrus.TraceLevel, false},
```
thaJeztah, line 87: When updating, could you avoid implicit fields here?

```go
{input: "debug", expected: logrus.DebugLevel, hasErr: false},
```

level_test.go
```diff
@@ -60,3 +60,43 @@ func TestLevelMarshalText(t *testing.T) {
...
+		{"invalid", logrus.Level(0), true},
+	}
+
+	for _, tt := range tests {
```
thaJeztah, line 91: Minor nit; we mostly converged on using `tc` (for "test case") for these tables; can you update it to use the same for consistency?

```go
for _, tc := range tests {
```

thaJeztah (changes requested):
Thanks! I had a look and it looks like most lines are already covered by tests, _except_ for these 3; https://github.com/sirupsen/logrus/blob/134c80f81c9667a2d8c306f159d474857a20b735/level.go#L88-L91


Left suggestions to replace the existing tests with your table approach; let me know if you need more info.

When updating, please amend / squash your commit 😅

PR #1589, opened 2026-09-08
fix: update sys dependence for CVE-2026-39824

yshngg:
See [NVD-CVE-2026-39824](https://nvd.nist.gov/vuln/detail/cve-2026-39824).


thaJeztah:
That vulnerability impacts the `NewNTUnicodeString` function on Windows;

- https://pkg.go.dev/vuln/GO-2026-5024

Which is not used by Logrus, so a false positive.

In addition, this module is a library module, and adheres to Go MVS (Minimum Version Selection); https://research.swtch.com/vgo-mvs, which means that it specifies the _minimum_ required version to use the module; users of this module can specify any SemVer compatible newer version of these dependencies (which they may need to do if they are impacted by this CVE).


I'm closing this PR because of the above, but feel free to continue the conversation.

